PR #23, opened 2015-10-01
Refactoring file locking out of tusd.Handler

### LINKED ISSUE #22: Remove locking from handler implementation
Hi!
I'm experimenting with the `tusd` implementation as the basis for a production-grade TUS server. One of the things I've noticed in our tests is that at times the `tusd.Handler` can get files to stay in locked state, with no way of being unlocked. We don't think we need this locking behavior in the handler, and can delegate this to the file store (we're using a DB as our back-end, which handles concurrent read/writes).

I have a (forthcoming) pull request which refactors the locking behavior to the `filestore.Filestore` type.

Best,
Adam

abkaplan07:
This resolves #22 

* Updated filestore.FileStore implementation to use simple locking scheme. Refactored to make interface use pointers. Removed locks from tusd.Handler
* Refactored sample code to use pointer implementation of FileStore.
* Moved ErrFileLocked variable to datastore.go

### DIFF
--- a/cmd/tusd/main.go
+++ b/cmd/tusd/main.go
@@ -39,9 +39,7 @@ func main() {
 	}
 
 	var store tusd.DataStore
-	store = filestore.FileStore{
-		Path: dir,
-	}
+	store = filestore.NewFileStore(dir)
 
 	if storeSize > 0 {
 		store = limitedstore.New(storeSize, store)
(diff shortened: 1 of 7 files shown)


Acconut:
Very similar mechanism has been implemented in #37.

PR #26, opened 2015-11-02
Make sure HEAD responses have no body

vayam:
Replacing sendError because HEAD requests can't have body.

Issue:  Request hangs when

- Missing version header
curl -X HEAD http://localhost:1080/files/210da45bac71209e09e7d1649daf2715 

- Provided Incorrect version
curl -X HEAD -H "Tus-Resumable: 1" http://localhost:1080/files/210da45bac71209e09e7d1649daf2715 v

- requested Invalid resource
curl -X HEAD -H "Tus-Resumable: 1.0.0" http://localhost:1080/files/InvalidResource -v

Solution: Don't use sendError and replace it with a no body response


Acconut:
Thanks for this fix! I actually moved to check whether to send the body into sendError to ensure only HEAD requests get an empty reponse.

vayam:
@Acconut cool. that looks better.

PR #29, opened 2015-11-05
Expose handler methods

gingermusketeer:
This PR is driven by a couple of desires for the app i am building. 
- I wanted to be able to remove the DELETE endpoint as it is not a feature needed by the app.
- I wanted to be able to customise the GET endpoint to return json with the status of the background job which is initiated by the upload rather than the actual file uploaded.
- The app i am building uses httprouter and was hoping to avoid nested http routers for performance reasons.

Hopefully these are reasonable from your perspective! 

Let me know if you want anything changed

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ func main() {
 	// Create a new HTTP handler for the tusd server by providing
 	// a configuration object. The DataStore property must be set
 	// in order to allow the handler to function.
-	handler, err := tusd.NewHandler(tusd.Config{
+	handler, err := tusd.NewRoutedHandler(tusd.Config{
 		BasePath:              "files/",
 		DataStore:             store,
 	})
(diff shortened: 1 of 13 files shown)


kvz:
@Acconut

gingermusketeer:
Just rebased this on the new changes in master

Acconut:
First of all, I want to apologize for letting you wait so long until I respond, but I had some private issues to deal with. :) Anyway, I can follow your arguments, except the last one (avoid nested http routers for performance reasons) but I haven't done any benchmarking on this so I will just accept it :)

I like the approach of exposing the handler so custom routing logic can be implemented but I do not think it is necessary to extract the routing to a new struct. You can use the exposed handler on your own or use the provided ServeHTTP method. Any opinions?

gingermusketeer:
@Acconut No problem. 

Avoiding nested routers is definitely not the main reason for this PR :)

I thought about just exposing the handlers and leaving everything else as is but felt that having these two different levels made the it cleaner

Acconut:
@gingermusketeer That would be great! But instead of BasicHandler, I would prefer UnroutedHandler – if you don't mid – to illustrate the difference between it and the Handler struct.

Acconut:
Closed in favour of #36.

PR #31, opened 2015-11-14
Minor fixes

vayam:
- Fix cors headers
- reuse reason variabl

### DIFF
--- a/handler.go
+++ b/handler.go
@@ -42,7 +42,7 @@ var ErrStatusCodes = map[error]int{
 	ErrInvalidOffset:       http.StatusBadRequest,
 	ErrNotFound:            http.StatusNotFound,
 	ErrFileLocked:          423, // Locked (WebDAV) (RFC 4918)
-	ErrMismatchOffset:       http.StatusConflict,
+	ErrMismatchOffset:      http.StatusConflict,
 	ErrSizeExceeded:        http.StatusRequestEntityTooLarge,
 	ErrNotImplemented:      http.StatusNotImplemented,
 	ErrUploadNotFinished:   http.StatusBadRequest,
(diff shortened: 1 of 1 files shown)


handler.go
```diff
@@ -145,7 +145,7 @@ func (handler *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 		if r.Method == "OPTIONS" {
 			// Preflight request
-			header.Set("Access-Control-Allow-Methods", "POST, HEAD, PATCH, OPTIONS")
+			header.Set("Access-Control-Allow-Methods", "POST, GET, HEAD, PATCH, DELETE")
```
Acconut: Why did you remote the OPTIONS method? A client may send a OPTIONS request to determinate the server's configuration.

handler.go
```diff
@@ -453,15 +453,17 @@ func (handler *Handler) sendError(w http.ResponseWriter, r *http.Request, err er
...
-	reason := err.Error()
+	reason := err.Error() + "\n"
 	if r.Method == "HEAD" {
 		reason = ""
 	}
 
-	w.Header().Set("Content-Type", "text/plain")
+	//https://golang.org/src/net/http/server.go#L1429
+	w.Header().Set("Content-Type", "text/plain; charset=utf-8")
+	w.Header().Set("X-Content-Type-Options", "nosniff")
```
Acconut: Great addition, but what about adding this to all responses and not just in case of an error?

handler.go
vayam: You are right. I will add `OPTIONS` back in.

handler.go
vayam: Yeah makes sense.

PR #34, opened 2015-11-21
Sorting before termination of uploads in limitedstore.go

anirudt:
This fix is in limitedstore/limitedstore.go, to intelligently terminate existing file uploads. The ```store.uploads``` map is sorted, and then analysed to remove the bigger existing uploads first. 

The extraneous Pair and Pairlist structures are created in order to sort the map by values. Sorry for the indentation issues, surprisingly they didn't sync.

### DIFF
--- a/limitedstore/limitedstore.go
+++ b/limitedstore/limitedstore.go
@@ -106,7 +106,7 @@ func (store *LimitedStore) ensureSpace(size int64) error {
 
   // Forward traversal through the
   // uploads in terms of size, biggest upload first
-  for _,k := sorted_uploads {
+  for _,k := range sorted_uploads {
     id := k.key
 		if err := store.terminate(id); err != nil {
 			return err
@@ -116,7 +116,6 @@ func (store *LimitedStore) ensureSpace(size int64) error {
 			// Enough space has been freed to store the new upload
 			return nil
 		}
-    j--
 	}
 
 	return nil
--- a/limitedstore/limitedstore.go
+++ b/limitedstore/limitedstore.go
@@ -14,6 +14,7 @@ package limitedstore
 import (
 	"github.com/tus/tusd"
 	"sync"
+  "sort"
 )
 
 type LimitedStore struct {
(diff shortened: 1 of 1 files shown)


Acconut:
I apologized for my delayed response. Anyway, this is a great addition. :)

limitedstore/limitedstore.go
```diff
@@ -82,8 +96,19 @@ func (store *LimitedStore) ensureSpace(size int64) error {
...
 	}
-
-	for id, _ := range store.uploads {
+  sorted_uploads := make(Pairlist, len(store.uploads))
+  i := 0
+  for u,h := range store.uploads {
+    sorted_uploads[i] = Pair{u, h}
+    i++
+  }
+  sort.Sort(sorted_uploads)
```
Acconut: Is there a particular reason why you used sort.Sort and then iterate through the slice manually instead of using sort.Reverse (https://golang.org/pkg/sort/#Reverse)? As far as I can tell from looking at the source code, they are equivalent in terms of efficiency.

limitedstore/limitedstore.go
```diff
@@ -26,6 +27,19 @@ type LimitedStore struct {
 	mutex *sync.Mutex
 }
 
+// Pair structure to perform map-sorting
+type Pair struct {
```
Acconut: I would be pleased if this struct is not exported outside of this package. However, I can do this by myself when merging this PR if you don't want to :)

Acconut:
> Sorry for the indentation issues, surprisingly they didn't sync.

Doesn't matter, I will just run it through `go fmt` to automatically fix this :)

limitedstore/limitedstore.go
anirudt: Right! Please find the changed version in the last commit.

limitedstore/limitedstore.go
anirudt: Sure. I've done it in 84ae1cb. :)

PR #35, opened 2015-11-25
Add test for content length on 404

gingermusketeer:
This is a test for an issue which was resolved in #31. This test was extracted from #29

### DIFF
--- a/handler_test.go
+++ b/handler_test.go
@@ -44,7 +44,7 @@ type httpTest struct {
 	ResHeader map[string]string
 }
 
-func (test *httpTest) Run(handler http.Handler, t *testing.T) {
+func (test *httpTest) Run(handler http.Handler, t *testing.T) *httptest.ResponseRecorder {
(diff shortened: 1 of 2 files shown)


Acconut:
Thank you for this change. However, I implement it using a simpler approach in 88aed988f9646eb22caa15800e8d804f6d7c35f5. :)

gingermusketeer:
Awesome. That is a lot simpler :)

gingermusketeer:
@Acconut Looking at https://github.com/tus/tusd/blob/master/handler_test.go#L77 i am not sure if the test in 88aed988f9646eb22caa15800e8d804f6d7c35f5 is actually checking that the body length is 0. Thoughts?

Acconut:
Oh man, right, I forgot. Thanks for this notice. I will fix this :)

Acconut:
Addressed this in dd6ce926510984d0d05c6ee6b9786540d3a9c0c7.

PR #37, opened 2015-12-09
Separate locking into data store implementation

Acconut:
The goal is to allow different locking mechanism to be used for different
storages. For example, in-memory for very few uploads, a file locker in
conjunction with the FileStore or an external service, such as ZooKeeper.

* [x] Add tests
* [x] Add documentation
* [x] Add file locker
* [x] Prevent data race when handling PATCH request


gingermusketeer:
Awesome. Thanks a lot!!

Acconut:
Hehe, no problem :)

PR #46, opened 2016-04-27
Support 0 as a "no timeout" setting

cjhenck:
If you are working behind a load balancer the load balancer might handle this and send an EOF.  In that case, you want to have no read and write timeouts at all, so we should support the 0 setting as "unlimited" just like the Connection itself.

### DIFF
--- a/cmd/tusd/main.go
+++ b/cmd/tusd/main.go
@@ -255,18 +255,32 @@ type Conn struct {
 }
 
 func (c *Conn) Read(b []byte) (int, error) {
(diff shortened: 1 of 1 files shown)


Acconut:
Thank you a lot; this looks very good. Could you please update the flag's description, too (https://github.com/cjhenck/tusd/blob/patch-1/cmd/tusd/main.go#L58)?

cjhenck:
Done!

Acconut:
Thank you a again. This will land in the next release.

PR #50, opened 2016-06-29
feat(cmd/tusd): Print the base path

Rio:
Just nice to see.

### DIFF
--- a/cmd/tusd/main.go
+++ b/cmd/tusd/main.go
@@ -157,6 +157,8 @@ func main() {
 	address := httpHost + ":" + httpPort
 	stdout.Printf("Using %s as address to listen.\n", address)
 
+	stdout.Printf("Using %s as the base path.\n", basepath)
+
 	stdout.Printf(composer.Capabilities())
 
 	go func() {


Acconut:
Looks good, thank you :)

Acconut:
Also, nice work on the tus-rust project :+1:

Rio:
Thanks, It's my first rust project but I'm a bit stuck on testing the api
without spawning servers and stuff. I will complete it at one point but
right now we want to take this server in production and do away with our
own wrapper around the go tus library. We're just having some problems with
getting the Location header to return either relative url's or an absolute
url of our choosing.

Acconut:
Sounds exciting.

Regarding your issue, I would point you towards the function which is responsible for generating the URLs ending up in the Location header, here: https://github.com/tus/tusd/blob/master/unrouted_handler.go#L556-L569

Basically, if you provide an absolute URL to the config (https://github.com/tus/tusd/blob/master/config.go#L26) it should work. However, looking at the code I am not entirely sure if it still works :S

PR #53, opened 2016-07-14
New Greeting Propoal

kvz:
This is a proposal for the new welcome message, thx also to @ajvanloon

### DIFF
--- a/cmd/tusd/cli/greeting.go
+++ b/cmd/tusd/cli/greeting.go
@@ -8,24 +8,26 @@ import (
 var greeting string
 
 func PrepareGreeting() {
+	// @todo: Introduce configurable metrics endpoint.
+	// See: https://github.com/tus/tusd/issues/52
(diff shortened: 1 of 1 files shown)


cmd/tusd/cli/greeting.go
```diff
@@ -8,23 +8,26 @@ import (
 var greeting string
 
 func PrepareGreeting() {
+	// @todo: Introduce configurable metrics endpoint.
+	// See: https://github.com/tus/tusd/issues/52
```
Acconut: Why did this end up here? :)

kvz: As it already acknowledges a configurable metrics endpoint via %s, rather than hardcoding in the text

Acconut:
Thank you and @ajvanloon! Merged in fbf378547ef96fe8057ac9d6caa644ef0491ba7c

PR #54, opened 2016-07-15
Adjust in basepath parameter in example from README.

juloliveira:
Without slash (/) don´t work and Including requirements projects in README


juloliveira:
Is better describes how install dependencies with "go get ./..."

PR #58, opened 2016-08-12
Add `Authorization` to `Access-Control-Allow-Headers`.

kaleworsley:


unrouted_handler.go
```diff
@@ -139,12 +139,13 @@ func (handler *UnroutedHandler) Middleware(h http.Handler) http.Handler {
 			if r.Method == "OPTIONS" {
 				// Preflight request
 				header.Set("Access-Control-Allow-Methods", "POST, GET, HEAD, PATCH, DELETE, OPTIONS")
-				header.Set("Access-Control-Allow-Headers", "Origin, X-Requested-With, Content-Type, Upload-Length, Upload-Offset, Tus-Resumable, Upload-Metadata")
+				header.Set("Access-Control-Allow-Headers", "Origin, X-Requested-With, Content-Type, Upload-Length, Upload-Offset, Tus-Resumable, Upload-Metadata, Authorization")
 				header.Set("Access-Control-Max-Age", "86400")
+				header.Set("Access-Control-Allow-Credentials", "true")
```
Acconut: I believe this is a default value which has a high security impact and I would prefer to not have credentials enabled by default.

unrouted_handler.go
```diff
@@ -139,12 +139,13 @@ func (handler *UnroutedHandler) Middleware(h http.Handler) http.Handler {
...
 
 			} else {
 				// Actual request
-				header.Set("Access-Control-Expose-Headers", "Upload-Offset, Location, Upload-Length, Tus-Version, Tus-Resumable, Tus-Max-Size, Tus-Extension, Upload-Metadata")
+				header.Set("Access-Control-Expose-Headers", "Upload-Offset, Location, Upload-Length, Tus-Version, Tus-Resumable, Tus-Max-Size, Tus-Extension, Upload-Metadata, Authorization")
```
Acconut: Why should the Authorization header by also expose to the browser? I thought it is only included in the HTTP requests but not responses?

Acconut:
Closing this pull request due to a lack of information. Feel free to leave a comment if you want to continue the discussion :)

pguth:
Using Chrome 56 and `tusd` I see this when setting `withCredentials: true` in `tus-js-client`:

screen shot 2017-03-22 at 17 33 03

Acconut:
@pguth This is expected. The `withCredentials` option is a powerful one which also may open a great vulnerability if not used properly. That's why tusd does not the the `Access-Control-Allow-Credentials` header which is required to make this option work. However, if you have no reverse proxy enabled, the `withCredentials` option does not have any effect because tusd does not use Cookies. May I ask why you want to use it, thou?

pguth:
For quick prototyping. I expected to get the cookies payload in the `pre-create` script and authenticate/allow the upload-request there.

Acconut:
> I expected to get the cookies payload in the pre-create script and authenticate/allow the upload-request there.

Cookies are not part of the payload but you should be able to use upload metadata for these purposes. They are passed to the hooks script.

pguth:
I guessed so, thanks! Maybe we could add a sentence to https://github.com/tus/tusd/blob/master/docs/hooks.md?!

Acconut:
Good idea, done!

PR #62, opened 2016-09-15
fix(hooks): check if hook exits on disk

Rio:
Prevents it from just crashing and returning 500
to the client.

### DIFF
--- a/cmd/tusd/cli/hooks.go
+++ b/cmd/tusd/cli/hooks.go
@@ -70,10 +70,18 @@ func invokeHookSync(typ HookType, info tusd.FileInfo, captureOutput bool) ([]byt
 		return nil, nil
 	}
 
-	name := string(typ)
-	stdout.Printf("Invoking %s hook…\n", name)
+	// Build path to hook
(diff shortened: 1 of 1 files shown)


Rio:
Eh... That wasn't me. Maybe use something like glide to make installing these dependencies a bit easier? Not really sure what the go to package management tool is in go.

Acconut:
*Thank you* for reporting this issue and also providing a fix. However, there is a slightly more efficient method which saves the call to `os.Stat`. This check will also be performed by the `os/exec` package (see https://github.com/golang/go/blob/80423f1e64f1e939cddc455a29e5111527cd16f8/src/os/exec/lp_unix.go#L20) and therefore we can inspect the error afterwards. I've taken the liberty of implementing this :)

Rio:
Thanks! Still got lots of go to learn.

Acconut:
Actually, this is not a technique specific to Go but can be performed in any language. Basically, instead of checking whether a file exists before accessing it, simply access it immediately and look whether you get an error saying the file could not be found. Makes it just slightly more efficient and can be in some cases even atomic.

PR #71, opened 2016-11-02
Disable UseDNS for sshd

tersmitten:
You can also use a [ssh-server role](https://github.com/Oefenweb/ansible-ssh-server) if you want to customize more settings or if you don't like the (inline) tasks.


kvz:
I think this is fine for now, thanks!

PR #73, opened 2016-11-09
Upgrade ansistrano roles

tersmitten:


kvz:
I think this one is invalid as it still caries a `v`, no?

tersmitten:
Yes, it's invalid

tersmitten:
I'll need update the fork first

PR #87, opened 2016-12-26
Fix README sample code errors

mrezai:
The README sample code should work now.


Acconut:
Thank you for this contribution. I added some more patches to the example in the README, so it should indeed work now.

PR #94, opened 2017-01-19
Add -s3-endpoint to command line arguments to support S3 compatilbe implementations like minio

mrezai:
How to test:
Download and run [minio](https://github.com/minio/minio) : ./minio server /tmp/minio-test-files
Go to http://127.0.0.1:9000 (minio web UI) and create a bucket with name "test"
Set AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY and AWS_REGION from output of minio.
Run : go run cmd/tusd/main.go -s3-bucket test -s3-endpoint http://127.0.0.1:9000.
Run tus-js-client demo and set upload endpoit to http://localhost:1080/files/ and chunk size to 5242880 bytes(5MB).
Upload some files to test.
Refresh minio web UI.

This maybe work with new version of Ceph(Kraken) but not with Jewel because of an old version of embedded web server(civetweb) that doesn't support absolute URI.

### DIFF
--- a/cmd/tusd/cli/composer.go
+++ b/cmd/tusd/cli/composer.go
@@ -32,12 +32,20 @@ func CreateComposer() {
 		store := filestore.New(dir)
 		store.UseIn(Composer)
 	} else {
-		stdout.Printf("Using 's3://%s' as S3 bucket for storage.\n", Flags.S3Bucket)
+		s3Config := aws.NewConfig()
+
+		if Flags.S3Endpoint == "" {
+			stdout.Printf("Using 's3://%s' as S3 bucket for storage.\n", Flags.S3Bucket)
+		} else {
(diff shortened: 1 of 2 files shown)


Acconut:
Thank you for this contribution! I tested it locally and it worked without any issues, good job.

> an old version of embedded web server(civetweb) that doesn't support absolute URI.

Would you mind explaining where the problem here lies as I am not familiar with that software? Is there something on our end we could improve?

mrezai:
It seems URLs in HTTP requests that generated by aws sdk are absolute and those doesn't support by old versions of civetweb.
More details:
http://tracker.ceph.com/issues/17657
https://github.com/civetweb/civetweb/issues/197
https://github.com/aws/aws-sdk-go/issues/344

Acconut:
I appreciate the links you provided a lot, they helped me to understand the situation. I guess we cannot do anything regarding this in the tusd project as the maintainer of the AWS Go SDK don't seem to have the intention to change this (rightfully so).

PR #99, opened 2017-01-27
HTTP 423 can be replaced by http.StatusLocked

oliverpool:
See https://golang.org/src/net/http/status.go?h=StatusLocked#L55


Acconut:
I am aware of its existence but it has been introduced in Go 1.7 (https://github.com/golang/go/commit/b9ec0024fbc18dd94eff7240afd82fac6b4d8fdc#diff-5b4f73f3a34f9f2404e099c1ced7143eR55) while we try to support Go 1.5+. You can also see this in the failing tests: https://travis-ci.org/tus/tusd/builds/195779508

Thank you but I don't think this is a good idea right now.

oliverpool:
Ok, sorry about that :confused:

Acconut:
Don't apologize :) You did nothing wrong by trying

PR #104, opened 2017-01-31
Cleanup Frey file

tersmitten:


.infra/Freyfile.hcl
```diff
@@ -167,7 +182,6 @@ setup {
     }
     roles {
       role = "{{{init.paths.roles_dir}}}/fqdn/1.2.1"
-      fqdn = "{{lookup('env', 'FREY_DOMAIN')}}"
```
kvz: Same concern as mentioned here: https://github.com/transloadit/uppy-server/pull/20/files#r98854186,

> I'm afraid this will break the hostname as mentioned in Slack

tersmitten: I should not break since it's defined [here](https://github.com/tus/tusd/blob/master/.infra/group_vars/all/_frey.yml#L25) now, but I can give it another test run if you want to :-)

kvz: Nope, that will do! Thanks for clarifying, merging this

PR #105, opened 2017-02-01
Prevent race condition on errorsTotal increment

oliverpool:
*Original discussion: https://github.com/tus/tusd/commit/f50f03fe6f357b98ebef6a97833503e121931149#commitcomment-20637394*

This PR prevents a race condition on the errorsTotal counting.

The HTTPError interface is also used to have detailed metrics (by `status` as
well):
```
# HELP tusd_errors_total Total number of errors per cause.
# TYPE tusd_errors_total gauge
tusd_errors_total{cause="upload not found",status="404"} 2
```
This can be helpful when displaying some statistics (errors 5xx, 4xx...)


metrics.go
```diff
@@ -80,7 +73,48 @@ func newMetrics() Metrics {
...
+	ptr, ok := e.m[err]
+	e.RUnlock()
+	if !ok {
+		// The ptr does not seem to exist for this err
+		// Hence we create it (using a write lock)
+		e.Lock()
+		// We ensure that the ptr wasn't created in the meantime
+		if ptr, ok = e.m[err]; !ok {
+			ptr = new(uint64)
+			*ptr = 0
```
Acconut: Initializing a pointer to its default value after `new()` is not necessary. It will be automatically done, see the Go spec: https://golang.org/ref/spec#Allocation

> The variable is initialized as described in the section on initial values.

metrics.go
```diff
@@ -80,7 +73,47 @@ func newMetrics() Metrics {
...
+		e.Lock()
+		// We ensure that the ptr wasn't created in the meantime
+		if ptr, ok = e.m[err]; !ok {
+			ptr = new(uint64)
+			e.m[err] = ptr
+		}
+		e.Unlock()
+	}
+	// We can then increase the counter
+	atomic.AddUint64(ptr, 1)
```
oliverpool: Actually the `ptr` could be returned and the incrementation done by `incErrorsTotal`.
The signature could then be `func (e *ErrorsTotalMap) retrievePointerFor(err HTTPError) *uint64`.

It would separate the concerns `retrievePointerFor` does not care what the `ptr` is made for and `incErrorsTotal()` performs an incrementation.

Acconut: I like this idea :+1:

oliverpool:
@Acconut: done!

Acconut:
After testing your branch locally I discovered a flaw in how the errors are stored in the map. Here are steps to reproduce:

1. Start the tusd server in background: `go run cmd/tusd/main.go --dir data-dir &`
2. Remove the data directory: `rm -r ./data-dir`
3. Try to create multiple uploads by running following curl command a few times: ` curl -X POST -H "Tus-Resumable: 1.0.0" http://localhost:1080/files/ -H "Upload-Length: 100" -v`. These requests will fail because the upload directory does not exist any more but that is expected and wanted.
4. Visit http://localhost:1080/metrics to view the metrics but you see a following error:

```
An error has occurred during metrics collection:

3 error(s) occurred:
* collected metric tusd_errors_total label:<name:"cause" value:"upload directory does not exist: ./data" > label:<name:"status" value:"500" > gauge:<value:1 >  was collected before with the same name and label values
* collected metric tusd_errors_total label:<name:"cause" value:"upload directory does not exist: ./data" > label:<name:"status" value:"500" > gauge:<value:1 >  was collected before with the same name and label values
* collected metric tusd_errors_total label:<name:"cause" value:"upload directory does not exist: ./data" > label:<name:"status" value:"500" > gauge:<value:1 >  was collected before with the same name and label values
```

The problem is following:

In the datastores an error object may be created every time they occur (see https://github.com/tus/tusd/blob/master/filestore/filestore.go#L67). Even though they share the same message, they are different instances and have different addresses in memory. If you then add all these errors to your ErrorsTotalMap they will be considered as different keys (because HTTPError is an interface and the error instances have different addresses) and insert a new key-value pair. In the example above, the ErrorsTotalMap would contain three entries, all pointing to an "upload directory does not exist" error. These duplicates cause Prometheus to blow up, when you try to collect them (https://github.com/tus/tusd/pull/105/files#diff-362f42fe2d092bfd04057043da99f142R79) because they have the same error message and status code.

My suggested solution to this problem would be to revert the ErrorsTotalMap to a `map[string]*uint64` again. However, we would lose the status code label in the `/metrics` page (which I personally do not see as a huge sacrifice). Or you could make it a `map[struct{Message string, Status int}]*uint64`. The difference is now that the map keys are not interfaces but structures now.

oliverpool:
Godd catch!

Another solution might be to include the address of the error (`&err`) in the metrics to ensure its uniqueness for prometheus (but it will be ugly if a new object is created everytime).

I would prefer to implement your second suggestion (`map[struct{Message string, Status int}]*uint64`) if you are fine with it.

Acconut:
> I would prefer to implement your second suggestion (map[struct{Message string, Status int}]*uint64) if you are fine with it.

Of course, I would be delighted if you could take on this task :)

oliverpool:
@Acconut: implemented!

Regarding the conflict, I think [this part of the code](https://github.com/tus/tusd/commit/ca95bf275ab246ec4b2fa4e01c36ddc504d0515c#diff-4e64569f67c7c0214e072f52dbe8d504R608) should be moved inside the `simplifiedHTTPError` function (it may be interesting to have the original message in some other logs: only the metrics should be simplified):

```go
func simplifiedHTTPError(err HTTPError) simpleHTTPError {
	msg := err.Error()

	// Errors for read timeouts contain too much information which is not
	// necessary for us and makes grouping for the metrics harder. The error
	// message looks like: read tcp 127.0.0.1:1080->127.0.0.1:53673: i/o timeout
	// Therefore, we use a common error message for all of them.
	if netErr, ok := err.(net.Error); ok && netErr.Timeout() {
		msg = "read tcp: i/o timeout"
	}
	return simpleHTTPError{
		Msg:  msg,
		Code: err.StatusCode(),
	}
}
```

Acconut:
Good idea!

oliverpool:
Completed.

Acconut:
Great! I will give it a final review soon and then merge it!

prometheuscollector/prometheuscollector.go
```diff
@@ -46,11 +46,11 @@ var (
 )
 
 type Collector struct {
-	metrics tusd.Metrics
+	metrics *tusd.Metrics
```
Acconut: Good catch! However, I would rather address this issue at the root, meaning `newErrorsTotalMap` should return a pointer. This should have the same effect and would prevent the lock from being copied.

oliverpool: But the pointer here remains needed, right?

2017-02-27-221005_2560x1080_scrot

Acconut: If you are talking about the one in the Collector struct, then I don't think so.

oliverpool: With the code I just pushed (https://github.com/tus/tusd/pull/105/commits/13bc64517ba84741ad26ade44c1e8aa3ed451eba) I get the following output:
```
go vet ./prometheuscollector
prometheuscollector/prometheuscollector.go:54: New passes lock by value: github.com/tus/tusd.Metrics contains github.com/tus/tusd.ErrorsTotalMap
prometheuscollector/prometheuscollector.go:56: literal copies lock value from metrics: github.com/tus/tusd.Metrics contains github.com/tus/tusd.ErrorsTotalMap
prometheuscollector/prometheuscollector.go:60: Describe passes lock by value: prometheuscollector.Collector contains github.com/tus/tusd.Metrics contains github.com/tus/tusd.ErrorsTotalMap
prometheuscollector/prometheuscollector.go:69: Collect passes lock by value: prometheuscollector.Collector contains github.com/tus/tusd.Metrics contains github.com/tus/tusd.ErrorsTotalMap
prometheuscollector/prometheuscollector.go:115: Handler passes lock by value: github.com/tus/tusd.Metrics contains github.com/tus/tusd.ErrorsTotalMap
prometheuscollector/prometheuscollector.go:116: call of New copies lock value: github.com/tus/tusd.Metrics contains github.com/tus/tusd.ErrorsTotalMap
exit status 1
```

But it seems to be fine on [AppVeyor](https://ci.appveyor.com/project/Acconut/tusd/build/1.0.421)

oliverpool:
It still appear here, but it might be a local problem if neither you, nor travis, nor appveyor have the issue !

Acconut:
That's interesting. Which Go version are you running? They changed it strictness a bit in 1.8 (https://golang.org/doc/go1.8#tool_vet) but this actually be visible on Travis...

oliverpool:
Solved with `go get -u github.com/tus/tus` (the `tus` import from the `prometheuscollector` subpackage wasn't up to date)

Acconut:
Thank you for your help and in particular your patience with me :heart: :+1:  I have done some minor cleanups and then squash merged your branch in https://github.com/tus/tusd/commit/74fca8c6c3033b7a55735095359c6e42e7ae532f.

Issue #108, opened 2017-02-10
about temp file


aixj1984:
when upload one file, in the dirct of data genenrate two tempfile。 one file is lock file ， end of upload deleted，but other pid file not delete

Acconut:
I'm sorry but I have no idea what you actually mean. You are correct that a `.lock` file is there as long as an client is uploading but there should be no "pid file". Please be more descriptive.

aixj1984:
[
![qq 20170213091108](https://cloud.githubusercontent.com/assets/2499352/22868082/ae84f0ee-f1cc-11e6-90ca-77cea9769e63.png)
](url)   please check the picture，I give the  tag on the picture.

Acconut:
Ok, thanks for clarifying things. This file is indeed a temporary one used for creating the `.lock` file. What concerns you about it?

aixj1984:
why don't delete it after upload file?

I findthe temp file in the folder , i think the file must be deleted .

Acconut:
You are correct, usually it will be removed however, it may have issues on Windows. I can look into this once I have some time left, as this has looks like a low priority.

Acconut:
This should have been solved with #117. Feel free to comment if this issue is still arising for you.

PR #117, opened 2017-03-02
tmplock must be closed before removal

oliverpool:
`defer` are called in LIFO style, so `os.Remove` was called before `tmplock.Close` (which is obviously wrong).

Since this a dependency, I don't know how it should be fixed...

*(don't update to the latest version of https://github.com/nightlyone/lockfile: if the same process acquire a lock two times, it won't fail because the pid is the same)*

### DIFF
--- a/vendor/gopkg.in/Acconut/lockfile.v1/lockfile.go
+++ b/vendor/gopkg.in/Acconut/lockfile.v1/lockfile.go
@@ -82,8 +82,10 @@ func (l Lockfile) TryLock() error {
 	if err != nil {
 		return err
 	} else {
-		defer tmplock.Close()
-		defer os.Remove(tmplock.Name())
+		defer func(){
+			tmplock.Close()
+			os.Remove(tmplock.Name())
+		}()
 	}
 
 	_, err = tmplock.WriteString(fmt.Sprintf("%d\n", os.Getpid()))


Acconut:
Great find! This could explain #108 where this temporary file is not removed because there is still an open file handler.

> (don't update to the latest version of https://github.com/nightlyone/lockfile: if the same process acquire a lock two times, it won't fail because the pid is the same)

I am aware of this situation (see https://github.com/nightlyone/lockfile/pull/15#issuecomment-248988619) and that's why I forked the repository to revert this change. For now we are using my personal version of it which is located at https://github.com/Acconut/lockfile/tree/v1.0.0. If you could send this PR to this repository, I would be very thankful!

oliverpool:
> https://github.com/Acconut/lockfile/tree/v1.0.0. If you could send this PR to this repository, I would be very thankful!

But I can't make a PR on a tag (and your master is up to date with https://github.com/nightlyone/lockfile)

Acconut:
That's correct, please still open a PR against the master branch and I will take care of changing the behaviour of the master branch.

Acconut:
The dependency with the necessary changes has been updated in 7f26757a40b27166b773aec306e63533e4db6747. Thank you for noticing :)

PR #123, opened 2017-04-07
Allow override of default options response

matt-wratt:
We currently expose this as a service which runs on a different host to some javascript clients that preform uploads.  We require the clients to authenticate before allowing them to upload.  We use the Authorization header for this which needs to be included in the Access-Control-Allow-Headers of the options response.  With the current Middleware implementation headers set early are overwritten and when the headers are set after it's too late.

This is a very simple approach to allow the caller to override the headers specified in the options response.  Any thoughts or feedback would be appreciated.

### DIFF
--- a/unrouted_handler.go
+++ b/unrouted_handler.go
@@ -138,6 +138,15 @@ func NewUnroutedHandler(config Config) (*UnroutedHandler, error) {
 // directly you will need to wrap at least the POST and PATCH endpoints in
 // this middleware.
 func (handler *UnroutedHandler) Middleware(h http.Handler) http.Handler {
+	return handler.MiddlewareWithOptionsCallback(h, func(w http.ResponseWriter, r *http.Request) {})
+}
+
+// MiddlewareWithOptionsCallback checks various aspects of the request and
+// ensures that it conforms with the spec. Also handles method overriding for
+// clients which cannot make PATCH AND DELETE requests. If you are using the
+// tusd handlers directly you will need to wrap at least the POST and PATCH
+// endpoints in this middleware.
+func (handler *UnroutedHandler) MiddlewareWithOptionsCallback(h http.Handler, cb http.HandlerFunc) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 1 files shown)


Acconut:
Frankly, I am not a fan of this change as it is not scalable (if we want to extend to to other methods in the future) and makes the already complicated HTTP handling even more complex). I do understand why you need to modify the Access-Control-Allow-Headers value but I would tackle this problem from a different angle.

>  With the current Middleware implementation headers set early are overwritten and when the headers are set after it's too late.

What do you think if we - instead of overwriting the entire header value in Middleware - just append the tus-specific values to a previously set header value (if present)? This way you are able to influence the header.

matt-wratt:
> What do you think if we - instead of overwriting the entire header value in Middleware - just append the tus-specific values to a previously set header value (if present)? This way you are able to influence the header.

That works for me, I'll make some changes.

matt-wratt:
@Acconut I've made some changes, what do you think?  My implementation could be overkill, but it ensures unique comma separated values in the CORS headers.

matt-wratt:
FYI I'll follow up the failing test... it due to the order of the headers now being a bit random.

Acconut:
Thank you for your work but I am not sure if deduplication is that important for us, in particular as it makes this code harder to understand. We can also use the http.Header.Add function instead of http.Header.Set which will not overwrite any previously set values. Here's an example: https://play.golang.org/p/09UnV0ahdW. You may notice that the output produces two lines for the Accept header but the HTTP specification allows this and it's actually equivalent to `Accept: GET, POST` (see http://stackoverflow.com/questions/4371328/are-duplicate-http-response-headers-acceptable).

What do you think?

matt-wratt:
Haha - I did notice the multiple lines for a single header, which prompted me to go with a far more complicated approach... if multiple lines work then I can work with that.

Allow me to revert some changes.

Acconut:
Thank you, this landed in 5600cf2d6ed18c4fe87ac96661c3fab134461057 alongside with a short test case :)

matt-wratt:
Much appreciated @Acconut!

PR #129, opened 2017-05-22
working alpine Dockerfile usable for local and autobuild

flaneurtv:
This Dockerfile works quite nicely and produces an image of only 256 MB. 

Docker entrypoint is configured to make use of the dirs /srv/tusd-data and /srv/tusd-hooks. You can therefore mount an outside dir with your private hooks onto /srv/tusd-hooks and tusd will automatically use them. 

As the image is using ENTRYPOINT for the basic start command, it is possible to provide additional startup args to tusd by adding them to `docker run` after the image name:

### Example Docker run command
`docker run --name tusd -it -v $PWD/tusd-data/:/srv/tusd-data/ -v $PWD/tusd-hooks/:/srv/tusd-hooks -e AWS_REGION=eu-central-1 flaneurtv/tusd -s3-bucket tusd-bucket -s3-endpoint https://s3.eu-central-1.amazonaws.com`

The Dockerfile should work for both, local builds as well as Docker Hub automatic builds. However, judging from the way things are structured in most "official" Docker builds, a little tweaking will still be needed to satisfy Docker, while this is not more than an educated guess.

I am happy to provide further help on the path towards becoming a Docker official image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,6 @@ RUN addgroup -g 1000 tusd \
     && apk del git
 
 WORKDIR /srv/tusd-data
-VOLUME ["/srv/tusd-data"]
-VOLUME ["/srv/tusd-hooks"]
 EXPOSE 1080
 USER tusd
 ENTRYPOINT ["/go/bin/tusd","-dir","/srv/tusd-data","--hooks-dir","/srv/tusd-hooks"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,30 @@
+FROM golang:1.7-alpine
+
+# Copy in the git repo from the build context
+COPY . /go/src/github.com/tus/tusd/
+
+# Create app directory
+
+RUN addgroup -g 1000 tusd \
+    && adduser -u 1000 -G tusd -s /bin/sh -D tusd \
+    && cd /go/src/github.com/tus/tusd \
+    && apk add --no-cache \
+        git \
+    && go get -d -v ./... \
+    && version="$(git tag -l --points-at HEAD)" \
+    && commit=$(git log --format="%H" -n 1) \
+    && GOOS=linux GOARCH=amd64 go build \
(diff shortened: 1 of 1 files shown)


Acconut:
First of all, thank you for helping us with that. However, as a non-experienced Docker user, I have a few questions:

> produces an image of only 256 MB

Honestly, I am surprised that the size of the image is that big. Alpine and the tusd binary on it's own should be rather small and the git dependency is removed again. Do you have any clues?

> As the image is using ENTRYPOINT for the basic start command, it is possible to provide additional startup args to tusd by adding them to docker run after the image name:

Nice :+1:

> The Dockerfile should work for both, local builds as well as Docker Hub automatic builds. However, judging from the way things are structured in most "official" Docker builds, a little tweaking will still be needed to satisfy Docker, while this is not more than an educated guess.

As said before, I have never used Docker Hub. Could you provide a few points what would needed to be done to set this up? Could we also just run it from Travis CI and would not have to use another service?

Acconut:
> the image size is actually only 77MB and not 256MB, as you can see here:

Okay, sounds good enough.

> But probably it is a better idea, if I do set things up and add whoever you want to an organisation I create for tus on Docker Hub. What do you think?

That would be very helpful, indeed. Would you be also so kind and give us admin access to the organization?

> As for the pull request, can we go ahead with this or do you need anything changed?

I think we can merge this right now. :)

trickkiste:
> That would be very helpful, indeed. Would you be also so kind and give us admin access to the organization?

Of course, but you need to tell me a Docker Hub username, which I can delegate admin rights to, please.

kvz:
Please add me (`kevinvz`) also

kvz:
Is there a distinction between owners and admins? I'd be great if the tus core team could be assigning new admins when they needed

Acconut:
I am `acconut` on Docker Hub

kvz:
I just remembered about this and wondered what the status was of the automated container builds and an official `_/tusd`?

Acconut:
It's good that you remembered it. I quickly took care of it: https://hub.docker.com/r/tusproject/tusd/
As @trickkiste said, it was super easy. Again, thank you for your help :)

Issue #138, opened 2017-07-12
s3store strips ID value from .info file


trickkiste:
When I use filestore, I get something like this:
{"ID":"fa551c95d1a502ca78fc6227a70f5561","Size":86406,"Offset":0,"MetaData":{},"IsPartial":false,"IsFinal":false,"PartialUploads":null}

When I use s3store, the ID value is missing in the .info file:
{"ID":"","Size":22479312,"Offset":0,"MetaData":{},"IsPartial":false,"IsFinal":false,"PartialUploads":null}

Acconut:
Good catch. It should be possible to reorder the NewUpload function (https://github.com/tus/tusd/blob/master/s3store/s3store.go#L166) so that we know the upload ID when the .info file is created. Basically the calls to CreateMultipartUpload and PutObject should happen in the opposite order. I haven't confirmed this theory yet but I am pretty sure it would work.

trickkiste:
That was exactly my thinking as well. But as I was not sure, if there might be a reason for this order, I preferred to provide a more unobtrusive patch first.

Acconut:
Great to hear, are you willing to attempt this patch? I would be pleased to assist you :)

trickkiste:
Should be as easy as moving the following part downwards below `id = uploadId + "+" + *res.UploadId`, right? 

```
	// Create object on S3 containing information about the file
	_, err = store.Service.PutObject(&s3.PutObjectInput{
		Bucket:        aws.String(store.Bucket),
		Key:           aws.String(uploadId + ".info"),
		Body:          bytes.NewReader(infoJson),
		ContentLength: aws.Int64(int64(len(infoJson))),
	})
	if err != nil {
		return "", fmt.Errorf("s3store: unable to create info file:\n%s", err)
	}
```
Sure I would attempt that. Can provide the patch in a minute. Do you see any other obstacles?

Acconut:
> Should be as easy as moving the following part downwards below `id = uploadId + "+" + *res.UploadId`, right?

Yes, but please remember to assign `info.ID = id` before calling `PutObject`.

> Do you see any other obstacles?

Except for tests, I don't :)

trickkiste:
OK, I'll attempt this patch then.

As for the tests, copy that. I have not seen any tests explicitly testing
for the existence of the ID value in the .info file. Neither in filestore
nor in s3store. I have added some to filestore, but have not made a patch
yet. Nor have I figured out yet, how to test this with S3.

[...]

PR #140, opened 2017-07-12
Fix for bug #138

trickkiste:
Fixes bug #138 

Seems to fix the issue. 
However, I am relatively new to Golang and did not examine in full depth, if not adding the ID to the .info file is done on purpose and for some reason specific to the inner workings of S3. 

@Acconut, can you please verify that this fix does not have any unwanted side effects.

### DIFF
--- a/s3store/s3store.go
+++ b/s3store/s3store.go
@@ -167,6 +167,7 @@ func (store S3Store) NewUpload(info tusd.FileInfo) (id string, err error) {
 	var uploadId string
 	if info.ID == "" {
 		uploadId = uid.Uid()
+		info.ID = uploadId
 	} else {
 		uploadId = info.ID
 	}


Acconut:
As you correctly identified the ID is not stored in the .info for a reason. I prefer to not persist incorrect or incomplete information, and therefore I do not see your fix as insufficient. However, as I described in https://github.com/tus/tusd/issues/138#issuecomment-315558494, there should be another solution.

trickkiste:
With incomplete you mean, not having the multipart upload ID available at
that point, do I understand that correctly?

Mit besten Grüßen,
Markus Kienast

--

Markus Kienast - CEO

FLANEUR TV
<mark@sfe.tv>mark@f <mark@trickkiste.at>laneur.tv
+43 699 18084401

Commercial Register: FN 363780 x
UID: ATU66601545
WWW: http:// <http://trickkiste.at/>flaneur.tv

2017-07-15 22:02 GMT+02:00 Marius <notifications@github.com>:

Acconut:
> With incomplete you mean, not having the multipart upload ID available at that point, do I understand that correctly?

Yes, you are correct.

PR #142, opened 2017-07-16
fixed ID value in .info

trickkiste:
I changed the order of events in NewUpload to facilitate a correct assignment of ID in the .info file, as discussed yesterday.

I also adjusted the tests to the new order of events and changed some minor things, to make test fail messages more accurate.

One conceptual issue however remains:
Now the s3store upload ID follows the convention "uploadID+multipartID". This is true for tusd request headers, URLs, the ID value in .info, but not for the actual filename on S3. It is still only uploadID and uploadID.info.

The question is, should either the filenames follow the scheme "uploadID+multipartID" as well - which is possible with the procedures in NewUpload now reordered - or the uploadID alone should be used everywhere and in addition to that a multipartID field should be introduced in .info.

As I presume the tusd does derive the multipartID from the request URL on upload requests, I guess the former solution is the way to go. I am suggesting this change, because the current situation might confuse users.

### DIFF
--- a/s3store/s3store.go
+++ b/s3store/s3store.go
@@ -168,23 +168,8 @@ func (store S3Store) NewUpload(info tusd.FileInfo) (id string, err error) {
 	if info.ID == "" {
 		uploadId = uid.Uid()
 	} else {
+		// certain tests set info.ID in advance
 		uploadId = info.ID
-	}
-
-	infoJson, err := json.Marshal(info)
-	if err != nil {
-		return "", err
-	}
-
-	// Create object on S3 containing information about the file
-	_, err = store.Service.PutObject(&s3.PutObjectInput{
-		Bucket:        aws.String(store.Bucket),
-		Key:           aws.String(uploadId + ".info"),
-		Body:          bytes.NewReader(infoJson),
-		ContentLength: aws.Int64(int64(len(infoJson))),
-	})
-	if err != nil {
-		return "", fmt.Errorf("s3store: unable to create info file:\n%s", err)
 	}
 
 	// Convert meta data into a map of pointers for AWS Go SDK, sigh.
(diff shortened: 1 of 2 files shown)


trickkiste:
Fixes bug #138

Acconut:
> The question is, should either the filenames follow the scheme "uploadID+multipartID" as well - which is possible with the procedures in NewUpload now reordered

It is possible but not easily doable. Basically, when creating a multipart upload you need to specify under which name the file should be saved in the end - uploadID in our case. In return, you get the corresponding multipart ID which you cannot define prior. Because of this cyclic problem you cannot use the multipart ID in the file name.

Furthermore, this would break existing application which depend on the files being named after the upload ID.

> uploadID alone should be used everywhere and in addition to that a multipartID field should be introduced in .info.

This would require reading the .info file for a number of operations which need to know the multipart ID, e.g. GetReader, Terminate, FinishUpload or ConcatUploads. Such a change would not benefit the overall performance.

> As I presume the tusd does derive the multipartID from the request URL on upload requests, I guess the former solution is the way to go. I am suggesting this change, because the current situation might confuse users.

I agree, but the current approach also has an advantage, as mentioned above. Thank you for bringing up this topic, of course!

trickkiste:
@Acconut, the patch should be ready for integration now. If you agree, I kindly ask to accept the pull request.

Acconut:
Merged in 028ba572063b2a5bcbfd9ceb4ceb808cf379af42 after I removed the manual ID assignment in GetInfo. Thank you again :)

trickkiste:
Thank you for the finishing touch.

Mit besten Grüßen,
Markus Kienast

--

Markus Kienast - CEO

FLANEUR TV
<mark@sfe.tv>mark@f <mark@trickkiste.at>laneur.tv
+43 699 18084401

Commercial Register: FN 363780 x
UID: ATU66601545
WWW: http:// <http://trickkiste.at/>flaneur.tv

2017-07-19 11:56 GMT+02:00 Marius <notifications@github.com>:

PR #146, opened 2017-07-27
Update Uid() func to v4 Unique id.

bienkma:
Reduce Collisions and Standard RFC 4122

### DIFF
--- a/uid/uid.go
+++ b/uid/uid.go
@@ -19,5 +19,11 @@ func Uid() string {
 		// for random bits.
 		panic(err)
 	}
+	// UUID version 4
+	id[6] = (id[6] & 0x0f) | (4<<4)
+
+	// SetVariant sets variant bits as described in RFC 4122.
+	id[8] = (id[8] & 0xbf | 0x80)
+	
 	return hex.EncodeToString(id)
 }


Acconut:
Do you mind explaining your thoughts behind this change?

bienkma:
The uuidv4 description https://en.wikipedia.org/wiki/Universally_unique_identifier#Random_UUID_probability_of_duplicates section uuidv4. "Some pseudorandom number generators lack necessary entropy to produce sufficiently pseudorandom numbers. For example, the WinAPI GUID generator, which uses a pseudorandom number generator, has been shown to produce UUIDs which follow a predictable pattern.[citation needed] RFC 4122 advises that "distributed applications generating UUIDs at a variety of hosts must be willing to rely on the random number source at all hosts. If this is not feasible, the namespace variant should be used."

Acconut:
Thank you for this citation but I don't think this is necessary in our case. The Uid function uses the crypto/rand package which implements "a cryptographically secure pseudorandom number generator" (https://golang.org/pkg/crypto/rand/) for generating those UIDs. Therefore, I do not believe that it lacks the "necessary entropy to produce sufficiently pseudorandom numbers". What do you think?

Acconut:
Closing this issue due to a lack of information. Feel free to leave a comment if you want to continue the discussion :)

Issue #147, opened 2017-08-04
AWS S3 ListParts API call returns max 1000 parts


koenvo:
**problem**
The ListParts API of S3 returns at most 1000 parts in a single call. Together with the MaxPartSize (hardcoded) of 6MB the maximum file size that I can upload to s3 is 6GB. When the "limit" of 1000 parts is reached the GetInfo method will return incorrect information causing ErrMismatchOffset errors.

**possible solution**
The ListParts call should be done in a loop and do additional calls while IsTruncated is set.  I'm willing to create a PR for this, altho I never coded in go before. Let's see :-)
It would also be nice to be able to change MaxPartSize. Is there a reason why 6MB is chosen?

Does anyone know a temporary solution? I prefer not build tusd from source.

trickkiste:
Hi @koenvo, I just ran into the exact same problem. Most of my uploads will be larger than 6GB, so this is crucial for me too.

According to http://docs.aws.amazon.com/AmazonS3/latest/dev/qfacts.html there is a hard limit of 10.000 parts in AWS S3 with 1000 max-parts per list request, as you described above.

I am wondering, how far you got in your investigation. I have just had a quick look at tusd code concerning that matter, but I presume the parts > 1000 are actually created but are not included in the s3.CompleteMultipartUploadInput response. Or do thing go wrong even earlier, e.g. do the parts uploaded past number 1000 receive the wrong content due to incorrect offset calculation.

My insights into the matter are still incomplete, I just sat down to investigate myself, but I wanted you to know, that I will also be working to resolve this and am hoping, that we can collaborate on this.

When you are reading this, please try to contact me in Slack, so we can coordinate our efforts.

My best regards,
Markus

koenvo:
hi @trickkiste,

Thanks for your message. I managed to create some go code including tests. Please have a look at PR #148 . 

Due to the hard limit of 10.000 parts the files are now limited to 60GB. In my case this would be enough. Another change could be to make the MaxPartSize configurable but I didn't look into this yet.

[slack-noob]where can i find you at slack?[/slack-noob]

Kind regard,
Koen

trickkiste:
Also, it seems to me, that having 
> MaxPartSize: 6 * 1024 * 1024,
> MinPartSize: 5 * 1024 * 1024,

hardcoded will not work for many large file scenarios, even if the 1000 max-parts limit for the s3.ListPartsInput response is dealt with correctly. 

[...]

PR #148, opened 2017-08-06
Take IsTruncated field of S3 ListParts API response into account

koenvo:
Fix for issue #147

### DIFF
--- a/s3store/s3store.go
+++ b/s3store/s3store.go
@@ -226,7 +226,7 @@ func (store S3Store) WriteChunk(id string, offset int64, src io.Reader) (int64, 
 	bytesUploaded := int64(0)
 
 	// Get number of parts to generate next number
-	parts, err := store.ListAllParts(id)
+	parts, err := store.listAllParts(id)
 	if err != nil {
 		return 0, err
 	}
@@ -303,7 +303,7 @@ func (store S3Store) GetInfo(id string) (info tusd.FileInfo, err error) {
 	}
 
 	// Get uploaded parts and their offset
-	parts, err := store.ListAllParts(id)
+	parts, err := store.listAllParts(id)
 	if err != nil {
 		// Check if the error is caused by the upload not being found. This happens
 		// when the multipart upload has already been completed or aborted. Since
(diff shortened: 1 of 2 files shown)


trickkiste:
There is actually a ListPartsPages function in the API, which does the iterating for you. You got a callback function, which is executed for each page and can be used to append the page contents to a parts array for example. Should simplify the code.

https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#S3.ListPartsPages

Acconut:
As @trickkiste said (https://github.com/tus/tusd/issues/147#issuecomment-322761255), the ListPartsPages function offers a more elegant code but is very tough to test with our current setup. Therefore we came to the conclusion that ListPartsPages in not optimal. However, this does not affect your patch, @koenvo.

Going to have a deep look into this PR now.

trickkiste:
@koenvo would you care to rename you ListParts method to ListAllParts to be more descriptive and differentiate it from the S3API method.

koenvo:
Done :-)

s3store/s3store.go
```diff
@@ -517,6 +501,33 @@ func (store S3Store) ConcatUploads(dest string, partialUploads []string) error {
 	return store.FinishUpload(dest)
 }
 
+func (store S3Store)ListAllParts(id string) (parts []*s3.Part, err error) {
```
Acconut: Please rename this function to `listAllParts`. The lower-cased letter will cause this function to not be available to other packages which is what we want from a private function.

s3store/s3store_test.go
```diff
@@ -240,9 +292,11 @@ func TestFinishUpload(t *testing.T) {
...
 	gomock.InOrder(
 		s3obj.EXPECT().ListParts(&s3.ListPartsInput{
-			Bucket:   aws.String("bucket"),
-			Key:      aws.String("uploadId"),
-			UploadId: aws.String("multipartId"),
+			Bucket:           aws.String("bucket"),
+			Key:              aws.String("uploadId"),
+			UploadId:         aws.String("multipartId"),
+			PartNumberMarker: aws.Int64(0),
+
```
Acconut: Please remove this empty line.

s3store/s3store.go
koenvo: Is there a way to keep `TestListParts` working (which should be renamed to `TestListAllParts` anyway)? Otherwise I need to test `listAllParts` through some of the other functions..

Acconut: No, we would need to test it through GetInfo and FinishUpload since the test cases are in a separate package.

koenvo: I updated both tests.

Acconut:
Thank you very much again :)

Issue #149, opened 2017-08-16
s3store MaxPartSize limits possible upload size way below S3 specs of 5 TB max


trickkiste:
AWS S3 specifies a maximum object size of 5TB. This object size can not be handled by tusd, as MaxPartSize is not calculated according to the file size of the upload, but has a hard limit way below the limit AWS specifies in their documentation.

Tusd's s3store sets this:
> MaxPartSize: 6 * 1024 * 1024,
> MinPartSize: 5 * 1024 * 1024,

AWS Specs states:
http://docs.aws.amazon.com/AmazonS3/latest/dev/qfacts.html
> Maximum object size | 5 TB
> Maximum number of parts per upload | 10,000
> Part size | 5 MB to 5 GB, last part can be < 5 MB

The maximum file size with 10,000 possible parts would be 60 GB with the current settings and the restriction to 10.000 parts max. As a side note it should be mentioned, that at the time of writing, anything greater than 1000 parts will currently not work due to bug #147, effectively restricting max upload size to 6 GB.

From my understanding, `MaxPartSize` needs to be `5 * 1024 * 1024 * 1024` and the `EffectivePartSize` needs to be calculated from the actual upload size taking into account the restraints set by the S3 specs.

`EffectivePartSize` should then be calculated as follows:
```
if (uploadSize <= MinPartSize)
  upload in one piece
elseif (uploadSize/10.000 <= MinPartSize)
  upload with EffectivePartSize = MinPartSize
elseif (uploadSize <= 5TB)
  EffectivePartSize = uploadSize/10.000
else
refuse to comply
```

As a prerequisite bug #147 needs to be fixed of course.

trickkiste:
The solution I am proposing would still assure that the EffectivePartSize will be set as small as possible, while at the same time enable uploads to the upper end of the S3 specs size limit.

Therefore the user would get the best of both:
* Minimal re-upload size if an upload needs to resume
* And max upload size of 5 TB total

Acconut:
This is indeed a limiting factor. Another possible solution (which is used by the data store for Google Cloud Storage) is to merge the upload recursively. This would work like following:

1. Upload parts 1 to 10,000 to S3
2. Merge the chunks 1 to 10,000 into a temporary object
3. Create a new multipart upload on S3
4. Copy the temporary object in a new part using UploadPartCopy (http://docs.aws.amazon.com/sdk-for-go/api/service/s3/#S3.UploadPartCopy)
[...]

PR #150, opened 2017-08-17
calculating PartSize based on size of upload

trickkiste:
This is a fix for bug #149.

It has been applied to latest commit in koenvo:s3-large-file-upload-fix branch, not to master and therefore shall be merged after the former has been merged successfully.

So this pull request might be premature, but I wanted to give you a chance to review it early and hope to get some feedback, to get it merged quickly.

My best regards,
Markus

### DIFF
--- a/s3store/s3store.go
+++ b/s3store/s3store.go
@@ -86,6 +86,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math"
 	"os"
 	"regexp"
 	"strings"
@@ -128,6 +129,13 @@ type S3Store struct {
 	// in bytes. This number needs to match with the underlying S3 backend or else
 	// uploaded parts will be reject. AWS S3, for example, uses 5MB for this value.
 	MinPartSize int64
+	// MaxMultipartParts is the maximum number of parts an S3 multipart upload is
+	// allowed to have according to AWS S3 API specifications.
+	// See: http://docs.aws.amazon.com/AmazonS3/latest/dev/qfacts.html
+	MaxMultipartParts int64
+	// MaxObjectSize is the maximum size an S3 Object can have according to S3
+	// API specifications. See link above.
+	MaxObjectSize int64
 }
 
 type S3API interface {
@@ -146,10 +154,12 @@ type S3API interface {
 // The MaxPartSize and MinPartSize properties are set to 6 and 5MB.
 func New(bucket string, service S3API) S3Store {
 	return S3Store{
-		Bucket:      bucket,
-		Service:     service,
-		MaxPartSize: 6 * 1024 * 1024,
-		MinPartSize: 5 * 1024 * 1024,
+		Bucket:            bucket,
+		Service:           service,
+		MaxPartSize:       5 * 1024 * 1024 * 1024,
+		MinPartSize:       5 * 1024 * 1024,
+		MaxMultipartParts: 10000,
+		MaxObjectSize:     5 * 1024 * 1024 * 1024 * 1024,
 	}
 }
 
@@ -224,6 +234,10 @@ func (store S3Store) WriteChunk(id string, offset int64, src io.Reader) (int64, 
 
 	size := info.Size
 	bytesUploaded := int64(0)
+	optimalPartSize := store.CalcOptimalPartSize(&size)
+	if optimalPartSize == 0 {
+		return bytesUploaded, nil
+	}
 
 	// Get number of parts to generate next number
 	parts, err := store.listAllParts(id)
@@ -243,7 +257,7 @@ func (store S3Store) WriteChunk(id string, offset int64, src io.Reader) (int64, 
 		defer os.Remove(file.Name())
 		defer file.Close()
 
-		limitedReader := io.LimitReader(src, store.MaxPartSize)
+		limitedReader := io.LimitReader(src, optimalPartSize)
 		n, err := io.Copy(file, limitedReader)
 		// io.Copy does not return io.EOF, so we not have to handle it differently.
 		if err != nil {
@@ -254,11 +268,11 @@ func (store S3Store) WriteChunk(id string, offset int64, src io.Reader) (int64, 
 			return bytesUploaded, nil
 		}
 
-		if (size - offset) <= store.MinPartSize {
+		if (size - offset) <= optimalPartSize {
 			if (size - offset) != n {
 				return bytesUploaded, nil
 			}
-		} else if n < store.MinPartSize {
+		} else if n < optimalPartSize {
 			return bytesUploaded, nil
 		}
 
@@ -546,4 +560,29 @@ func isAwsError(err error, code string) bool {
 		return true
 	}
 	return false
+}
+
+func (store S3Store) CalcOptimalPartSize(size *int64) int64 {
+	switch {
+	// We can only manage files up to MaxObjectSize, else we need to fail.
+	case *size > store.MaxObjectSize:
+		return 0
+	// When upload is smaller or equal MinPartSize, we upload in just one part.
+	case *size <= store.MinPartSize:
+		return store.MinPartSize
+	// When we need 9999 parts or less with MinPartSize.
+	case *size/store.MinPartSize < store.MaxMultipartParts:
+		return store.MinPartSize
+	// When we can divide our upload into exactly MaxMultipartParts parts with
+	// no bytes leftover, we will not need an spare last part.
+	// Also, when MaxObjectSize is equal to MaxPartSize * MaxMultipartParts
+	// (which is not the case with the current AWS S3 API specification, but
+	// might be in the future or with other S3-aware stores), we need this in
+	// order for our Multipart-Upload to reach full MaxObjectSize.
+	case int64(math.Mod(float64(*size), float64(store.MaxMultipartParts))) == 0:
+		return *size / store.MaxMultipartParts
+	// In all other cases, we need a spare last piece for the remaining bytes.
+	default:
+		return *size / (store.MaxMultipartParts - 1)
+	}
 }
--- a/s3store/s3store.go
+++ b/s3store/s3store.go
@@ -565,27 +565,33 @@ func (store S3Store) CalcOptimalPartSize(size int64) (int64, error) {
 	switch {
 	// We can only manage files up to MaxObjectSize, else we need to fail.
 	case size > store.MaxObjectSize:
+		fmt.Print("0 ")
 		return 0, fmt.Errorf("CalcOptimalPartSize: size of %v bytes exceeds MaxObjectSize of %v bytes", size, store.MaxObjectSize)
 	// When upload is smaller or equal MinPartSize, we upload in just one part.
 	case size <= store.MinPartSize:
+		fmt.Print("A ")
 		return store.MinPartSize, nil
 	// When we need 9999 parts or less with MinPartSize.
 	case size/store.MinPartSize < store.MaxMultipartParts:
+		fmt.Print("B ")
 		return store.MinPartSize, nil
-	// When we can divide our upload into exactly MaxMultipartParts parts with
-	// no bytes leftover, we will not need an spare last part.
-	// Also, when MaxObjectSize is equal to MaxPartSize * MaxMultipartParts
+	// If our upload divides up exactly into MaxMultipartParts parts with
+	// no bytes leftover, we will not need an spare last part. So we can go with
+	// a straight division. Otherwise we would exceed MaxPartSize in a scenario,
+	// where MaxObjectSize is equal to MaxPartSize * MaxMultipartParts
 	// (which is not the case with the current AWS S3 API specification, but
-	// might be in the future or with other S3-aware stores), we need this in
-	// order for our Multipart-Upload to reach full MaxObjectSize.
+	// might be in the future or with other S3-aware stores).
 	case size%store.MaxMultipartParts == 0:
+		fmt.Print("C ")
 		return size / store.MaxMultipartParts, nil
-	// If the last part would be larger than MaxPartSize, which is only the case
-	// when we are close to MaxObjectSize, we have to go with MaxPartSize.
-	case size%(store.MaxMultipartParts-1) > store.MaxPartSize:
-		return store.MaxPartSize, nil
-	// In all other cases, we need a spare last piece for the remaining bytes.
+	// In all other cases, we need to round up to the next integer, as long as we
+	// stay below MaxPartSize.
+	case size/store.MaxMultipartParts < store.MaxPartSize:
+		fmt.Print("D ")
+		return size/store.MaxMultipartParts + 1, nil
+	// If non of the above matches, we have exceeded the MaxPartSize limit above.
 	default:
-		return size / (store.MaxMultipartParts - 1), nil
+		fmt.Print("X ")
+		return size/store.MaxMultipartParts + 1, fmt.Errorf("CalcOptimalPartSize: to upload %v bytes optimalPartSize %v must exceed MaxPartSize %v", size, (size/store.MaxMultipartParts + 1), store.MaxPartSize)
 	}
 }
--- a/s3store/s3store.go
+++ b/s3store/s3store.go
@@ -86,7 +86,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"math"
 	"os"
 	"regexp"
 	"strings"
@@ -234,8 +233,8 @@ func (store S3Store) WriteChunk(id string, offset int64, src io.Reader) (int64, 
 
 	size := info.Size
 	bytesUploaded := int64(0)
-	optimalPartSize := store.CalcOptimalPartSize(&size)
-	if optimalPartSize == 0 {
+	optimalPartSize, err := store.CalcOptimalPartSize(size)
+	if err != nil {
 		return bytesUploaded, nil
 	}
 
@@ -562,27 +561,31 @@ func isAwsError(err error, code string) bool {
 	return false
 }
 
-func (store S3Store) CalcOptimalPartSize(size *int64) int64 {
+func (store S3Store) CalcOptimalPartSize(size int64) (int64, error) {
 	switch {
 	// We can only manage files up to MaxObjectSize, else we need to fail.
-	case *size > store.MaxObjectSize:
-		return 0
+	case size > store.MaxObjectSize:
+		return 0, fmt.Errorf("CalcOptimalPartSize: size of %v bytes exceeds MaxObjectSize of %v bytes", size, store.MaxObjectSize)
 	// When upload is smaller or equal MinPartSize, we upload in just one part.
-	case *size <= store.MinPartSize:
-		return store.MinPartSize
+	case size <= store.MinPartSize:
+		return store.MinPartSize, nil
 	// When we need 9999 parts or less with MinPartSize.
-	case *size/store.MinPartSize < store.MaxMultipartParts:
-		return store.MinPartSize
+	case size/store.MinPartSize < store.MaxMultipartParts:
+		return store.MinPartSize, nil
 	// When we can divide our upload into exactly MaxMultipartParts parts with
 	// no bytes leftover, we will not need an spare last part.
 	// Also, when MaxObjectSize is equal to MaxPartSize * MaxMultipartParts
 	// (which is not the case with the current AWS S3 API specification, but
 	// might be in the future or with other S3-aware stores), we need this in
 	// order for our Multipart-Upload to reach full MaxObjectSize.
-	case int64(math.Mod(float64(*size), float64(store.MaxMultipartParts))) == 0:
-		return *size / store.MaxMultipartParts
+	case size%store.MaxMultipartParts == 0:
+		return size / store.MaxMultipartParts, nil
+	// If the last part would be larger than MaxPartSize, which is only the case
+	// when we are close to MaxObjectSize, we have to go with MaxPartSize.
+	case size%(store.MaxMultipartParts-1) > store.MaxPartSize:
+		return store.MaxPartSize, nil
 	// In all other cases, we need a spare last piece for the remaining bytes.
 	default:
-		return *size / (store.MaxMultipartParts - 1)
+		return size / (store.MaxMultipartParts - 1), nil
 	}
 }
--- a/s3store/s3store.go
+++ b/s3store/s3store.go
@@ -173,6 +173,11 @@ func (store S3Store) UseIn(composer *tusd.StoreComposer) {
 }
 
 func (store S3Store) NewUpload(info tusd.FileInfo) (id string, err error) {
+	// an upload larger than MaxObjectSize must throw an error
+	if info.Size > store.MaxObjectSize {
+		return "", fmt.Errorf("s3store: upload size of %v bytes exceeds MaxObjectSize of %v bytes", info.Size, store.MaxObjectSize)
+	}
+
 	var uploadId string
 	if info.ID == "" {
 		uploadId = uid.Uid()
@@ -562,11 +567,6 @@ func isAwsError(err error, code string) bool {
 }
 
 func (store S3Store) CalcOptimalPartSize(size int64) (optimalPartSize int64, err error) {
-	// an upload larger than MaxObjectSize must throw an error
-	if size > store.MaxObjectSize {
-		return 0, fmt.Errorf("CalcOptimalPartSize: upload size of %v bytes exceeds MaxObjectSize of %v bytes", size, store.MaxObjectSize)
-	}
-
 	switch {
 	// When upload is smaller or equal MinPartSize, we upload in just one part.
 	case size <= store.MinPartSize:
@@ -607,6 +607,5 @@ func (store S3Store) CalcOptimalPartSize(size int64) (optimalPartSize int64, err
 	if optimalPartSize > store.MaxPartSize {
 		return optimalPartSize, fmt.Errorf("CalcOptimalPartSize: to upload %v bytes optimalPartSize %v must exceed MaxPartSize %v", size, optimalPartSize, store.MaxPartSize)
 	}
-
 	return optimalPartSize, nil
 }
--- a/s3store/s3store.go
+++ b/s3store/s3store.go
@@ -235,7 +235,7 @@ func (store S3Store) WriteChunk(id string, offset int64, src io.Reader) (int64, 
 	bytesUploaded := int64(0)
 	optimalPartSize, err := store.CalcOptimalPartSize(size)
 	if err != nil {
-		return bytesUploaded, nil
+		return bytesUploaded, err
 	}
 
 	// Get number of parts to generate next number
@@ -561,20 +561,16 @@ func isAwsError(err error, code string) bool {
 	return false
 }
 
-func (store S3Store) CalcOptimalPartSize(size int64) (int64, error) {
+func (store S3Store) CalcOptimalPartSize(size int64) (optimalPartSize int64, err error) {
+	var errorstrings []string
+
 	switch {
-	// We can only manage files up to MaxObjectSize, else we need to fail.
-	case size > store.MaxObjectSize:
-		fmt.Print("0 ")
-		return 0, fmt.Errorf("CalcOptimalPartSize: size of %v bytes exceeds MaxObjectSize of %v bytes", size, store.MaxObjectSize)
 	// When upload is smaller or equal MinPartSize, we upload in just one part.
 	case size <= store.MinPartSize:
-		fmt.Print("A ")
-		return store.MinPartSize, nil
+		optimalPartSize = store.MinPartSize
 	// When we need 9999 parts or less with MinPartSize.
 	case size/store.MinPartSize < store.MaxMultipartParts:
-		fmt.Print("B ")
-		return store.MinPartSize, nil
+		optimalPartSize = store.MinPartSize
 	// If our upload divides up exactly into MaxMultipartParts parts with
 	// no bytes leftover, we will not need an spare last part. So we can go with
 	// a straight division. Otherwise we would exceed MaxPartSize in a scenario,
@@ -582,16 +578,22 @@ func (store S3Store) CalcOptimalPartSize(size int64) (int64, error) {
 	// (which is not the case with the current AWS S3 API specification, but
 	// might be in the future or with other S3-aware stores).
 	case size%store.MaxMultipartParts == 0:
-		fmt.Print("C ")
-		return size / store.MaxMultipartParts, nil
-	// In all other cases, we need to round up to the next integer, as long as we
-	// stay below MaxPartSize.
-	case size/store.MaxMultipartParts < store.MaxPartSize:
-		fmt.Print("D ")
-		return size/store.MaxMultipartParts + 1, nil
-	// If non of the above matches, we have exceeded the MaxPartSize limit above.
+		optimalPartSize = size / store.MaxMultipartParts
+	// In all other cases, we need to round up to the next integer.
 	default:
-		fmt.Print("X ")
-		return size/store.MaxMultipartParts + 1, fmt.Errorf("CalcOptimalPartSize: to upload %v bytes optimalPartSize %v must exceed MaxPartSize %v", size, (size/store.MaxMultipartParts + 1), store.MaxPartSize)
+		optimalPartSize = size/store.MaxMultipartParts + 1
 	}
+
+	// an upload larger than MaxObjectSize must throw an error
+	if size > store.MaxObjectSize {
+		errorstrings = append(errorstrings, fmt.Sprintf("CalcOptimalPartSize: upload size of %v bytes exceeds MaxObjectSize of %v bytes", size, store.MaxObjectSize))
+	}
+	// optimalPartSize must never exceed MaxPartSize
+	if optimalPartSize > store.MaxPartSize {
+		errorstrings = append(errorstrings, fmt.Sprintf("CalcOptimalPartSize: to upload %v bytes optimalPartSize %v must exceed MaxPartSize %v", size, optimalPartSize, store.MaxPartSize))
+	}
+	if len(errorstrings) > 0 {
+		err = fmt.Errorf(strings.Join(errorstrings, "\n"))
+	}
+	return optimalPartSize, err
 }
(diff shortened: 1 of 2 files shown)


s3store/s3store.go
```diff
@@ -536,3 +549,21 @@ func isAwsError(err error, code string) bool {
 	}
 	return false
 }
+
+func (store S3Store) CalcPartSize(size *int64) (effectivePartSize int64) {
```
Acconut: Any reason why you decided to use a pointer here?

s3store/s3store.go
```diff
@@ -259,11 +256,11 @@ func (store S3Store) WriteChunk(id string, offset int64, src io.Reader) (int64,
 			return bytesUploaded, nil
 		}
 
-		if (size - offset) <= store.MinPartSize {
+		if (size - offset) <= effectivePartSize {
```
Acconut: Are you sure that **Min**PartSize has been replaced here correctly?

s3store/s3store.go
```diff
@@ -259,11 +256,11 @@ func (store S3Store) WriteChunk(id string, offset int64, src io.Reader) (int64,
...
 			if (size - offset) != n {
 				return bytesUploaded, nil
 			}
-		} else if n < store.MinPartSize {
+		} else if n < effectivePartSize {
```
Acconut: Are you sure that **Min**PartSize has been replaced here correctly?

s3store/s3store.go
```diff
@@ -536,3 +549,21 @@ func isAwsError(err error, code string) bool {
...
+	case *size/10000 < store.MinPartSize:
+		effectivePartSize = store.MinPartSize
+	case int64(math.Mod(float64(*size), 10000)) == 0:
+		effectivePartSize = *size / 10000
+	case *size/9999 >= store.MaxPartSize:
+		effectivePartSize = store.MaxPartSize
+	default:
+		effectivePartSize = *size / 9999
+	}
+	return effectivePartSize
```
Acconut: I am sorry but I am lost with this switch case here. Could you explain why you ditched your proposed algorithm from https://github.com/tus/tusd/issues/149?

s3store/s3store.go
trickkiste: With this new approach, we are calculating an optimal effectivePartSize, which needs to respect the MinPartSize and MaxPartSize limits set in the S3 API specification and take into account absolute max-parts limit of 10.000 parts. However, after the point of calculation, the min and max values are not relevant for our transfer anymore.

Each part we upload will need to align to effectivePartSize except for the last one, which can and will in most cases be lower than effectivePartSize and can also be lower than MinPartSize. Still, whether it is actually lower than MinPartSize does not have any relevance anymore.

However, the parts uploaded prior to the last one, must never be smaller than effectivePartSize, as we either would have to increase the size of another part to compensate for that, or we would need to upload an additional part in order to get everything transferred within the absolute max-parts limit of 10.000 parts. Otherwise we would have to exceed the absolute parts-max limit of 10.000 parts by at least one.

So after the point of calculation of the optimal part size (to suffice all restrictions in S3 API), MinPartSize and MaxPartSize loose all their relevance, as effectivePartSize does respect these limits anyways.

s3store/s3store.go
trickkiste: Same as above.

s3store/s3store.go
```diff
@@ -148,7 +149,7 @@ func New(bucket string, service S3API) S3Store {
 	return S3Store{
 		Bucket:      bucket,
 		Service:     service,
-		MaxPartSize: 6 * 1024 * 1024,
+		MaxPartSize: 5 * 1024 * 1024 * 1024,
```
koenvo: This should be `525 * 1024 * 1024`. (5TB / 10.000 parts)

s3store/s3store.go
trickkiste: The proposed "algorithm" was rather a rough human language description of what needs to be implemented, than an actual algorithm. The switch statement here does implement that outline and I have tested it thoroughly. The test case does test all relevant corner cases. But I have run the code for all sizes from MaxPartSize*9999-1 upwards.

s3store/s3store.go
trickkiste: Efficiency. Mainly for my "testing the code for ALL possible file sizes from 1 Byte to 5 TB" experiment.

s3store/s3store.go
trickkiste: @koenvo good point!

While I would propose to solve that differently, as going your way causes confusion, since the S3 API spec states a MaxPartSize of "5 * 1024 * 1024 * 1024" and new contributors will not intuitively understand, why that is different here. 

I would therefore propose to add two other variables here and adapt calcPartSize take them into account, which - as you correctly found out - is not the case now as far as MaxUploadSize goes.

```
MaxUploadParts: 10000
 MaxUploadSize: 5 * 1024 * 1024 * 1024 * 1024
```

trickkiste: I would also adapt the tests to test for this limit, so mistakes like that one can not happen anymore.

s3store/s3store.go
trickkiste: I will add some documentation of what is actually going on there.

s3store/s3store.go
trickkiste: If you don't like it, we can drop it.
If there is a good reason to refrain from it, then please elaborate, as I am always eager to learn something new!

s3store/s3store.go
trickkiste: I have improved on the algorithm, simplified it as much as possible and made sure it will still work, even if Amazon changes current limits in S3 API spec (or if other implementations might use different values).

I have adapted existing tests and improved my own testcase. `go fmt ./s3store` has been run on the code. The patch is also based on the current master which already includes koenvo's patch.

Please let me know, if you need anything else done on this. I am eager to get this merged.

trickkiste:
@Acconut I have taken all comments into consideration and came up with an improved implementation. The fact, that in case of interruption the last part needs to be re-transmitted, remains. But that was the case before already.

However, I will keep thinking about this problem and there might be a way to fix this, somewhere down the direction you proposed earlier. But it is going to be a mayor rewrite and until then, I suppose it is best to go with the current patch, so that we have a system, able to successfully handle files up to MaxObjectSize, until we come up with a better way to do the same.

s3store/s3store.go
```diff
@@ -547,3 +561,28 @@ func isAwsError(err error, code string) bool {
 	}
 	return false
 }
+
+func (store S3Store) CalcOptimalPartSize(size *int64) int64 {
```
Acconut: Any reason why you are using a pointer as the argument here instead of just the scalar `int64`?

s3store/s3store.go
```diff
@@ -547,3 +561,28 @@ func isAwsError(err error, code string) bool {
...
+	switch {
+	// We can only manage files up to MaxObjectSize, else we need to fail.
+	case *size > store.MaxObjectSize:
+		return 0
```
Acconut: In such a case, an explicit error should be returned.

s3store/s3store.go
```diff
@@ -547,3 +561,28 @@ func isAwsError(err error, code string) bool {
...
+	// When we need 9999 parts or less with MinPartSize.
+	case *size/store.MinPartSize < store.MaxMultipartParts:
+		return store.MinPartSize
+	// When we can divide our upload into exactly MaxMultipartParts parts with
+	// no bytes leftover, we will not need an spare last part.
+	// Also, when MaxObjectSize is equal to MaxPartSize * MaxMultipartParts
+	// (which is not the case with the current AWS S3 API specification, but
+	// might be in the future or with other S3-aware stores), we need this in
+	// order for our Multipart-Upload to reach full MaxObjectSize.
+	case int64(math.Mod(float64(*size), float64(store.MaxMultipartParts))) == 0:
```
Acconut: I believe this case could be merged with the default one since it should not make a difference here whether we need a spare last part or not. But maybe I am missing something.

trickkiste: If we go without this, we will not be able to fully upload a file with MaxObjectSize, in case MaxPartSize*MaxMultipartParts == MaxObjectSize. I would keep this to be prepared for eventual changes to the current S3 specs, which has happened on multiple occasions in the past, as far as I remember. That way, we are prepared for all eventualities.

s3store/s3store.go
trickkiste: I was writing a test, which would run this function for ALL possible file sizes. Therefore I was trying to be as efficient on memory-use as possible, as these are quite a lot of test cases.

trickkiste: Do you want me to change that?

s3store/s3store.go
trickkiste: OK, I was a little lazy there.

s3store/s3store.go
Acconut: > Therefore I was trying to be as efficient on memory-use as possible, as these are quite a lot of test cases.

Please don't do this kind of premature optimization. Unless you have proven using a benchmark that this change provides an improvement, stay away for that. Code readability and ease to understanding it is usually more important than performance.

Also I have serious doubt that this would bring any speed-ups since you need to dereference the pointer multiple times and also need to allocate memory for the pointer which could mean even worse memory-efficient. However, I am not a 100% sure on how Golang handles stack pointers but it doesn't matter here.

> Do you want me to change that?

Yes, please do so.

s3store/s3store.go
```diff
@@ -547,3 +561,28 @@ func isAwsError(err error, code string) bool {
...
+		return *size / store.MaxMultipartParts
+	// In all other cases, we need a spare last piece for the remaining bytes.
+	default:
+		return *size / (store.MaxMultipartParts - 1)
```
Acconut: Why is the -1 here? The last spare part can also have a size of up to `size / MaxMultipartParts` but it just won't all be used. Also we need to double check whether this integer division is not rounding down sometimes or else uploads would not finish.

trickkiste: I am afraid, you misread `MaxMultipartParts` here. The "- 1" in `(store.MaxMultipartParts - 1) `does not reduce the part size, but the total number of parts, we are dividing up our upload by. The part size actually grows due to that.

- MaxMultipartParts
is the maximum number of parts an S3 multipart upload is allowed to have according to AWS S3 API specifications.

However, compiling this answer, I actually found a bug in the algorithm, another special case, which needs to be checked for. I'll attend to that at the bottom of this post.

As for math.Mod, I just realised, there is the "%" operator for the remainder of an integer division, so of course that one is better suited, as it produces predictable results with no rounding problems. I will switch over to that.

As for the algorithm per se, the division by `MaxMultipartParts` will almost never produce an exact partition, but a number rounded down to an integer and there will be a remainder to this division. That remainder is what will end up in the last spare part. As we expect there to be a remainder, we need to divide the upload size by 9999 not 10.000, so we have a part left for the remainder. That is what is going on here.

The section above the `default:` checks, whether a division by 10.000 is applicable, by checking if a division `size / MaxMultipartParts` will produce any remainder or not. And only if it is a division without a remainder, we divide by `MaxMultipartParts` not `(MaxMultipartParts-1)` (10.000 not 9.999 in our case).

Here are a few examples with simple numbers for illustration:
MaxMultipartParts = 10
MinPartSize = 5
MaxPartSize = 10
MaxObjectSize = 100

size = 100
size / MaxMultipartParts  
= 100 / 10 = 10 with no remainder 
... No remainder, so we will divide by MaxMultipartParts and have 10 parts with size 10.
Total: 10 parts - SUCCESS.

size = 98
size / MaxMultipartParts  
= 99 / 10 = 9 with a remainder of 8
... So if we would use the result of this division as our optimalPartSize, then we would end up with 10 parts of size 9 and an 11th part of size 8. Which does not work, as we would exceed MaxMultipartParts.

You could say, well then let's just use the integer result of this division as the optimalPartSize (ergo 9). The fact that this division seemingly already produces, what we expect as optimalPartSize, is only a result of the fact, that MaxMultipartParts and MaxPartSize both are 10 in our example. So don't be mislead and rather ignore that fact.

Next we divide the size by (MaxMultipartParts - 1), so we have one spare part left, just in case.
size = 98
size / (MaxMultipartParts - 1)
= 99 / 9 = 10 with a remainder of 8
... So we have 9 parts of size 10 and a 10th part of size 8. 
Total: 10 parts - SUCCESS

I skipped size = 99, as it is a special case, and actually reveals a bug in the current algorithm:
size = 99
size / MaxMultipartParts
= 99 / 10 = 9 with a remainder of 9
... That means, we would end up with 10 parts of size 9 and an 11th part of size 9 again. In the current algorithm, as we find a rest in this division, we need to progress to the next case. 

But if we divide by (MaxMultipartParts - 1), we wont be lucky as well in this case:
= 99 / 9 = 11 with no remainder in that special case
... we end up with 9 parts of size 11, which exceeds MaxPartSize.

So we need to add another check:
case: size % (MaxMultiparts - 1) > MaxPartSize 
(ergo, if the last part would be larger than MaxPartSize)
then optimalPartSize = MaxPartSize
... we end up with 9 parts of size 10 and a last part of size 9. 
Total: 10 parts - SUCCESS

I'll play that with a few other numbers:
MaxMultipartParts = 20
MinPartSize = 5
MaxPartSize = 10
MaxObjectSize = 200

size 200
200 / 20 = 10 no remainder
... 20 parts with size 10. 
Total: 20 parts - SUCCESS.

size 199
199 / 20 = 9 with a remainder of 19
199 / 19 = 10 with a remainder of 9. The remainder is less than MaxPartSize, so apply the default.
... 19 parts with size 10 and a last one with size 9. 
Total: 20 parts - SUCCESS.

size 180
180 / 20 = 9 with no remainder
... 20 parts with size 9. 
Total: 20 parts - SUCCESS.

size 181
181 / 20 = 9 with a remainder of 1
181 / 19 = 9 with a remainder of 10. The remainder is not greater than MaxPartSize, so we apply default.
... 19 parts with size 9 and a last one with size 10. 
Total: 20 parts - SUCCESS.

We could also apply optimalPartSize = MaxPartSize here, so we would end up with 18 parts with size 10 and a last one with size 1. Total: 19 parts. As we thrive for the maximum number of parts with a minimum PartSize, the former solution is better.

size 182
182 / 20 = 9 with a remainder of 2
182 / 19 = 9 with a remainder of 11. The remainder is greater than MaxPartSize, so we apply MaxPartSize.
... 18 parts with size 10 and a last one with size 2. 
Total: 19 parts - SUCCESS.

size 189
189 / 20 = 9 with a remainder of 9
189 / 19 = 9 with a remainder of 18. The remainder is greater than MaxPartSize, so we apply MaxPartSize.
... 18 parts with size 10 and a last one with size 9. 
Total: 19 parts - SUCCESS.

size 190
190 / 20 = 9 with a remainder of 10 remainder
190 / 19 = 10 with no remainder. The remainder is less than MaxPartSize, so we apply default.
... 19 parts with size 10. 
Total: 19 parts - SUCCESS.

size 191
191 / 20 = 9 with a remainder of 11. We have a remainder, so we proceed to next case.
191 / 19 = 10 with a remainder of 1. The remainder is less than MaxPartSize, so we apply default.
... 19 parts with size 10 and a last one with size 1. 
Total: 20 parts - SUCCESS.

The addition of this additional switch should produce a universally applicable algorithm and replacing math.Mod with "%" operator, will fix your other concern.

trickkiste: I also need to test for lastpart > MaxPartSize in my tests.

trickkiste:
@Acconut, I cleaned up and simplified the algorithm.

Added error handling also. That part might need revision, as I don't know yet, what good standards are for doing that.

And I added a few more test patterns.

I tested the algorithm with a bunch of different settings for Min/MaxPartSize, MaxObjectSize and MaxMultipartParts. All produced positive results.

s3store/s3store.go
```diff
@@ -547,3 +560,40 @@ func isAwsError(err error, code string) bool {
...
+
+func (store S3Store) CalcOptimalPartSize(size int64) (optimalPartSize int64, err error) {
+	var errorstrings []string
+
+	switch {
+	// When upload is smaller or equal MinPartSize, we upload in just one part.
+	case size <= store.MinPartSize:
+		optimalPartSize = store.MinPartSize
+	// When we need 9999 parts or less with MinPartSize.
+	case size/store.MinPartSize < store.MaxMultipartParts:
```
Acconut: Could we remove the integer division here and test for `size < MinPartSize*MaxMultipartParts` instead? I don't think this makes a difference here but I would like to avoid integer division whenever possible since it does not behave intuitively first :)

s3store/s3store.go
```diff
@@ -547,3 +560,40 @@ func isAwsError(err error, code string) bool {
...
+		optimalPartSize = store.MinPartSize
+	// If our upload divides up exactly into MaxMultipartParts parts with
+	// no bytes leftover, we will not need an spare last part. So we can go with
+	// a straight division. Otherwise we would exceed MaxPartSize in a scenario,
+	// where MaxObjectSize is equal to MaxPartSize * MaxMultipartParts
+	// (which is not the case with the current AWS S3 API specification, but
+	// might be in the future or with other S3-aware stores).
+	case size%store.MaxMultipartParts == 0:
+		optimalPartSize = size / store.MaxMultipartParts
+	// In all other cases, we need to round up to the next integer.
```
Acconut: Would you be so kind and add to the comment that integer division will round down, if necessary. This will help understanding the "+1"

trickkiste: No problem.

s3store/s3store.go
```diff
@@ -562,27 +561,39 @@ func isAwsError(err error, code string) bool {
...
 	default:
-		return *size / (store.MaxMultipartParts - 1)
+		optimalPartSize = size/store.MaxMultipartParts + 1
+	}
+
+	// an upload larger than MaxObjectSize must throw an error
+	if size > store.MaxObjectSize {
+		errorstrings = append(errorstrings, fmt.Sprintf("CalcOptimalPartSize: upload size of %v bytes exceeds MaxObjectSize of %v bytes", size, store.MaxObjectSize))
```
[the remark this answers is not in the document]
trickkiste: I wanted optimalPartSize still to be calculated and returned in any case, as it helps to debug, how the algorithm is behaving. Especially useful for the tests.

s3store/s3store.go
trickkiste: We can do that. My emphasis lies on readability and understandability and this change does rather further them then harm them.

trickkiste:
@Acconut, I am all for running the algorithm for all file sizes from `0 to MaxObjectSize+1` with the settings you proposed.
```
MaxPartSize:       5 * 1024,
MinPartSize:       5,
MaxMultipartParts: 10000,
MaxObjectSize:     5 * 1024 * 1024,
```
Except, I would probably go for 

> MaxMultipartParts: 1000

 or at least 

> MaxObjectSize: MaxPartSize*MaxMultipartParts

 to cover that edge case as well.

I had run such a simulation with a prior version of the algorithm and less test cases. So I would have to run it anew, as we have had a couple of iterations since. I would be equally happy, if you would run it ;-)

s3store/s3store.go
trickkiste: Doing the division in float64 and then rounding up the result is no option I guess, as we might end up with rounding errors due to the float nature of things?!

It would enhance readability, but probably introduce a whole new set of problems.

Acconut: I believe the errors which occur due to the use of a floating point number is too minimal to be harmful for our usecase but I am not keen on testing that :)
Also, rounding functions in Go also returns a float64 not integers. Let's keep the current approach.

s3store/s3store.go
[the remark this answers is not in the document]
Acconut: IMO there is no need to calculate a part size if the upload exceeds the maximum object size because the upload will never be possible. It's confusion why this check is down here.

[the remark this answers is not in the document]
trickkiste: OK. So I return part size 0 and the error then. Coming in a second.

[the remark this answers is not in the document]
trickkiste: I will have to restructure the tests again to accommodate that change. One more commit coming.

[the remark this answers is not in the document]
trickkiste: I am too quick to push. One more fix pending.

s3store/s3store_test.go
```diff
@@ -24,6 +24,138 @@ var _ tusd.TerminaterDataStore = s3store.S3Store{}
...
+			err += fmt.Sprintf("Testcase #%v size %v: size exceeds MaxObjectSize %v but no error returned\n", index, size, MaxObjectSize)
+		}
+		if debug && optimalPartSize == 0 {
+			fmt.Printf("Testcase #%v size %v: size exceeds MaxObjectSize %v\n", index, size, MaxObjectSize)
+		}
+		if optimalPartSize > 0 {
+			equalparts = size / optimalPartSize
+			lastpartsize = size % optimalPartSize
+			if optimalPartSize < MinPartSize {
+				err += fmt.Sprintf("Testcase #%v size %v, %v parts of size %v, lastpart %v: optimalPartSize < MinPartSize %v\n", index, size, equalparts, optimalPartSize, lastpartsize, MinPartSize)
```
Acconut: Any reason why you are not using t.Errorf here and in the following if statement?

s3store/s3store_test.go
```diff
@@ -24,6 +24,138 @@ var _ tusd.TerminaterDataStore = s3store.S3Store{}
...
+	var HighestApplicablePartSize int64 = MaxObjectSize / MaxMultipartParts
+	if MaxObjectSize%MaxMultipartParts > 0 {
+		HighestApplicablePartSize++
+	}
+	var RemainderWithHighestApplicablePartSize int64 = MaxObjectSize % HighestApplicablePartSize
+
+	// some of these tests are actually duplicates, as they specify the same size
+	// in bytes - two ways to describe the same thing. That is wanted, in order
+	// to provide a full picture from any angle.
+	testcases := []int64{
```
Acconut: It would be nice if the test cases where separated into situations where a part size can and cannot be calculated, making the result assertion and error checking for straight forward.

s3store/s3store_test.go
trickkiste: No. As said, with error handling, I am hoping for your advice. My Golang experience is limited and I did not read up on everything yet. I was generating strings and only using t.Errorf at the bottom, to make sure, I see all errors not just the first one. Don't know, if this is necessary.

s3store/s3store_test.go
trickkiste: It is true, that the current set of checks in TestCalcOptimalPartSize are not straight forward.

However, in my opinion the assertions should work with any size you throw at them, no matter, if we expect a success of an error from CalcOptimalPartSize. One reason for that is, that certain testcases might produce the one result with set A of S3 settings but might produce another with set B.

Before we moved the MaxObjectSize condition above the calculation of the optimalPartSize and by that means creating a scenario, where CalcOptimalPartSize could output 0 as a result, the tests were actually very straight forward, because they were linear. No if condition in there. That is the reason, why I preferred optimalPartSize being calculated, even if size > MaxObjectSize.

One could also argue, that the purpose of CalcOptimalPartSize is calculating said part size and checking whether the size of the upload is larger than MaxObjectSize should not even concern this function but actually would belong somewhere else!

E.g. in NewUpload, because why would we create such an upload in the first place, if we could never finish it.

Therefore I would propose to move that condition to NewUpload and keep CalcOptimalPartSize and it's tests clean.

Don't you agree?

trickkiste:
Would you share your current test setup? I'd like to collaborate.

trickkiste:
I pushed my proposal on how to simplify CalcOptimalPartSize.

s3store/s3store_test.go
Acconut: That's why I asked :) FYI, t.Error will continue execution after it has been invoked while t.Fatal will immediately stop the test case (see https://golang.org/pkg/testing/#B.Errorf). For example, I use t.Fatal if a database connection could not be made and further testing is useless. On the other hand, t.Error is used for asserting that the result has specific properties.

s3store/s3store_test.go
```diff
@@ -43,119 +43,129 @@ func TestCalcOptimalPartSize(t *testing.T) {
...
+
+	var debug = false
+	var equalparts, lastpartsize int64
+
+	// sanity check
+	if store.MaxObjectSize > store.MaxPartSize*store.MaxMultipartParts {
+		t.Errorf("MaxObjectSize %v can never be achieved, as MaxMultipartParts %v and MaxPartSize %v only allow for an upload of %v bytes total.\n", store.MaxObjectSize, store.MaxMultipartParts, store.MaxPartSize, store.MaxMultipartParts*store.MaxPartSize)
+	}
+
+	for size := int64(1); size <= store.MaxObjectSize; size++ {
```
Acconut: Please start with `size = 0`. In theory, empty uploads are also allowed.

s3store/s3store_test.go
```diff
@@ -43,119 +43,129 @@ func TestCalcOptimalPartSize(t *testing.T) {
...
+			fmt.Printf("Size %v, %v parts of size %v, lastpart %v, does exceed MaxObjectSize %v.\n", size, equalparts, optimalPartSize, lastpartsize, size > store.MaxObjectSize)
 		}
 	}
 	// fmt.Println("HighestApplicablePartSize", HighestApplicablePartSize)
 	// fmt.Println("RemainderWithHighestApplicablePartSize", RemainderWithHighestApplicablePartSize)
 }
 
+func TestAllPartSizes(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping test in short mode.")
```
Acconut: Skipping this test is not necessary I believe. For me it completes in 0.2s.

s3store/s3store_test.go
```diff
@@ -43,119 +43,129 @@ func TestCalcOptimalPartSize(t *testing.T) {
...
+	// sanity check
+	if store.MaxObjectSize > store.MaxPartSize*store.MaxMultipartParts {
+		t.Errorf("MaxObjectSize %v can never be achieved, as MaxMultipartParts %v and MaxPartSize %v only allow for an upload of %v bytes total.\n", store.MaxObjectSize, store.MaxMultipartParts, store.MaxPartSize, store.MaxMultipartParts*store.MaxPartSize)
+	}
+
+	for size := int64(1); size <= store.MaxObjectSize; size++ {
+		optimalPartSize, calcError := store.CalcOptimalPartSize(size)
+		equalparts = size / optimalPartSize
+		lastpartsize = size % optimalPartSize
+		assert.False(optimalPartSize < store.MinPartSize, fmt.Sprintf("Size %v, %v parts of size %v, lastpart %v: optimalPartSize < MinPartSize %v.\n", size, equalparts, optimalPartSize, lastpartsize, store.MinPartSize))
```
Acconut: You can save the call to fmt.Sprintf by using assert.Falsef  and similar methods (see https://godoc.org/github.com/stretchr/testify/assert#Assertions.Falsef).

s3store/s3store_test.go
```diff
@@ -43,119 +43,129 @@ func TestCalcOptimalPartSize(t *testing.T) {
...
+	var debug = false
+	var equalparts, lastpartsize int64
+
+	// sanity check
+	if store.MaxObjectSize > store.MaxPartSize*store.MaxMultipartParts {
+		t.Errorf("MaxObjectSize %v can never be achieved, as MaxMultipartParts %v and MaxPartSize %v only allow for an upload of %v bytes total.\n", store.MaxObjectSize, store.MaxMultipartParts, store.MaxPartSize, store.MaxMultipartParts*store.MaxPartSize)
+	}
+
+	for size := int64(1); size <= store.MaxObjectSize; size++ {
+		optimalPartSize, calcError := store.CalcOptimalPartSize(size)
```
Acconut: Is there a reason why you didn't check for `calcError == nil` here? For the range of `size` is should never return an error.

s3store/s3store_test.go
trickkiste: Well for you.
31.883s for me!
MacBook Air Mid 2012

s3store/s3store_test.go
trickkiste: Yes there is.
I view the range we are testing here as arbitrary and therefore the assertions should work with any range you choose. I actually would much prefer to test up to MaxObjectSize+1. An assertion `calcError == nil` would not work in this case.

s3store/s3store_test.go
trickkiste: Roger that.

s3store/s3store_test.go
trickkiste: Also I believe that having identical assertions in the two tests concerning CalcOptimalPartSize helps prevent confusion. As we are testing the same thing, just different ranges of test cases, so why should the test differ.

Ideally, we would externalise the assertions into a function, which we use in both tests, which would assure, they stay the same over time. But as this would complicate things instead of making them more readable, I did refrain from trying that.

s3store/s3store_test.go
trickkiste: When I do this: 
```
assert.Falsef(optimalPartSize < store.MinPartSize, "Size %v, %v parts of size %v, lastpart %v: optimalPartSize < MinPartSize %v.\n", size, equalparts, optimalPartSize, lastpartsize, store.MinPartSize)
```

I get that: 
```
assert.Falsef undefined (type *assert.Assertions has no field or method Falsef)
```

With `go version go1.7.6 linux/amd64`

Acconut: That's because the vendored version of testify (https://github.com/tus/tusd/tree/master/vendor/github.com/stretchr/testify) is not a recent one and does not feature the Falsef method (alongside others). I will update the dependency tomorrow on master so you can merge it into this branch.

trickkiste: OK, I'll merge the new master then.
But let's get this to the finishing line soon now.
I am finishing the TestNewUploadLargerMaxObjectSize test and that should be it then.
Or are you missing anything else?

s3store/s3store_test.go
Acconut: I disagree. Call it preference or not, I am not a fan of having different scenarios into one test.

>  I actually would much prefer to test up to MaxObjectSize+1

I don't see it this way. This should be its own separate test since the expected outcome and the used assertions are different. Furthermore, this will make debugging issues easier since you can run a single test isolated using `go test -run`.

s3store/s3store_test.go
Acconut: > However, in my opinion the assertions should work with any size you throw at them, no matter, if we expect a success or an error from CalcOptimalPartSize. 

I don't share this opinion. Test should assert the returned output from a function if you provide it with the specific input, making them straightforward to understand and also debug if necessary. For example, let's take `store.MaxObjectSize + 1`. Should a part size be calculated for this input? I don't know and I cannot tell from this test case. Frankly, I cannot even tell you whether your test verifies that the behavior of CalcPartSize does not change.

>  One reason for that is, that certain testcases might produce the one result with set A of S3 settings but might produce another with set B. Therefore having dividing the test into a set, which is expected to fail, and one, which is expected to succeed, might not actually be realistic.

If you make the test inputs depend on store.MaxPartSize (etc.) then this should not be an issue, as you did.

> Therefore I would propose to move that condition to NewUpload and keep CalcOptimalPartSize and it's tests clean and universal.

Seen that you already took care of this, I welcome this change :)

s3store/s3store_test.go
```diff
@@ -150,16 +151,16 @@ func TestAllPartSizes(t *testing.T) {
...
-		assert.False(size%optimalPartSize == 0 && equalparts > store.MaxMultipartParts, fmt.Sprintf("Size %v, %v parts of size %v, lastpart %v: more parts than MaxMultipartParts %v.\n", size, equalparts, optimalPartSize, lastpartsize, store.MaxMultipartParts))
-		assert.False(size%optimalPartSize > 0 && equalparts > store.MaxMultipartParts-1, fmt.Sprintf("Size %v, %v parts of size %v, lastpart %v: more parts than MaxMultipartParts %v.\n", size, equalparts, optimalPartSize, lastpartsize, store.MaxMultipartParts))
-		assert.False(lastpartsize > store.MaxPartSize, fmt.Sprintf("Size %v, %v parts of size %v, lastpart %v: lastpart > MaxPartSize %v.\n", size, equalparts, optimalPartSize, lastpartsize, store.MaxPartSize))
-		assert.False(lastpartsize > optimalPartSize, fmt.Sprintf("Size %v, %v parts of size %v, lastpart %v: lastpart > optimalPartSize %v.\n", size, equalparts, optimalPartSize, lastpartsize, optimalPartSize))
+		assert.False(optimalPartSize < store.MinPartSize, "Size %v, %v parts of size %v, lastpart %v: optimalPartSize < MinPartSize %v.\n", size, equalparts, optimalPartSize, lastpartsize, store.MinPartSize)
```
Acconut: They have to be assert.Falsef calls, I believe.

s3store/s3store_test.go
trickkiste: Well, that is certainly a valid argument.

But if one would follow that perspective, the test would need to be ripped apart even further. Each one of the case statements in CalcOptimalPartSize then would represent its own test case in your scenario. 

Readability will be totally gone then. 

No single point to understand, what we are expecting from CalcOptimalPartSize anymore. And you would have to define universal scenarios for sizes, which will trigger the expected behaviour. Which will become rather complicated, when you consider, that MinPartSize, MaxpartSize, MaxMultipartParts and MaxObjectSize are variable too and might be very different in other S3 API implementations.

You might end up with a lot more tests, than you are prepared to handle.

s3store/s3store_test.go
trickkiste: It works. The output is correct.
But of course, we can pump this up to the latest testify version and switch to assert.Falsef.

s3store/s3store_test.go
trickkiste: I believe there is a difference between testing something like WriteChunk, which by itself depends on other APIs, functions and methods and a helper function with only one input and one output and no additional level of abstraction.

For WriteChunk the quality of the data on the input can differ quite significantly, both in value, but also in its structural composition. And moreover, we depend on other functions, methods and interfaces below it.

While CalcOptimalPartSize just takes a number and will return another one, or an error. But it will not venture off into other layers of abstraction.

s3store/s3store_test.go
Acconut: Sorry, my bad. I wasn't aware that assert.False also accepts a formattable string. So everything is fine here :)

Acconut:
Merged in 2df5d116725f952da53d8a9b2a75fab0df784833, thanks :)

PR #161, opened 2017-11-28
GetInfo now properly retrieves info for composed objects

peixian:
- This affects the GCSStore only.

Before:
The GCSStore did not search for the proper prefix, which meant that composed files were not seen and always returned 0 for `Upload-Offset` for completed uploads.

Now:
GCSStore checks for all files matching the hash, including completed files and the info file. 

Specifics: 
Before, when an object was composed and finished uploading, it took the filename of `<RANDOM_HASH>`, while tusd used the prefix of `<RANDOM_HASH>_`, resulting in it not finding any files, and stating the uploaded file did not exist, resulting in an `Upload-Offset` of `0`. This PR now checks the bucket for all files with the prefix of `<RANDOM_HASH>`, which now accounts for composed files. 

cc:
@devonbalicki @vayam

### DIFF
--- a/gcsstore/gcsservice.go
+++ b/gcsstore/gcsservice.go
@@ -333,6 +333,7 @@ func (service *GCSService) FilterObjects(params GCSFilterParams) ([]string, erro
 
 	it := bkt.Objects(service.Ctx, &q)
 	names := make([]string, 0)
+loop:
 	for {
 		objAttrs, err := it.Next()
 		if err == iterator.Done {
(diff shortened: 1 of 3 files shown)


gcsstore/gcsservice.go
```diff
@@ -342,19 +343,27 @@ func (service *GCSService) FilterObjects(params GCSFilterParams) ([]string, erro
...
 		// know the object name we are working with is in the format
 		// [uid]_tmp_[recursion_lvl]_[chunk_idx]. The only time we filter
 		// these temporary objects is on a delete operation so we can just
 		// append and continue without worrying about index order
-		if len(split) == 4 {
+
+		switch len(split) {
+		case 1:
+			names = []string{objAttrs.Name}
+			break loop
```
Acconut: I am wondering if this labelled break could be replaced with a simple `continue` since this would jump to the beginning of the for loop?

peixian: This is a break because in the case of the partitions and the composed file remaining in GCS (say you have 3 chunks and the composed file already, but the delete failed on the chunks) the iterator doesn't proceed to go through the chunks. If a composed file is found (in the case of `len(split) == 1`, we should just exit out of the loop entirely,.

Acconut: Oh my bad, I confused the labeled break with a goto statement and thought that this would jump to the beginning of the loop but instead it terminates the loop. In that case, it's all good :+1:

Acconut:
Thank you :)

PR #162, opened 2017-11-28
[wip] K8s

kiloreux:


.infra/kube/deployment.yaml
```diff
@@ -0,0 +1,19 @@
...
+  namespace: tus
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: tusd
+    spec:
+      containers:
+      - image: docker.io/kiloreux/tusd:latest
```
Acconut: We have an official docker image at https://hub.docker.com/r/tusproject/tusd/ but I am not sure if it's always the latest master.

.infra/kube/deployment.yaml
```diff
@@ -0,0 +1,19 @@
...
+        imagePullPolicy: Always
+        args: ["-port=8080"]
```
Acconut: The current Frey setup uses more arguments (see https://github.com/tus/tusd/blob/master/.infra/Freyfile.hcl#L161). At least `-behind-proxy` and `-max-size` are needed. Also, what size does the disk have?

.infra/kube/ingress-tls.yaml
```diff
@@ -0,0 +1,22 @@
...
+metadata:
+  name: tusd
+  namespace: tus
+  annotations:
+    kubernetes.io/tls-acme: "true"
+    kubernetes.io/ingress.class: "nginx"
+spec:
+  tls:
+  - hosts:
+    - dzybha.com
```
Acconut: Still a leftover from experimenting?

.infra/kube/ingress-tls.yaml
```diff
@@ -0,0 +1,22 @@
...
+    secretName: tusd-tls
+  rules:
+  - host: dzybha.com
```
Acconut: Still a leftover from experimenting?

.infra/kube/ingress-tls.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: tusd
+  namespace: tus
+  annotations:
+    kubernetes.io/tls-acme: "true"
+    kubernetes.io/ingress.class: "nginx"
```
Acconut: We have a custom nginx configuration which contain a few necessary settings: https://github.com/tus/tusd/blob/master/.infra/files/nginx.conf

.scripts/deploy_gcloud.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# Set magic variables for current FILE & DIR
+__dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+__root="$(cd "$(dirname "${__dir}")" && pwd)"
+
+# Store the new image in docker hub
+docker build --quiet -t kiloreux/tusd:latest -t kiloreux/tusd:$TRAVIS_COMMIT ${__root};
+docker login -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD";
+docker push kiloreux/tusd:$TRAVIS_COMMIT;
+docker push kiloreux/tusd:latest;
```
Acconut: Could this be tusproject/tusd?

.infra/kube/ingress-tls.yaml
kiloreux: We actually deploy to this domain so we don't risk ruining anything. Once we are sure and want to merge to master we change this to our master domain. Otherwise we have no way of seeing the changes.

.scripts/deploy_gcloud.sh
kiloreux: It could be of course. But I don't have the permissions there and It was already late so no one will grant me that. I figured I would go with my own setup and experiment on building and then we will just push images to the tusproject/tusd repository

.infra/kube/deployment.yaml
kiloreux: Already aware of that one like already mentioned I don't have permissions there.

.scripts/deploy_gcloud.sh
kvz: iirc, we use docker automated builds to push to tusproject/tusd or no? so that will collide then. I do agree of course that we should have just one container, and that is `tusproject/tusd`, but we need to figure out how to best piece this all together then

kvz:
This should likely have a `[WIP]` prefix

.infra/kube/ingress-tls.yaml
kiloreux: What specific parameters do you mean? Can you be more precise about them please.

.scripts/deploy_gcloud.sh
Acconut: > we use docker automated builds to push to tusproject/tusd or no?

I am not sure whether this feature got finished, but I actually prefer using Travis CI for pushing the new docker container. This keeps the test/build/deploy process on one service instead of using Travis and Docker.

.infra/kube/ingress-tls.yaml
Acconut: Well, all of them are required. :) I am not sure which SSL configuration is used in this PR but they are obviously necessary. Also, some settings (https://github.com/tus/tusd/blob/master/.infra/files/nginx.conf#L52-L65) are more tus-specific and therefore needed.

kvz:
+1

Let’s give @kiloreux access to tusproject/tusd and push straight to it from the same build-chain.

Or better: let’s create a cibot user that can do this :)

kvz:
I do wonder if their system is a requirement to be included in docker’s stdlib

PR #171, opened 2018-03-02
Log hooks to stdout

peixian:
Hooks currently get logged into stderr, which (if you're running tus in GKE) result in hook logs being erroneously reported as errors. This PR sets hook logs to go to stdout. 

It's also possible that this should be a user configurable setting, I'm open to suggestions on how this should be implemented.

### DIFF
--- a/cmd/tusd/cli/log.go
+++ b/cmd/tusd/cli/log.go
@@ -10,6 +10,6 @@ import (
 var stdout = log.New(os.Stdout, "[tusd] ", 0)
 var stderr = log.New(os.Stderr, "[tusd] ", 0)
 
-func logEv(eventName string, details ...string) {
-	tusd.LogEvent(stderr, eventName, details...)
+func logEv(logOutput *log.Logger, eventName string, details ...string) {
+	tusd.LogEvent(logOutput, eventName, details...)
 }
--- a/cmd/tusd/cli/hooks.go
+++ b/cmd/tusd/cli/hooks.go
@@ -70,16 +70,16 @@ func invokeHook(typ HookType, info tusd.FileInfo) {
 func invokeHookSync(typ HookType, info tusd.FileInfo, captureOutput bool) ([]byte, error) {
 	switch typ {
 	case HookPostFinish:
-		logEv("UploadFinished", "id", info.ID, "size", strconv.FormatInt(info.Size, 10))
+		logEv(stdout, "UploadFinished", "id", info.ID, "size", strconv.FormatInt(info.Size, 10))
 	case HookPostTerminate:
-		logEv("UploadTerminated", "id", info.ID)
+		logEv(stdout, "UploadTerminated", "id", info.ID)
 	}
 
 	if !Flags.FileHooksInstalled && !Flags.HttpHooksInstalled {
 		return nil, nil
 	}
 	name := string(typ)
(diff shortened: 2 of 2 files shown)


Acconut:
First of all, sorry for my delayed response.

I don't recall a specific reason why these hooks events are written to stderr, in particular if all other non-error events are logged on stdout. So, I am more than open for this change.

cmd/tusd/cli/log.go
```diff
@@ -7,9 +7,12 @@ import (
 	"github.com/tus/tusd"
 )
 
-var stdout = log.New(os.Stdout, "[tusd] ", 0)
 var stderr = log.New(os.Stderr, "[tusd] ", 0)
 
-func logEv(eventName string, details ...string) {
-	tusd.LogEvent(stderr, eventName, details...)
```
Acconut: As far as I can tell, logEv is only used for events which should be written to stdout. So I think it should be enough to change this line to `tusd.LogEvent(stderr, eventName, details...)` without the rest of your changes. Could it be that I am missing something?

peixian: Nope, it can definitely be like that, I just wasn't sure if there was some coding standard to use logging as a global variable. I'll change it.

peixian: Actually, looking at this again @Acconut, https://github.com/tus/tusd/pull/171/files#diff-87d4bf0db9b6383dae738a3c1168365eR99 should be logging to stderr, since it's actually reporting a hook error.

Acconut: Ok, I am also fine with that :+1:

cmd/tusd/cli/log.go
```diff
@@ -2,14 +2,14 @@ package cli
...
 	"github.com/tus/tusd"
 )
 
-var stdout = log.New(os.Stdout, "[tusd] ", 0)
-var stderr = log.New(os.Stderr, "[tusd] ", 0)
-
-func logEv(eventName string, details ...string) {
-	tusd.LogEvent(stderr, eventName, details...)
+func logEv(logOutput *log.Logger, eventName string, details ...string) {
+	if logOutput == nil {
```
Acconut: I don't think that this check is needed since you updated all usages of logEv.

cmd/tusd/cli/log.go
```diff
@@ -2,14 +2,14 @@ package cli
 
 import (
 	"log"
-	"os"
 
 	"github.com/tus/tusd"
 )
 
-var stdout = log.New(os.Stdout, "[tusd] ", 0)
-var stderr = log.New(os.Stderr, "[tusd] ", 0)
```
Acconut: Would you mind moving the definition of stdout and stderr back into log.go? I think the semantically belong into this file since they are used for logging. Furthermore, these values are also used in composer.go, so they don't belong to hooks.go

cmd/tusd/cli/log.go
peixian: Yep, definitely unneeded.

cmd/tusd/cli/log.go
peixian: Will do.

Acconut:
Thank you very much for this! BTW, GitHub does not send notifications if a new commit is pushed to a PR, so I am sorry if you have been waiting for me reaction.

PR #182, opened 2018-04-29
Implement Upload-Defer-Length Extension

acj:
This is a first pass at supporting the upload-defer-length protocol extension. It's worked well in my own testing, but I suspect that there may be edge cases that aren't being handled properly. For example, it's difficult for the S3 backend to ensure that the minimum chunk size is respected when the full file size isn't known.

Happy to discuss and iterate on this further.


acj:
@Acconut That approach sounds good to me. I'll work on implementing it for this PR. Thank you for the detailed suggestion.

The S3 data store is my primary use case, so I'd like to make it capable of deferring the upload length -- probably in a separate PR. I think that the key problem is recognizing when we've written a chunk smaller than `MinPartSize` (which should be the terminal chunk) and then try to write a subsequent chunk.

Do you think it would suffice to add a field to `FileInfo`, something like `HaveWrittenShortChunk bool`, that we can inspect in the S3 data store's `WriteChunk` function and then abort the upload with an appropriate error?

gcsstore/gcsstore.go
```diff
@@ -257,3 +258,7 @@ func (store GCSStore) GetReader(id string) (io.Reader, error) {
 
 	return r, nil
 }
+
+func (store GCSStore) SupportsDeferredLength() bool {
+	return true
```
Acconut: Frankly, I am not so sure whether GCSStore supports this extension out of the box.

Acconut:
No worries, I wasn't really available either. May I ask what's the status on this PR? Is it finished or are you still working on it?

acj:
It's finished in terms of implementation, but I wanted to ask if any additional tests are needed. Since you mentioned it in another comment, is there anything specific to GCSStore that concerns you?

acj:
> I believe there is still a function missing in the PATCH handler which permanently stores the upload's size once it's known for uploads with deferred length.

Yes, good catch. I've added a `PutInfo` function on `DataStore` to handle persistence.

> since it has not been shown yet that GCSStore is compatible with this extension, I would not assume it is.

Fair enough. I haven't found any documentation that suggests a problem, but it would be good to test this against a real GCS bucket. I've disabled it for now.

Please let me know if any other changes or tests are needed.

Acconut:
> I've added a PutInfo function on DataStore to handle persistence.

Great idea, however I would have made the function less general, e.g. `DeclareLength` instead of `PutInfo`. The reason is that users may assume that the PutInfo function is able of overwriting any value in the FileInfo struct. However, this is not always the case:
For example, the ID, IsPartial, IsFinal  and PartialUploads properties should never be changed after the upload has been created. The Offset property should only be changed when WriteChunk is involved. On S3 (and also possibly other stores) metadata cannot be modified after an upload has been created.
Therefore, I would propose to just add a `DeclareLength(length int) error` function to the LengthDeferrerDataStore inferface. In that turn we can also get rid of the `SupportsDeferredLength` which I only proposed so that we do not have an empty interface.

Sorry for this inconvenience but does this make sense? Everything else is looking perfect :+1:

acj:
No worries. That makes sense and sounds good to me. I'll make the suggested change.

> the ID, IsPartial, IsFinal and PartialUploads properties should never be changed after the upload has been created

Do you think this property should be documented somewhere, e.g. on each field of the `FileInfo` struct? It seems obvious in retrospect, but I didn't think about it when I was working on the last batch of changes.

acj:
Okay, I did a bit of branch cleanup and force-pushed. Your `DeclareLength` idea worked out very well, I think. It's cleaner, simpler, and easier to understand. :+1:

Acconut:
This look amazing, thank you very much! I had a quick look over this but I am going to test it in full depth this week.

unrouted_handler.go
```diff
@@ -115,7 +119,7 @@ func NewUnroutedHandler(config Config) (*UnroutedHandler, error) {
 	}
 
 	// Only promote extesions using the Tus-Extension header which are implemented
-	extensions := "creation,creation-with-upload"
+	extensions := "creation,creation-with-upload,creation-defer-length"
```
Acconut: `creation-defer-length` should only be added if the data storage support it, similar to how the Termination extension is handled in the next lines.

unrouted_handler.go
```diff
@@ -437,12 +450,27 @@ func (handler *UnroutedHandler) PatchFile(w http.ResponseWriter, r *http.Request
...
+	if !info.SizeIsDeferred && info.Offset == info.Size {
 		w.Header().Set("Upload-Offset", strconv.FormatInt(offset, 10))
 		handler.sendResp(w, r, http.StatusNoContent)
 		return
 	}
 
+	if handler.composer.UsesLengthDeferrer && info.SizeIsDeferred && r.Header.Get("Upload-Length") != "" {
+		uploadLength, err := strconv.ParseInt(r.Header.Get("Upload-Length"), 10, 64)
+		if err != nil || uploadLength < 0 {
+			handler.sendError(w, r, ErrInvalidOffset)
```
Acconut: `ErrInvalidOffset` seems to not be the correct error here. What about `ErrInvalidUploadLength`?

unrouted_handler.go
```diff
@@ -437,12 +450,27 @@ func (handler *UnroutedHandler) PatchFile(w http.ResponseWriter, r *http.Request
...
+			return
+		}
+
+		info.Size = uploadLength
+		info.SizeIsDeferred = false
+		if err := handler.composer.LengthDeferrer.DeclareLength(id, info.Size); err != nil {
```
Acconut: An additional check if the new length is not smaller than the current offset (`info.Offset <= uploadLength`) and the new length does not exceed the file limit (see https://github.com/tus/tusd/blob/master/unrouted_handler.go#L264-L268) before the call to `DeclareLength` would be needed.

unrouted_handler.go
```diff
@@ -437,12 +450,27 @@ func (handler *UnroutedHandler) PatchFile(w http.ResponseWriter, r *http.Request
...
 
 	// Do not proxy the call to the data store if the upload is already completed
-	if info.Offset == info.Size {
+	if !info.SizeIsDeferred && info.Offset == info.Size {
 		w.Header().Set("Upload-Offset", strconv.FormatInt(offset, 10))
 		handler.sendResp(w, r, http.StatusNoContent)
 		return
 	}
 
+	if handler.composer.UsesLengthDeferrer && info.SizeIsDeferred && r.Header.Get("Upload-Length") != "" {
```
Acconut: Although this check is perfectly fine, I would love to see an error if someone tries to declare the length for an upload for a) an upload where length deferring is not enabled and b) an upload where the length has already been declared.

unrouted_handler.go
```diff
@@ -460,7 +488,7 @@ func (handler *UnroutedHandler) writeChunk(id string, info FileInfo, w http.Resp
 	offset := info.Offset
 
 	// Test if this upload fits into the file's size
-	if offset+length > info.Size {
+	if !info.SizeIsDeferred && offset+length > info.Size {
 		return ErrSizeExceeded
 	}
```
Acconut: The following lines where maxSize is defined also need a consideration if the size is deferred. Something like:

```go
maxSize := info.Size - offset
// If the upload's length is deferred and the PATCH request does not contain the Content-Length
// header (which is allowed if 'Transfer-Encoding: chunked' is used), we still need to set limits for
// the body size.
if info.SizeIsDeferred {
	if handler.config.MaxSize > 0 {
		// Ensure that the upload does not exceed the maximum upload size
		maxSize = handler.config.MaxSize - offset
	} else {
		// If no upload limit is given, we will allowed arbitrary sizes
		maxSize = math.MaxInt64
	}
}
if length > 0 {
	maxSize = length
}
```

acj:
Thanks again for the detailed feedback. I've fixed the things you pointed out, rebased onto current master, and pushed. Let me know if anything else needs to be done.

unrouted_handler.go
```diff
@@ -437,12 +454,35 @@ func (handler *UnroutedHandler) PatchFile(w http.ResponseWriter, r *http.Request
...
+				info.SizeIsDeferred = false
+				if err := handler.composer.LengthDeferrer.DeclareLength(id, info.Size); err != nil {
+					handler.sendError(w, r, err)
+					return
+				}
+			} else {
+				handler.sendError(w, r, ErrInvalidUploadLength)
+			}
+		} else {
+			handler.sendError(w, r, ErrNotImplemented)
```
kvz: I think christmas trees like these should be avoided if reasonably possible by returning home early. I think there's an official term for it, but I couldn't find it. Maybe this post that I did find helps illustrate the concept tho http://blog.timoxley.com/post/47041269194/avoid-else-return-early :)

kvz: i realize this pr has been merged already and it's nothing critical. just a friendly tip for future endeavours :)

acj: Thanks for the feedback, @kvz. I agree -- an early return would be better. I also realized that this PR doesn't add `Upload-Defer-Length` to the CORS allowed header list, so I'll create a followup PR that addresses both.

Issue #187, opened 2018-05-12
Some dependencies are missing from vendor directory.


miladz68:
When I am doing a git clone and go build, it says that it misses some packages. So I submit this since some people may try to build this from source they might be unable to do so.

here is the output:

```
git clone https://github.com/tus/tusd.git
go build -o tusd cmd/tusd/main.go
```

```
cmd/tusd/cli/composer.go:13:2: cannot find package "github.com/aws/aws-sdk-go/aws" in any of:
	/home/milad/go/src/github.com/tus/tusd/vendor/github.com/aws/aws-sdk-go/aws (vendor tree)
	/usr/local/go/src/github.com/aws/aws-sdk-go/aws (from $GOROOT)
	/home/milad/go/src/github.com/aws/aws-sdk-go/aws (from $GOPATH)
s3store/s3store.go:98:2: cannot find package "github.com/aws/aws-sdk-go/aws/awserr" in any of:
	/home/milad/go/src/github.com/tus/tusd/vendor/github.com/aws/aws-sdk-go/aws/awserr (vendor tree)
	/usr/local/go/src/github.com/aws/aws-sdk-go/aws/awserr (from $GOROOT)
	/home/milad/go/src/github.com/aws/aws-sdk-go/aws/awserr (from $GOPATH)
cmd/tusd/cli/composer.go:14:2: cannot find package "github.com/aws/aws-sdk-go/aws/credentials" in any of:
	/home/milad/go/src/github.com/tus/tusd/vendor/github.com/aws/aws-sdk-go/aws/credentials (vendor tree)
	/usr/local/go/src/github.com/aws/aws-sdk-go/aws/credentials (from $GOROOT)
	/home/milad/go/src/github.com/aws/aws-sdk-go/aws/credentials (from $GOPATH)
cmd/tusd/cli/composer.go:15:2: cannot find package "github.com/aws/aws-sdk-go/aws/session" in any of:
	/home/milad/go/src/github.com/tus/tusd/vendor/github.com/aws/aws-sdk-go/aws/session (vendor tree)
	/usr/local/go/src/github.com/aws/aws-sdk-go/aws/session (from $GOROOT)
	/home/milad/go/src/github.com/aws/aws-sdk-go/aws/session (from $GOPATH)
cmd/tusd/cli/composer.go:16:2: cannot find package "github.com/aws/aws-sdk-go/service/s3" in any of:
[...]

Issue #185, opened 2018-05-09
Pass all HTTP request headers to hooks


miladz68:
Suggestion:
It would be a vey nice feautre to pass all request headers in the hooks so they can be used to authorized uploads and downloads.

Acconut:
As many people have asked about this feature, we are currently open to add this functionality.

vith:
Carrying on from https://github.com/tus/tusd/issues/184#issuecomment-388109821

For things like logging, authorization, and authentication it would be helpful to have a way to access the `http.Request` in the context of the tus event/hook or vice versa.

When using tusd as a library, even if you wrap the handler method to access the request, it's difficult to associate that request with the upload ID generated by PostFile, for example.

miladz68:
I will try to take a look at the source code and make a PR if it is OK.

ljufa:
I also need access to headers in request in hook. Metadata approach will work if client has all data needed, which is not always the case.
In my case there is js-client that does not have access to jwt token but 'access token'. In hook I need to call oauth service to resolve jwt by access token, then call other service using jwt token.

Only not sure does it have to be limited to webhooks only?

sagikazarmark:
Instead of passing every header to the hooks, how about making authorization an extension of the core protocol?

Either by using the `Authorization` header or by introducing an `Upload-Authorization` header.

I know that metadata can be used for this purpose, but introducing it as an extension allows custom implementations to verify a JWT token for example on its own, instead of waiting for a pre-create hook.

Acconut:
> Instead of passing every header to the hooks, how about making authorization an extension of the core protocol?

That would also be a possible way but I also see other possibilities in passing request headers. For example, you may examine the `Host` header to see which server the user is uploading to and then make different decisions based on this.

> I know that metadata can be used for this purpose, but introducing it as an extension allows custom implementations to verify a JWT token for example on its own, instead of waiting for a pre-create hook.

I agree that authentication can be done using metadata and in most cases it's the approach I recommend. However, there are two issues with it:

The metadata values (and possible authentication details) can be read by anyone who can send a HEAD request to the upload URL. If this is not acceptable to you, you need to utilize single-use token for creating uploads which is a non-trivial task.

Furthermore, you would like to minimize the changes necessary to your front-end when adding tus uploads. If you already have a middleware to add authentication details to HTTP requests, you would like to reuse this functionality which is currently not easily possible.

sagikazarmark:
[...]

PR #188, opened 2018-05-12
Adds Request headers to HTTP hooks and also adds the missing dependencies

miladz68:
This PR solves both issues #187 and #185 . 
I know that it a bad practice to address 2 seperate issues in a single PR, but I am new to making PRs and didn't plan it ahead.

on issue #185 in which I asked to pass all http request headers to http hooks: 
I tried to solve this by adding a HTTPRequestHeader field to FileInfo struct, then capture headers in the request and pass them when hooks are activated. I also had to modiy tests a little to make sure this work correctly.

on issue #187 I simply added the dependencies to the vendor directory.

again I am new to golang and PRs, so feel free to ask for modificaitons( or spliting this PR to 2 seperate PRs if necessary).
Thanks


Acconut:
First of all, I want to thank you very much for spending your time working on this feature. While I agree that we should implement it, I see a few fundamental flaws in your approach (ignoring the dependency changes):

1. The HTTP headers are added to the FileInfo struct which has semantically nothing to do with the HTTP requests and  I don't see this as a good internal design. On the other hand, I am not sure where else to store this information, so I need to invest more time into this.
2. All HTTP headers from the request to tusd are directly added to the hook request and can overwrite important values which are needed for correct functionality and integrity (e.g. Content-Length, Content-Type, Host)
3. The hook handler is not able to distinguish between which headers have been set by tusd (e.g. Hook-Name) and which have been set by the client (e.g. Authorization). This may be a critical security vulnerability.

My solution would be to not add the client headers to the hook request headers but instead pass them inside the HTTP request body. In this way, there is no confusion about the origin of a single header pair. What do you think?

miladz68:
I will create another PR, I know this is a mess

miladz68:
@Acconut 
firstly forgive my broken English.
The solution that you are proposing is definitely better than mine.
but about the flaw number 1 that you have mentioned, I did not want to use FileInfo at first, but when I passed the the headers as another struct I faced a problem here[(https://github.com/tus/tusd/blob/master/cmd/tusd/cli/hooks.go#L33)]. Basically adding another struct to this line makes hookDataStore not an instance of datastore, and I couldn't find a way to get around this (although I didn't spend much time thinking). 
I will try to look deeper into the code and fix the flaws that you are mentioning

Acconut:
> Basically adding another struct to this line makes hookDataStore not an instance of datastore, and I couldn't find a way to get around this (although I didn't spend much time thinking).

I am not sure what you mean, you should be able to add another property to the hookDataStore struct without violating interfaces. But I don't think it matters anymore. :)

> I will try to look deeper into the code and fix the flaws that you are mentioning

:+1: Feel free to ask if you have more questions

PR #190, opened 2018-05-25
Handle GCS object not found error

vially:
This makes the Google Cloud Storage store consistent with the S3 store implementation by making `GetInfo` return `ErrNotFound` when the object is not found.

This also makes the HEAD request return `404` instead of `500` when attempting to get the upload status of a non-existent file stored on GCS.


Acconut:
That's looking good, thank you very much for this contribution :+1:

PR #191, opened 2018-06-11
Add CORS support for Upload-Defer-Length header

acj:
This is a followup to #182 that adds the `Upload-Defer-Length` header to the `Access-Control-{Allow,Expose}-Headers` lists and addresses feedback from the earlier PR.

cc @Acconut @kvz

### DIFF
--- a/unrouted_handler.go
+++ b/unrouted_handler.go
@@ -174,12 +174,12 @@ func (handler *UnroutedHandler) Middleware(h http.Handler) http.Handler {
 			if r.Method == "OPTIONS" {
 				// Preflight request
 				header.Add("Access-Control-Allow-Methods", "POST, GET, HEAD, PATCH, DELETE, OPTIONS")
-				header.Add("Access-Control-Allow-Headers", "Origin, X-Requested-With, Content-Type, Upload-Length, Upload-Offset, Tus-Resumable, Upload-Metadata")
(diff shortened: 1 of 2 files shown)


unrouted_handler.go
```diff
@@ -461,26 +461,24 @@ func (handler *UnroutedHandler) PatchFile(w http.ResponseWriter, r *http.Request
...
-				if err := handler.composer.LengthDeferrer.DeclareLength(id, info.Size); err != nil {
-					handler.sendError(w, r, err)
-					return
-				}
-			} else {
-				handler.sendError(w, r, ErrInvalidUploadLength)
-			}
-		} else {
+		if !handler.composer.UsesLengthDeferrer {
 			handler.sendError(w, r, ErrNotImplemented)
```
Acconut: Please add an explicit return here.

unrouted_handler.go
```diff
@@ -461,26 +461,24 @@ func (handler *UnroutedHandler) PatchFile(w http.ResponseWriter, r *http.Request
...
 		}
+		if !info.SizeIsDeferred {
+			handler.sendError(w, r, ErrInvalidUploadLength)
```
Acconut: Please add an explicit return here.

Acconut:
Thank you very much :+1:

PR #193, opened 2018-06-14
Allow AWS-SDK to handle gathering credentials

enobrev:
Previously, you could _only_ use environment variables.   This change enables the SDK to handle credentials according to its standard "credential chain".  Per the [docs](https://github.com/aws/aws-sdk-go#configuring-credentials):

> By default the SDK will source credentials automatically from its default credential chain. See the session package for more information on this chain, and how to configure it. The common items in the credential chain are the following:
> * Environment Credentials - Set of environment variables that are useful when sub processes are created for specific roles.
> * Shared Credentials file (~/.aws/credentials) - This file stores your credentials based on a profile name and is useful for local development.
> * EC2 Instance Role Credentials - Use EC2 Instance Role to assign credentials to application running on an EC2 instance. This removes the need to manage credential files in production.

From my testing, this change should not adversely affect existing installs, since the Environment variables will still work.

### DIFF
--- a/cmd/tusd/cli/composer.go
+++ b/cmd/tusd/cli/composer.go
@@ -11,7 +11,6 @@ import (
 	"github.com/tus/tusd/gcsstore"
 
 	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/credentials"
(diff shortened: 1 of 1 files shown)


Acconut:
Great, thank you very much. I added a comment to the README about this and merged :+1:

PR #195, opened 2018-06-22
Updates uploadLength logic in PatchFile

egalpin:
There is currently an issue wherein PATCH-ing a file which had `Upload-Defer-Length` with `Upload-Length` only results in error. I believe it's the case that the `uploadLength` is checked against the `MaxSize` in the handler, which by default is `0.00` (i.e. no limit). This should cover the default case.


Acconut:
Good catch, thank you very much

PR #198, opened 2018-07-16
Add documentation about HTTP Hooks status code

ruv:
HTTP Status Code meaning was missed.

Source code reference: https://github.com/tus/tusd/blob/311cdf253e50a21b12c0cae094ed0c2d20c66d5c/cmd/tusd/cli/hooks.go#L137


Acconut:
Thank you very much :+1:

PR #199, opened 2018-07-25
Add support for concurrent GetObjectSize requests

tab1293:
In production on Vimeo we were seeing an issue where requests to calculate the current offset of an upload were taking an incredibly long time. We discovered that this was only happening for uploads that had a large number of composable parts. A head request to a file with >200 parts was taking over a minute to respond. 

In order to speed this process up, this PR changes the `GetObjectSize` calls in the store's `GetInfo` method to run concurrently. We limit this concurrency with a semaphore that allows at most 32 goroutines to be working on this task at a time. 

This performance improvement required that we pass context to `GCSService` rather than using an internal and shared one. The reason being that if one `GetObjectSize` call in a goroutine fails, we need a way to cancel all subsequent calls. 

cc @justinruggles @vayam @devonbalicki @erunion

### DIFF
--- a/gcsstore/gcsservice.go
+++ b/gcsstore/gcsservice.go
@@ -59,14 +59,14 @@ type GCSReader interface {
 // operations that are required to enable the tus protocol
 // to work with Google's cloud storage.
 type GCSAPI interface {
-	ReadObject(params GCSObjectParams) (GCSReader, error)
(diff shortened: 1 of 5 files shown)


Acconut:
That's a smart idea, thanks for implementing it.

PR #211, opened 2018-10-15
Move to Standard library Context package

bench:
Since december 2016, `context` is part of the  **Standard Go library.** 

It is the same code as `golang.org/x/net/context`.

The only thing that is missing is `ctxhttp` package but it is not used by tusd sources.

This modification implies to give up compatibility with go 1.5 and 1.6. But considering that right now GCSStore and Consul can't work on versions prior to 1.7, defining 1.7 as the minimal version could be a good choice no?

What do you guys think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Windows in various formats of the
 
 ### Compile from source
 
-The only requirement for building tusd is [Go](http://golang.org/doc/install) 1.5 or newer.
+The only requirement for building tusd is [Go](http://golang.org/doc/install) 1.7 or newer.
 If you meet this criteria, you can clone the git repository, install the remaining
 depedencies and build the binary:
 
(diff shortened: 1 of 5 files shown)


Acconut:
Thank you very much for this contribution.

> This modification implies to give up compatibility with go 1.5 and 1.6. But considering that right now GCSStore and Consul can't work on versions prior to 1.7, defining 1.7 as the minimal version could be a good choice no?

I don't think this is true. You should be able to use tusd without GCSStore and Consul dropping the minimum version to Go 1.5.

I am not sure if there are still people using tusd which Go 1.5 or 1.6 but I would like to avoid raising the minimum version for such a small change. I don't think we have a disadvantage when using `golang.org/x/net/context` over the `context` package from the standard library, do we?

That being said, I do like this change and would like to consider it for the next major release of tusd (see https://github.com/tus/tusd/issues/212).

bench:
ok thank you @Acconut for your reply, i got your point.

So let's close this MR and fix this breaking CI when the time comes.

Acconut:
I agree, thank you very much for taking the time to contribute :+1:

Issue #207, opened 2018-10-05
Request to write the info file after upload completes


ldadams:
Just a suggestion regarding how storage systems might better understand when file uploads are complete.  We are using Tus with S3 and for the most part everything is working out very well.  One thing we are having to work around is how best to use S3 events to process newly uploaded files.  Currently the .info file is written before the actual file is uploaded.  This makes it a bit tricker to have an efficient way trigger follow up actions to the actual uploaded file.  S3 allows for suffix and prefix filters for events but it's my understanding that current the s3Store does not allow for configuration of the prefix or "folder" that .info and actual upload files are placed.  Also, the actual uploaded file doesn't have an extension so we aren't able to filter just for something like .file.

The current approach we are taking is to use a post-finish hook to write an SQS message.  This will work fine but would be great to take advantage of the already existing events triggered in S3.

I would be willing to add a feature to configure the s3Stores ability to place info files and actual uploaded files in folders / S3 Prefixes if I might get a bit of guidance on how best to contribute that.  This would allow us to not listen for .info file events and just actual file uploads.

Thanks!

Acconut:
> Request to write the info file after upload completes

One comment to the title: It is not easily possible to write the info file after the upload completes. The info file contains part of the upload state which is needed for resuming (e.g. after the connection is interrupted). Therefore, we need to store the state somewhere persistently which is S3 in this case.

> The current approach we are taking is to use a post-finish hook to write an SQS message.

Personally, this sounds like the best approach to me. tusd knows best when an upload is finished especially when working with concatenating uploads. However, I can also understand that this may not be a good approach for your architecture.

> I would be willing to add a feature to configure the s3Stores ability to place info files and actual uploaded files in folders / S3 Prefixes if I might get a bit of guidance on how best to contribute that. This would allow us to not listen for .info file events and just actual file uploads.

I have never thought about different prefixes for the info and binary files but it makes sense. Only very few people in the past have asked about specifying folders for S3 uploads, so we haven't added support for it. However, I would be happy to help with and accept a PR if you are willing to supply.

ldadams:
Thanks for the response.  We implemented the SQS functionality and works well.  If we decided that we would rather have S3 native support by separating out info and actual files will come back and help put this together.  Thanks again!

PR #213, opened 2018-10-20
Support prefix on S3 object names

acj:
Currently, all objects in S3 are placed in the root of the bucket. This makes tusd adoption difficult for existing applications that use a different convention, such as using S3's support for prefixes to place objects in a pseudo-directory like `s3://my_bucket/media/incoming/<upload id>`. This PR introduces an `ObjectPrefix` field on `S3Store` to enable the use of prefixes.

Addresses one of the requests in #207.

### DIFF
--- a/cmd/tusd/cli/composer.go
+++ b/cmd/tusd/cli/composer.go
@@ -35,6 +35,7 @@ func CreateComposer() {
 		// Derive credentials from default credential chain (env, shared, ec2 instance role)
 		// as per https://github.com/aws/aws-sdk-go#configuring-credentials
 		store := s3store.New(Flags.S3Bucket, s3.New(session.Must(session.NewSession()), s3Config))
+		store.ObjectPrefix = Flags.S3ObjectPrefix
 		store.UseIn(Composer)
 
 		locker := memorylocker.New()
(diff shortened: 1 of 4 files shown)


acj:
@Acconut - when you have time, I'd love to get some feedback on this.

Acconut:
I'm sorry for the delay, I thought I commented on it literally a few hours ago. But maybe I just missed the submit button.

Anyways, I wanted to ask for your reasoning behind this PR. I was wondering about it since the referenced issue #207 took another approach.

acj:
No worries. My reasoning is mostly what I mentioned above — established applications that have a large collection of existing objects in a directory (prefix) may want newly uploaded files to be placed in that same directory. In our case, the directory is a long-used convention that realistically can't be changed. tusd doesn't currently provide a solution for this since all files are placed at the root of the bucket, and there's no way to override that behavior. My goal here is to provide a simple, non-breaking parameter that solves that problem.

Regarding the other PR, we experimented with using a post-upload hook to move/rename the uploaded files into our existing directory, but the operation was far too slow and unpredictable for our use case.

Acconut:
Thank you for the detailed information and the nice PR :+1: I am happy to incorporate this feature into tusd.

ridhozhr10:
guys, i just trying to make this available too in gcs store. and i discover when using prefix with underscore, i get bug because when creating chunk, idx suffix is splitted with underscore.

i think its best to validate prefix name first in startup

PR #216, opened 2018-11-10
etcd3locker: Update formatting of docs; timeout -> ttl substitution

chen-anders:
I saw that godoc didn't render my documentation within the code very well so I fixed up the formatting in the hopes that things look better now.

In addition, I saw various instances of `timeout` and `ttl` when it should be `ttl` so I did a quick correction of places that had `sessionTimeout` / `timeout` to make it all consistent.


Acconut:
Wow, thank you very much for the quick updates :)

PR #220, opened 2018-11-23
Fix typos

lucab85:
Fix typos


Acconut:
Thank you very much! As mostly non-native speakers we do make a lot of spelling mistakes.

lucab85:
I perfectly understand because mee too I'm non-native (Italian). Have a nice day ;-)

PR #225, opened 2019-01-03
refactor(image): switch to a multistage build for the container image

Rio:
This makes it easier to reason about tusd's run environment. It also
significantly lowers the docker image size from 280MB for the current 0.11.0
image to only 26MB. Also not having the entire build environment in the
shipped images makes it quite attractive to use this image as a base
for other images. Those could contain some custom hooks for example.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,12 @@
-FROM golang:1.7-alpine
+FROM golang:1.7-alpine AS builder
 
 # Copy in the git repo from the build context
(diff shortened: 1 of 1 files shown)


Acconut:
Thank you, that is very interesting. I am no Docker expert but I remember reading that putting all commands into a single RUN statement avoids unnecessary layers and reduces the image size.

Is it now recommended to use multistage builds instead? Is this a new feature of Docker?

PR #228, opened 2019-01-09
Handle "NotFound" error code from HeadObject

### LINKED ISSUE #226: The new Creation Defer Length Support Breaks existing configurations
The new changes introduced in https://github.com/tus/tusd/commit/33d12533e79db367bbd432ad912b84b0bdd4ce32 break existing Minio Configurations.
Before this commit, using minio as a s3 server worked flawlessly, but now I'm getting an `404 Not Found` Error on the first PATCH Request and nothing else. It seems like the .info files can't be found (I've checked minio and they do seem to exist and have the correct content).
I'm running the `UnroutedHandler` on go 1.11.4 with the latest version of the tus-js-client & minio/minio:latest.
I'm using no special config options except MaxSize, BasePath and the provided Channels.

acj:
A case that isn't covered by #219 is when the `HeadObject` call to the S3 endpoint returns `NotFound` instead of `NoSuchKey` or `AccessDenied`. Depending on the permissions granted to the request, any of these errors may be returned.

This also accommodates third party implementations of the S3 interface, such as Minio, whose behavior may differ slightly from S3's.

Resolves #226.

### DIFF
--- a/s3store/s3store.go
+++ b/s3store/s3store.go
@@ -364,7 +364,7 @@ func (store S3Store) GetInfo(id string) (info tusd.FileInfo, err error) {
 		Key:    store.keyWithPrefix(uploadId + ".part"),
 	})
 	if err != nil {
-		if !isAwsError(err, s3.ErrCodeNoSuchKey) && !isAwsError(err, "AccessDenied") {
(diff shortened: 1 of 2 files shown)


Acconut:
Thank you very much for this quick fix!

PR #233, opened 2019-01-18
S3Store: IAM fixes for incomplete parts

acj:
This is a followup from #219 and addresses two issues related to IAM permissions. 

First, the use of `HeadObject` as an optimization requires the `s3:ListBucket` permission to be granted, and using both `HeadObject` and `GetObject` to reference .part files was difficult to reason about. This is resolved by using `GetObject` only. ~Second, buckets that are configured with object versioning would have needed the `s3:DeleteObject` and `s3:DeleteObjectVersion` permissions to be granted, which is not ideal and represents a breaking change. By overwriting the existing .part object (if any) with an empty one, we can avoid the complexity associated with deleting objects.~ (See comments below.)

See #226 for more discussion.

### DIFF
--- a/s3store/s3store.go
+++ b/s3store/s3store.go
@@ -137,6 +137,7 @@ type S3API interface {
 	GetObject(input *s3.GetObjectInput) (*s3.GetObjectOutput, error)
 	CreateMultipartUpload(input *s3.CreateMultipartUploadInput) (*s3.CreateMultipartUploadOutput, error)
 	AbortMultipartUpload(input *s3.AbortMultipartUploadInput) (*s3.AbortMultipartUploadOutput, error)
+	DeleteObject(input *s3.DeleteObjectInput) (*s3.DeleteObjectOutput, error)
 	DeleteObjects(input *s3.DeleteObjectsInput) (*s3.DeleteObjectsOutput, error)
 	CompleteMultipartUpload(input *s3.CompleteMultipartUploadInput) (*s3.CompleteMultipartUploadOutput, error)
 	UploadPartCopy(input *s3.UploadPartCopyInput) (*s3.UploadPartCopyOutput, error)
--- a/s3store/s3store.go
+++ b/s3store/s3store.go
@@ -134,7 +134,6 @@ type S3API interface {
 	PutObject(input *s3.PutObjectInput) (*s3.PutObjectOutput, error)
 	ListParts(input *s3.ListPartsInput) (*s3.ListPartsOutput, error)
 	UploadPart(input *s3.UploadPartInput) (*s3.UploadPartOutput, error)
-	HeadObject(input *s3.HeadObjectInput) (*s3.HeadObjectOutput, error)
 	GetObject(input *s3.GetObjectInput) (*s3.GetObjectOutput, error)
 	CreateMultipartUpload(input *s3.CreateMultipartUploadInput) (*s3.CreateMultipartUploadOutput, error)
 	AbortMultipartUpload(input *s3.AbortMultipartUploadInput) (*s3.AbortMultipartUploadOutput, error)
--- a/s3store/s3store.go
+++ b/s3store/s3store.go
@@ -369,19 +369,13 @@ func (store S3Store) GetInfo(id string) (info tusd.FileInfo, err error) {
 		offset += *part.Size
 	}
 
-	headResult, err := store.Service.HeadObject(&s3.HeadObjectInput{
-		Bucket: aws.String(store.Bucket),
-		Key:    store.keyWithPrefix(uploadId + ".part"),
-	})
+	incompletePartObject, err := store.getIncompletePartForUpload(uploadId)
 	if err != nil {
-		if !isAwsError(err, s3.ErrCodeNoSuchKey) && !isAwsError(err, "NotFound") && !isAwsError(err, "AccessDenied") {
-			return info, err
-		}
-
-		err = nil
+		return info, err
 	}
-	if headResult != nil && headResult.ContentLength != nil {
-		offset += *headResult.ContentLength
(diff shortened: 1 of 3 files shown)


acj:
While writing tests for the HeadObject "Forbidden" error, I realized that the behavior of GetInfo would be different from WriteChunk in terms of how the .part object is handled, which could further complicate things. Using GetObject everywhere instead of HeadObject makes the workflow easier to reason about, imo, and it also sidesteps the permissions issues. I've pushed that change.

acj:
> Truncating the part files instead of removing them is a very non-intuitive action and while both approaches may result in the same upload handling, it will probably cause a lot of confusion if the users end up with a massive amount of empty files in their buckets.

I agree. We should try to find a better solution if possible.

> Also, since I have never used object versioning, I would have one question: Is it not possible to delete the most recent object version just using s3:DeleteObject and without s3:DeleteObjectVersion?

No, unfortunately that will result in an AccessDenied error in the case of a versioned object. If the request has both s3:DeleteObject and s3:DeleteObjectVersion permissions, then S3 will insert a delete marker as the most recent version and will return 404 for subsequent GetObject requests. If the application doesn't delete each version of the object, then the bucket owner is responsible for setting up an expiration policy to purge deleted versions.

Maybe it's safe to assume that tusd users who have object versioning enabled will already have included the s3:DeleteObjectVersion permission, which would be required for the Terminate extension to work with the current setup (prior to #219). My only concern is that there may be existing deployments that use versioned objects but don't use Terminate, and the new requirement of s3:DeleteObjectVersion will be a breaking change for them. Does that seem acceptable?

Acconut:
> No, unfortunately that will result in an AccessDenied error in the case of a versioned object. If the request has both s3:DeleteObject and s3:DeleteObjectVersion permissions, then S3 will insert a delete marker as the most recent version and will return 404 for subsequent GetObject requests. If the application doesn't delete each version of the object, then the bucket owner is responsible for setting up an expiration policy to purge deleted versions.

Ok, thank you very much for clarifying this. I have never used versioning for S3, so that was really helpful.

> Maybe it's safe to assume that tusd users who have object versioning enabled will already have included the s3:DeleteObjectVersion permission, which would be required for the Terminate extension to work with the current setup (prior to #219). My only concern is that there may be existing deployments that use versioned objects but don't use Terminate, and the new requirement of s3:DeleteObjectVersion will be a breaking change for them. Does that seem acceptable?

Yes, I think this is acceptable for now. This compromise allows us to get your PR merged quickly and fix the problems other people have encountered as well. If we then see that many people have issues with upgrading tusd, we can act accordingly (maybe we can even include a test into tusd for checking whether it has the correct permission in the future). Does this sound good for you?

acj:
Sounds good to me. I've removed the empty-object piece from this PR, fixed a related issue that was causing DeleteObjects errors not to be returned properly, updated the docs, and rebased onto master.

acj:
@Acconut I've gotten a report of `mismatched offset` errors that may be related to these changes (or perhaps the earlier PR). I'm investigating and will post back soon.

acj:
I traced the problem to inconsistent behavior between DeleteObject and DeleteObjects on versioned buckets. The response from DeleteObjects seems to indicate that a delete marker has been added for the object:

```
{
  DeleteMarker: true,
  VersionId: "[...]"
}
```

but the object can still be fetched with GetObject, and the AWS console doesn't show a delete marker for it. This behavior isn't documented in the S3 docs, AFAICT, and I don't see any missing permissions in my configuration that would explain it. Because the .part object is still accessible, GetInfo is including it in its offset calculation and returning an incorrect value -- hence the mismatched offset error.

However, using DeleteObject (singular) creates a delete marker as expected and has the benefit of being simpler. I've updated this branch to use it.

If anyone is using an S3 bucket with object versioning enabled and is willing to help out, I would be grateful for the extra testing and eyeballs.

acj:
I've tested the latest changes on a versioned bucket with many different chunk sizes and haven't had any issues.

The requirement for `s3:DeleteObjectVersion` doesn't seem to apply to `DeleteObject`, so we could remove that mention from the docs and error logging if we want. That addresses the concern that I had earlier. (It's also possible that my prior config was to blame, or I was using DeleteObjects incorrectly -- not sure. It's been challenging to track the various configurations and permissions.)

Since there's been a bit of churn here, this is the current state of things:

 * Replace HeadObject with GetObject, which simplifies the set of IAM permissions that are required and makes S3Store's behavior easier to reason about.
 * Replace DeleteObjects with DeleteObject, which simplifies the set of IAM permissions that are required and also simplifies the code a bit.

@Acconut, how are you feeling about this PR?

Acconut:
Wow, thank you very much for your research! It is astonishing how different the DeleteObject and DeleteObjects methods are. Have you talked to a person from AWS to confirm if this behavior is intentional and if we can rely on this (undocumented?) behavior?

Other than that, I think this PR is looking good and, unless you have remaining todos, I would be open for merging this.

Also, apologies for my delay and short comments, private life is getting in the way.

acj:
> Have you talked to a person from AWS to confirm if this behavior is intentional

I haven't, but I would like to chat with them about it. The issue with s3:DeleteObjectVersion might have been confusion on my part, but I couldn't find an explanation for the delete marker behavior with DeleteObjects. It's puzzling and concerning.

> and if we can rely on this (undocumented?) behavior?

The DeleteObject behavior that I'm seeing is consistent with the official AWS docs, so I think we're on solid ground now.

> unless you have remaining todos

The only remaining item was to remove the mention of s3:DeleteObjectVersion from the comments and error message, which I've just done. I think it's ready.

> Also, apologies for my delay and short comments, private life is getting in the way.

No worries, and I hope all is well. Thanks for your patience and feedback.

PR #236, opened 2019-02-02
Forward status code and body from Pre-create http hook

### LINKED ISSUE #170: Pre-create (http) hook should be able to set a specific status code
When the http hook returns an error code (4xx-5xx), then there is currently no mechanism to return a specific status code from tusd server other then 500 Internal Server Error. Quote from the documentation: "a non-zero exit code tells tusd to reject the request with a 500 Internal Server Error response containing the process' output from stderr."
I think, it would be better to use the response code and payload (message) from the http hook and forward is as a response to the user/client.
That way if there was a user error (4xx codes - bad request), then the user would know what to fix in the request and could send it again. Sending the original message would direct the user what and how to fix in the request.
Or the user would know that there is something wrong with the server (5xx codes).

oliverpool:
Fix #170

### DIFF
--- a/cmd/tusd/cli/hooks.go
+++ b/cmd/tusd/cli/hooks.go
@@ -32,7 +32,11 @@ type hookDataStore struct {
 
 func (store hookDataStore) NewUpload(info tusd.FileInfo) (id string, err error) {
 	if output, err := invokeHookSync(HookPreCreate, info, true); err != nil {
-		return "", fmt.Errorf("pre-create hook failed: %s\n%s", err, string(output))
+		hookErr := fmt.Errorf("pre-create hook failed: %s\n%s", err, string(output))
+		if statusErr, ok := err.(tusd.HTTPError); ok {
+			return "", tusd.NewHTTPError(hookErr, statusErr.StatusCode())
+		}
+		return "", hookErr
 	}
 	return store.DataStore.NewUpload(info)
 }
@@ -144,7 +148,7 @@ func invokeHttpHook(name string, typ HookType, info tusd.FileInfo, captureOutput
 	}
 
 	if resp.StatusCode >= http.StatusBadRequest {
-		return body, fmt.Errorf("endpoint returned: %s\n%s", resp.Status, body)
+		return body, tusd.NewHTTPError(fmt.Errorf("endpoint returned: %s\n%s", resp.Status, body), resp.StatusCode)
 	}
 
 	if captureOutput {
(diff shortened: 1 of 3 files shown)


Acconut:
This is looking good, thank you very much! I only have three comments:

1. This PR is marked as a WIP. Do you still plan on adding more commits to it?
2. Would you mind updating the documentation (https://github.com/tus/tusd/blob/master/docs/hooks.md#http-hooks) to include a short sentence about this feature?
3. Could you post the response body of a POST request if the pre-create hook fails, just to see what it looks like?

oliverpool:
1. I marked it WIP, because I wrote it blind, without actually testing it. I just tested it and added a commit
2. Sure, will update the doc, as soon as we are clear regarding 1. :-)
3. I just updated my code, to forward the response of the hook. Before that, it returned:
```
pre-create hook failed: 403 Forbidden
[original reply of the hook]
```
But if the hook returned some JSON, then it was all messed up (because of the sentence pretended to the response).

---

To sum up the state of this PR now:
- if the hook response is OK, then this response is discarded an the upload proceeds
- if the hook response is not OK, this response is forwarded as-is to the client (keeping the returned status code)
The tus server logs 2 lines:
```
event="HookInvocationError" ... error="endpoint returned: {Code} {CodeStatus}" // the body is not logged
event="ResponseOutgoing" ... error="pre-create hook failed: endpoint returned: {Code} {CodeStatus}" // the body is not logged
```

I choose not to log the body inside tus, because:
- it might be multiple lines (which srews the logs a bit)
- I think it is more the responsibility of the hook to know why it failed (for tus the metrics are more interesting I guess)

Acconut:
> * if the hook response is OK, then this response is discarded an the upload proceeds
> * if the hook response is not OK, this response is forwarded as-is to the client (keeping the returned status code)

Ok, sounds good :+1: 

> I choose not to log the body inside tus.

I agree with that decision to keep the log but exclude the hook response body.

The only thing I am a bit concerned about is the extension of the HTTPError interface. Since it's public (on purpose), some people may use it in their applications. But that's probably too few people, so I hope it should be fine.

So the only missing part is documentation, isn't it?

oliverpool:
> The only thing I am a bit concerned about is the extension of the HTTPError interface. Since it's public (on purpose), some people may use it in their applications. But that's probably too few people, so I hope it should be fine.

You are right, that this is actually a breaking change.

> So the only missing part is documentation, isn't it?

I've just updated it :-)

oliverpool:
Always a pleasure to contribute to this project (thanks to the helpful and reactive maintainers :-)

Acconut:
Ah, that's very nice to hear :) Thank you

PR #237, opened 2019-02-04
Docs: consullocker/etcd3locker - correct UnlockUpload method docs

chen-anders:
The following statement in the docs is not accurate for consullocker and etcd3locker. 

> If no such lock exists, no error will be returned. 

This PR removes that statement for both locker implementations.


Acconut:
You're right, thank you very much!

Issue #239, opened 2019-02-05
Add option to disable post-receive hook


brandonkal:
When using HTTP hooks, it would be nice to disable the post-receive hook especially when the receiving server is low power (think raspberry pi) or is slow to respond.

For instance, I have a system with a slow REST API response time (500ms). A single upload can easily tax the CPU with moderate web traffic. My current  solution is to create a seperate php script on this server that checks the event type and calls its actual REST API if the event is not a post-receive. This chained approach works, but a simple cli option to simply not send the HTTP POST would be an appreciated simplification.

Acconut:
I can understand your situation, thanks for bringing this up. Let's discuss implementation details in your PR.

Acconut:
A whitelist for hook events has been added in https://github.com/tus/tusd/commit/f1fe5e26034504f084c97cd53cccc65271425095.

PR #240, opened 2019-02-06
Post receive disable

brandonkal:
This PR adds the option to disable to the post-receive hook via a new command line flag. It remains enabled by default.

Related: #239

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 tusd/data
+/tusd
 cover.out
 data/
 node_modules/
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 <img alt="Tus logo" src="https://github.com/tus/tus.io/blob/master/assets/img/tus1.png?raw=true" width="30%" align="right" />
 
-> **tus** is a protocol based on HTTP for *resumable file uploads*. Resumable
+> **tus** is a protocol based on HTTP for _resumable file uploads_. Resumable
 > means that an upload can be interrupted at any moment and can be resumed without
 > re-uploading the previous data again. An interruption may happen willingly, if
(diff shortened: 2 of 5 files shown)


brandonkal:
Note that most of the doc changes are the result of Prettier cleaning things up. If this is an issue, feel free to edit this PR or let me know. I have made the doc changes its own commit to make review easier.

Acconut:
Thank you very much for this PR!

It would be nice if you could undo the unrelated documentation changes as they make reviewing your patch a bit harder. Anyways, I think if we go ahead and offer a solution to disable the post-receive hook, we can also go a step further and allow users to disable/enable any event to they want. However, I am unsure whether this should be a blacklist (similar to your PR) or a whitelist where the user must explicitly activate each event.

The usage could look like this:
```sh
tusd -dir ./data -hooks-dir ./hooks -disable-hook-events post-receive,pre-create

# or with a whitelist

tusd -dir ./data -hooks-dir ./hooks -enable-hook-events post-create,post-finish
```

What do you think about this?

Acconut:
> I would prefer the blacklist approach if it went that route as my docker-compose's command option is already quite long.

I can understand your reasoning but a whitelist is probably better in terms of future compatibility. If we are going to add another hook in the future, people to not have to opt out of it when they upgrade their tusd version.

> This is my first time hacking with golang so I am not sure how to go about parsing a string list like that.

I am happy to assist if you want to give it a try. The `strings` package (https://golang.org/pkg/strings/) should include all necessary tools. You basically just have to split the parameter at the comma character and eventually trim all additional whitespace.

Acconut:
A whitelist for hooks has been added in https://github.com/tus/tusd/commit/f1fe5e26034504f084c97cd53cccc65271425095.

PR #253, opened 2019-03-18
remove unnecessary goroutines

kshvakov:


Acconut:
Would you mind explaining the reasons behind these changes?

kshvakov:
It makes no sense to run the atomic update in the goroutines, it does not speed up the execution time,
and in some cases can only increase it. Atomic update is fast and no blocking. 


'Select' prevents block when sending event into the UploadProgress channel, because there may be no reader from the channel. 
https://github.com/tus/tusd/pull/253/files#diff-4e64569f67c7c0214e072f52dbe8d504R833

Acconut:
I agree with removing the goroutines from those small functions.

However, I am not on the same page with you about the `select` blocks. If you don't want to receive upload progress event, you can disable them all together by setting `NotifyUploadProgress` to false. Otherwise, I think it's the handlers job to ensure that the channel is either buffered or always free to avoid blocking. Do you disagree here?

kshvakov:
@Acconut I think you are right.

kshvakov:
@Acconut What are we going to do with this PR? I can create a new PR with only goroutines fixes.

Issue #264, opened 2019-04-11
Unix socket binding


Xowap:
**Is your feature request related to a problem? Please describe.**
My deployment architecture allows me to map Git branches to deployment environments. Such environments are created automatically, many residing on the same server. Each of those environments needs one instance of each component (web server, database, and in this case `tusd`).

In order to multiplex the access to those services, most of the time using a different HTTP `Host` is sufficient. This is done by a `nginx` reverse proxy on each server that handles the different name-based virtual hosts, the HTTPS and various details.

If all my services used TCP ports, I would have to do a lot of book keeping to know which service uses which port. In this case, if I have to bind `tusd` to a TCP port, I have to keep track of all the TCP ports it's being bound to in relation to which Git branch and this which vhost.

**Describe the solution you'd like**
What I do is bind my services to Unix sockets. This way I can bind sockets to predictable locations based on the project and branch name, like `/run/{project}/{component}/{branch}.sock`.

I'm thinking about a new `tusd` option which would be `-unix-sock` and would signify to discard `-host` and `-port` while binding to the specified path.

**Describe alternatives you've considered**
Other options for me would include

- A port/project/service book keeper
- A wrapper around the binary to redirect a Unix socket to the TCP port (using a different netns or something like that)

But they seem largely overkill in comparison to an implementation that should be a 15-lines diff.

**Can you provide help with implementing this feature?**
I'm in an exploratory stage for my project (which should conclude within hours, very tight deadline). If I need so (80% chances) then I'll implement it myself and make a PR, unless someone does it before the next 3 hours (which I certainly don't expect). Then I'll happily implement the tests, documentation or any other requirements the project might have. Unless the feature doesn't interest you of course.

Xowap:
Fixed with pull request actually

PR #265, opened 2019-04-12
Create CLI option to bind to UNIX socket

Xowap:
As suggested in tus/tusd#264, this PR brings the ability for `tusd` to bind to a UNIX socket.

The implementation adds a `-unix-sock` CLI option which if set to a non-empty value will bind `tusd` to the provided path and discard the TCP host and port.

I'm unsure about the standards of the project, so please me how I can improve this.

### DIFF
--- a/cmd/tusd/cli/flags.go
+++ b/cmd/tusd/cli/flags.go
@@ -8,6 +8,7 @@ import (
 var Flags struct {
 	HttpHost          string
 	HttpPort          string
+	HttpSock          string
 	MaxSize           int64
 	UploadDir         string
 	StoreSize         int64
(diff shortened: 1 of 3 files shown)


Acconut:
Thank you for this PR, it looks very good, all in all. I will try it on my own in the next few days and come back if any problems occur. In the meantime, do you know how this will behave on Windows (or other platforms not supporting UNIX sockets)?

Xowap:
Good!

Regarding the Windows support, that's an excellent question, I suppose that an error will be raised by the go standard library. I've found [this issue](https://github.com/golang/go/issues/26072) which suggests it's going to fail yet it's going to be implemented one day. As it's specific to the implementation of Go, I'd suggest to keep the option available in Windows so that if the code is compiled on a compatible implementation then it automatically works. Though it might get confusing, maybe do that and document it in the help message?

In any case, I'll have to test it.

Xowap:
Awesome :+1:

Acconut:
I am currently working on a few smaller improvement to tusd, so the UNIX socket support will be shipped in the new release in a few days. And thank you again for the contribution

PR #275, opened 2019-05-22
support object prefix on GCS object names

ridhozhr10:
reference : #213 

since prefixing object name available in s3 object, i think its good to be able to prefix object name in gcs too.
thanks @acj for making that so i can replicate it for GCS store

### DIFF
--- a/gcsstore/gcsstore.go
+++ b/gcsstore/gcsstore.go
@@ -45,10 +45,11 @@ type GCSStore struct {
 
 // New constructs a new GCS storage backend using the supplied GCS bucket name
 // and service object.
-func New(bucket string, service GCSAPI) GCSStore {
+func New(bucket string, objPrefix string, service GCSAPI) GCSStore {
 	return GCSStore{
-		Bucket:  bucket,
-		Service: service,
+		Bucket:       bucket,
+		ObjectPrefix: objPrefix,
+		Service:      service,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


gcsstore/gcsstore.go
```diff
@@ -32,17 +33,23 @@ type GCSStore struct {
...
+
 	// Service specifies an interface used to communicate with the Google
 	// cloud storage backend. Implementation can be seen in gcsservice file.
 	Service GCSAPI
 }
 
 // New constructs a new GCS storage backend using the supplied GCS bucket name
 // and service object.
-func New(bucket string, service GCSAPI) GCSStore {
+func New(bucket string, objPrefix string, service GCSAPI) GCSStore {
```
Acconut: Adding the objPrefix argument is unfortunately as breaking change and we would like to avoid them for now. However, I think the caller should be able to set the ObjectPrefix after calling New, so the argument is now needed actually:

```
store := gcsstore.New("bucket", gcsService)
store.ObjectPrefix = "prefix/"
```

Acconut:
Thank you for this PR, it's looking very good. However, would you mind adding a test where an object prefix is defined?

ridhozhr10:
> Thank you for this PR, it's looking very good. However, would you mind adding a test where an object prefix is defined?

thanks for reviewing, and i already committed new change. 
can you help me how to figure out that conflict in master branch ? since i fork this repo to my account.

Acconut:
Thank you very much, this has been merged in https://github.com/tus/tusd/commit/53b86abc354d9a8fb8f7407529cffc2185892261.

PR #278, opened 2019-05-26
Add original request to info to be passed to hooks

dourvaris:
Implements passing the original request to hooks. closes https://github.com/tus/tusd/issues/185

Review appreciated, not sure if approach is correct but couldn't see a better place to add the OriginalRequest except to the FileInfo since that's what's being sent to the hooks - the alternative would be to create a separate channel for hooks which seemed like overkill.

Example payload - maybe should remove some of those tus headers like`Upload-Metadata` as they are redundant?
```json
{
  "ID": "",
  "Size": 626,
  "SizeIsDeferred": false,
  "Offset": 0,
  "MetaData": {
    "content_type": "text/plain",
    "filename": "README.md"
  },
  "IsPartial": false,
  "IsFinal": false,
  "PartialUploads": null,
  "OriginalRequest": {
    "Host": "localhost:1080",
    "Proto": "HTTP/1.1",
    "RequestURI": "/files/",
    "RemoteAddr": "127.0.0.1:58512",
    "Headers": {
      "Accept": [
        "*/*"
      ],
      "Accept-Encoding": [
        "gzip, deflate"
      ],
      "Connection": [
        "keep-alive"
      ],
      "Content-Length": [
        "0"
      ],
      "Tus-Resumable": [
        "1.0.0"
      ],
      "Upload-Length": [
        "626"
      ],
      "Upload-Metadata": [
        "content_type dGV4dC9wbGFpbg==,filename UkVBRE1FLm1k"
      ],
      "User-Agent": [
        "python-requests/2.20.1"
      ]
    }
  }
}
```

### DIFF
--- a/concat_test.go
+++ b/concat_test.go
@@ -32,20 +32,12 @@ func TestConcat(t *testing.T) {
 
 	SubTest(t, "Partial", func(t *testing.T, store *MockFullDataStore) {
 		SubTest(t, "Create", func(t *testing.T, store *MockFullDataStore) {
-			store.EXPECT().NewUpload(FileInfo{
-				Size:           300,
-				IsPartial:      true,
-				IsFinal:        false,
-				PartialUploads: nil,
-				MetaData:       make(map[string]string),
-			}).Return("foo", nil)
-
 			handler, _ := NewHandler(Config{
 				BasePath:  "files",
 				DataStore: store,
 			})
 
-			(&httpTest{
+			test := &httpTest{
 				Method: "POST",
 				ReqHeader: map[string]string{
 					"Tus-Resumable": "1.0.0",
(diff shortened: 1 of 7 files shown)


Acconut:
Thank you very much for this PR and addressing an often requested feature.

> not sure if approach is correct but couldn't see a better place to add the OriginalRequest except to the FileInfo since that's what's being sent to the hooks - the alternative would be to create a separate channel for hooks which seemed like overkill.

I can see where that's coming from, however, I am not sure if the FileInfo struct is the correct place for information about the original request for two reasons:
- The FileInfo struct contains data about the uploaded file, so semantically the original request does not really belong there.
- By adding the OriginalRequest property, the datastores will save the original request in their backends  which will then also always be loaded for every request. For example, the FileStore puts the original request into its `.info` file, where this information does not really belong.

Out of curiosity, what idea with channels did you have in mind? Could you explain this thought as it might be better suited?

> maybe should remove some of those tus headers likeUpload-Metadata as they are redundant?

I am not too concerned about this and think that this would just cause more effort than necessary.

dourvaris:
Thanks for confirming that the approach is bad.

> Out of curiosity, what idea with channels did you have in mind? Could you explain this thought as it might be better suited?

Essentially convert the hook channels to take "Event" types instead of FileInfo:
```golang
type struct CreatedUploadEvent {
  FileInfo FileInfo
  Request *http.Request
}
if handler.config.NotifyCreatedUploads {
   handler.CreatedUploads <- CreatedUploadEvent{
        FileInfo: info,
        Request: req,
        // possibly other data here}
}
```

Acconut:
>  Essentially convert the hook channels to take "Event" types instead of FileInfo:

That's a great idea! Since it would be a backwards-compatibility breaking change, I added it to the list of possible changes for the next tusd major version.

PR #281, opened 2019-06-12
adopted change in prometheus client lib.

thirsch:
I was not able to build the docker container unless doing the changes contained in this pr. Is it correct? Has the handler been in another package before?

### DIFF
--- a/cmd/tusd/cli/metrics.go
+++ b/cmd/tusd/cli/metrics.go
@@ -7,6 +7,7 @@ import (
 	"github.com/tus/tusd/prometheuscollector"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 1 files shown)


Acconut:
It seems correct, yes. As far as I can tell, `prometheus.Handler` was deprecated in favor of `promhttp.Handler` and got finally removed from Prometheus two days ago (see https://github.com/prometheus/client_golang/commit/761a2ff07c763475bff727730e823a8f4d501f36#diff-b83b53c6fd7712ee815387960471957bL53). That's why you are likely the first person to notice this.

Your PR looks good and should fix it, thank you very much. We have plans to improve our dependency management in the future to prevent situations like this.

PR #282, opened 2019-06-12
Set the docker user to tusd.

thirsch:
User tusd has been created but is not selected for the tusd process within the container. Fixed in this Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,3 +33,5 @@ WORKDIR /srv/tusd-data
 EXPOSE 1080
 ENTRYPOINT ["tusd"]
 CMD ["--hooks-dir","/srv/tusd-hooks"]
+
+USER tusd


kvz:
@kiloreux looks good to you?

Acconut:
Thank you for this change. I am not Docker expert so would you mind explaining if it's actually necessary to have a dedicated tusd user inside the Docker container? Does that provide a benefit or makes the setup just more complicated?

kiloreux:
Look good to me. 
Basically we do this to limit permissions of the container process to that specific user instead of the default user (this one comes from the original image).

thirsch:
@kiloreux, thanks for approving my pr. Who is managing the docker account? Looks like there is no autobuild in place, as the new version has not been built yet.

Acconut:
We manage the tusproject/tusd Docker image and it has autobuilds enabled but they are currently failing: https://cloud.docker.com/u/tusproject/repository/registry-1.docker.io/tusproject/tusd/builds/68adb8d5-e08e-4f0e-8630-1de78eb9e617

Since I don't have the time to look into this currently, would you mind giving the failing Docker build a try? That would help us a lot.

thirsch:
> We manage the tusproject/tusd Docker image and it has autobuilds enabled but they are currently failing: https://cloud.docker.com/u/tusproject/repository/registry-1.docker.io/tusproject/tusd/builds/68adb8d5-e08e-4f0e-8630-1de78eb9e617
> 
> Since I don't have the time to look into this currently, would you mind giving the failing Docker build a try? That would help us a lot.

Thanks for your feedback. I'd be happy to help. Unfortunately, I'm only getting a 404. Maybe because I have no permissions to the repo? My username at docker hub is thirsch as well.

Acconut:
> Unfortunately, I'm only getting a 404. Maybe because I have no permissions to the repo? My username at docker hub is thirsch as well.

Good call, thank you very much! I added you to tusproject/tusd, so you should be able to access the build output. Does it work for you now?

PR #284, opened 2019-06-23
Installing gcc and libc-dev for having a matching gcc and linker in the image.

thirsch:
This fixes the issues with the docker build process.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ COPY . /go/src/github.com/tus/tusd/
 WORKDIR /go/src/github.com/tus/tusd
 
 RUN apk add --no-cache \
-        git \
+        git gcc libc-dev \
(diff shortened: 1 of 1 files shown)


thirsch:
Somehow the base image must have been changed in the last few days and the compiler and linker does not match against each other. By installation gcc and libc-dev manually, we can enforce the correct versions for the build process. Another option would have been to disable native compilation by setting CGO_ENABLED = 0.

thirsch:
Perfect, it's working now. Could you also apply a new version tag if the master is stable from your point of view or should we create a new branch for having 0.13.0 + the fixes as 0.13.1, because 0.13.0 couldn't be built on docker as well.

Acconut:
Of course, the release is on its way: https://github.com/tus/tusd/releases/tag/0.13.1

PR #289, opened 2019-07-01
Added script to generate official-images/library/tusd file.

thirsch:
As promised, here is the generator-script to create a new official-images tusd file. It automatically builds docker tags based on git tags and starts with version 0.13.1. Created and tested on macOS.

### DIFF
--- a/.scripts/generate-docker-library.sh
+++ b/.scripts/generate-docker-library.sh
@@ -1,5 +1,11 @@
 #!/usr/bin/env bash
 
+#
+# official-images tag file generator
+#
+# usage: ./generate-docker-library.sh > [official-images-folder]/library/tusd
+#
+
 cat <<-EOH
 # This file is generated via https://github.com/tus/tusd/blob/master/generate-docker-library.sh
 Maintainers: tus.io (@tus), Thomas A. Hirsch (@thirsch)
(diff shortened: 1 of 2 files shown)


Acconut:
Thank you very much for adding this but I am a bit lost here. Could you quickly explain what this script is for and how it's sued? And is there a reason why it is located in this repository and not inside https://github.com/tus/official-images? Also, could you move this file into the `.scripts/` folder?

One last thing: We may be start prefixing our version tags with `v`, such as `v1.3.4` after the next major release. Would this script need modifications to handle those tags then?

thirsch:
> Thank you very much for adding this but I am a bit lost here. Could you quickly explain what this script is for and how it's sued? And is there a reason why it is located in this repository and not inside https://github.com/tus/official-images? Also, could you move this file into the `.scripts/` folder?

Sure, sorry for not explaining this while opening the PR. This script generates the official-images tag file https://github.com/tus/official-images/blob/master/library/tusd and has to be in the tusd repository to be able to list all git tags. We can't place the script in the official-images repo, because we need the ability to list the git tags of the tusd repo and there is no particular folder to have custom scripts there. 

> One last thing: We may be start prefixing our version tags with `v`, such as `v1.3.4` after the next major release. Would this script need modifications to handle those tags then?
Ok, thanks for the info. I'll test it and adjust the script if necessary.

Issue #173, opened 2018-03-28
How to get the absolute url of s3 when using s3 ad dataStore?


samyakjain:
Currently i am getting url as 

http://localhost:1333/files/da222199e55445353d1bd5341332cab0+VM8tESxKGHTy0wJJt_44OX1pzUHlQIdmB4jUS7PqA10B9ZuxRN6QaUsj.pyAT4BWfETjTFM.8ZFfxYcCCzAm7.h2m0HXkMpPq53Vf_xV8Yxnbpbu7z2x4d_4TCfaig.E

but i want is absolute url to s3. Something like 
https://s3-ap-southeast-1.amazonaws.com/<bucket_name>/some-id

Acconut:
Currently, you would have to generate the absolute URL on your own. The tusd upload ID (the last segment in the upload URL) consists of the file ID and multipart ID separated by a `+`. The multipart ID is of no use after the upload is finished but you can use the file ID to generate the S3 URL.

In your case, the URL would be 
https://s3-ap-southeast-1.amazonaws.com/bucket_name/da222199e55445353d1bd5341332cab0

samyakjain:
I eventually figured that out. 

Thanks for the reply.

zvictor:
is there any place in the response object where we can get only the file ID? I can only find `uploadURL` and, I have to say, it feels quite hacky to have to break it down in order to extract the ID.

should we change it to include the id and maybe more?

```js
{
  ...
  tus: {
    uploadURL: '...',
    id: '...',
    bucket: '...',
    storage: '...'
  }
}
```

Acconut:
@zvictor Unfortunately, is no such functionality yet. We do have plans to expose for storage details (such as the upload ID, bucket, etc) to hooks but it's not implemented. That being said, I am happy to assist you if you want to give this feature a try!

kyleparisi:
For lazy reference:

```js
uploadURL.split("/").slice(-1)[0].split("+")[0]
```

PR #301, opened 2019-08-14
S3 multipart id fix

arbakker:
The upload identifier of a completed upload, when tusd is configured to use s3 as storage backend, will look like `${upload_identifier}+${multipart_identifier}` (I am using the following [script](https://gist.github.com/arbakker/4d7b0870d5f62f87ff13c40589788023) to create the upload). The upload identifier (the field `ID`) is returned both returned in the `.info` metadata file and the JSON payload of the webhooks:

```json
{
    "ID": "832839f804c60eb680525504c12df04b+5054f51f-3722-441c-b9cd-64e59d22dfae",
    "Size": 2656287,
    "SizeIsDeferred": false,
    "Offset": 2656287,
    "MetaData": {
        "filename": "my_file.something"
    },
    "IsPartial": false,
    "IsFinal": false,
    "PartialUploads": null
}
```

When using local file storage as storage backend the `ID` field only contains the upload-identifier (`${upload_identifier}`). According to this [comment](https://github.com/tus/tusd/issues/173#issuecomment-377012396), the multipart identifier is of no use after the upload is finished. In conclusion; this seems like a bug to me. 

Included in the pull request is fix for this behavior and contains checks whether the identifier contains a multipart id and strips it (only when the upload completes). The fix also rewrites `.info` file after the upload is completed, to strip the multipart id from the the `.info` file.

### DIFF
--- a/s3store/s3store.go
+++ b/s3store/s3store.go
@@ -517,6 +517,18 @@ func (store S3Store) FinishUpload(id string) error {
 			Parts: completedParts,
 		},
 	})
+	if err != nil {
+		return err
+	}
+
+	// rewrite ID field in .info metadata file to contain only upload id
+	info, err := store.GetInfo(id)
+	if err != nil {
+		return err
+	}
+
+	info.ID = uploadId
+	err = store.writeInfo(uploadId, info)
 
 	return err
 }
--- a/unrouted_handler.go
+++ b/unrouted_handler.go
@@ -326,6 +326,10 @@ func (handler *UnroutedHandler) PostFile(w http.ResponseWriter, r *http.Request)
 		info.Offset = size
 
 		if handler.config.NotifyCompleteUploads {
+			// strip multipartupload identifier from upload identifier (s3 storage)
+			if strings.Contains(info.ID, "+") {
+				info.ID = strings.Split(info.ID, "+")[0]
+			}
 			handler.CompleteUploads <- info
 		}
 	}
@@ -609,6 +613,10 @@ func (handler *UnroutedHandler) finishUploadIfComplete(info FileInfo) error {
 
 		// ... send the info out to the channel
 		if handler.config.NotifyCompleteUploads {
+			// strip multipartupload identifier from upload identifier (s3 storage)
+			if strings.Contains(info.ID, "+") {
+				info.ID = strings.Split(info.ID, "+")[0]
+			}
 			handler.CompleteUploads <- info
 		}
 


Acconut:
Thank you for your efforts but I don't consider this a bug. This is very much the intended behavior. The upload ID from S3 storage always consists of the object ID and the multipart upload ID, so omitting some information in the upload info is not correct. An ID must be unique and must not change over time. Even if the multipart upload ID is of little use, I don't consider this helpful. Or am I missing something?

arbakker:
I consider this a bug because there is a discrepancy between the resulting filename/upload_id (`${upload_id}` ) and the upload id communicated through the metadata .info file and the json hook payloads (`${upload_id}+${multipart_id}`.

So the uploaded file is saved on S3 with key `${upload_id}`, while the metadata .info and json payload indicate the identifier is `${upload_id}+${multipart_id}`.

This is not the case for other storage backends (at least for local file storage). I would expect tusd to show consistent behavior no matter the storage backend used. 

I see two solutions for ensuring consistent behavior when using a S3 storage backend:

- change the filename of the uploaded file to the `${upload_id}+${multipart_id}` scheme
- stripping the multipart upload id from the json payload in post-finish event and from the metadata .info so the identifier looks like everywhere `${upload_id}` (this pull request)

To be honest I do not understand why only the S3 storage is only using the multipart upload identifier; are multipart uploads only available on TUSD in combination with an S3 backend?

Acconut:
Thanks for the clarification, I can understand where you are coming from.

> I consider this a bug because there is a discrepancy between the resulting filename/upload_id (${upload_id} ) and the upload id communicated through the metadata .info file and the json hook payloads (${upload_id}+${multipart_id}.

That's right but we never promise that the upload's filename will match the upload ID (please correct me if we wrote that somewhere). I know it's inconvenient that the file has a different name than upload ID but we have plans to address this. In a future release, the .info file and hook payload will include additional data specifying where the file has been stored exactly (e.g. file path, S3 bucket, object name). So then this should not be a problem anymore.

>  To be honest I do not understand why only the S3 storage is only using the multipart upload identifier; are multipart uploads only available on TUSD in combination with an S3 backend?

The multipart uploads are an implementation detail of how the S3 storage is built. As a user, you don't need to be concerned about it and the use of AWS multipart upload, don't change the functionality of tusd. Furthermore, the multipart ID is embedded in the upload ID to allow for faster processing since we don't have to lookup the multipart ID somewhere else.

arbakker:
Thanks for the explanation, good to know the reasoning behind the implementation. In hindsight I should have started with opening a issue instead of immediately starting on a PR...  You live and learn :smile:.

Great to hear you are planning on addressing this issue in a future release. Do you have estimate of the timeline for this release?

Acconut:
In most cases it's helpful to open an issue first, even if it's just to discuss what's the best approach for a potential fix :)

> Great to hear you are planning on addressing this issue in a future release. Do you have estimate of the timeline for this release?

No, there is no ETA available and I would not wait for its release since it might take some more time.

KrishnaPG:
>  I know it's inconvenient that the file has a different name than upload ID but we have plans to address this. In a future release, the .info file and hook payload will include additional data specifying where the file has been stored exactly (e.g. file path, S3 bucket, object name).

Hooks are not reliable due to their asynchronous nature and timing issues. It would be great if Tus could support an RPC that allows one to query the filename given the upload id.

We tried to use the hooks to keep track of the uploads and their corresponding filename mappings, but the first blocking hook `preCreate` does not get an ID and the one that gets ID `postCreate` is an async one, which means there are race conditions with the next blocking event, which is `preFinish` for small files. As a result we sometimes loose the destination information as can be seen in the below screenshot:

image

What is preferred is, if TusD could support writing these mappings to a DB directly (in a more reliable way), or allows queries (in RPC style) to be made so that it can read the info files and give back the key information

Acconut:
An API for internal use is an interesting idea, but I don't think we will get to one in the near future.

Instead, I recommend you to make sure your hooks can properly handle the asynchronous nature of hooks (one always has to consider this when handling any kind of webhooks):
- Make sure that the `pre-finish` hook works if the `post-create` hook has not run yet (e.g. create the database rows if necessary)
- Make sure that the `post-finish` hook does not override any data from the `pre-finish` hook if it already ran
- Make sure you have atomic writes to your database, so no data gets lost (e.g. use transactions in SQL)

Using this approach we can handle any order of hook invocations :) It's a bit more involved but makes a very stable system.

KrishnaPG:
Thank you @Acconut 

> Make sure that the pre-finish hook works if the post-create hook has not run yet (e.g. create the database rows if necessary)
Make sure that the post-finish hook does not override any data from the pre-finish hook if it already ran 

That is how we are currently doing it. But for some reason it is causing us to loose the `upload.Storage` destination value (as shown in the earlier above screenshot). 

Does the `req.body.Upload..Storage` field get populated and sent in both scenarios: `pre-finish` and  `post-create` ?  Looks like one of them does not receive that field, resulting in empty destination value. Here is our code that is causing that above screenshot:

````js
function postCreateHook(req, res) {
  // create the database entry
  return createUploadDBEntry(req).then(...);
}

function createUploadDBEntry(req) {
  const upload = req.body.Upload;
  const { filename: fileName, filetype: fileType, description } = upload.MetaData;
  return knex(cUploadsTableName)
    .insert({
      id: upload.ID,
      fileName,
      destination: JSON.stringify(upload.Storage || {}),     //<-- this one here somehow missing for one of the hooks
    })
    .onConflict("id")
    .ignore(); // in case already exists, ignore this attempt
}

function preFinishHook(req, res) {
  const upload = req.body.Upload;
  // try creating the database entry if not exists yet.
  return createUploadDBEntry(req)
    .then(() => {
     // update the status of uploading
      return knex(cUploadsTableName)
        .where({ id: upload.ID, status: "Uploading" })
        .update({  status: "Uploaded"  });
    })
    .then(...);
````

Acconut:
> Does the `req.body.Upload..Storage` field get populated and sent in both scenarios: `pre-finish` and `post-create` ?

It should and if it doesn't, this is a bug. Can you confirm that it is missing in one of them by printing the value of the hook's body?

KrishnaPG:
Thank you @Acconut Tested it, and the hook body is alright. No problem there. It was some other bug on our part.

To combat the race condition (and ensure the `preFinishHook` gets called no matter what), we have an additional notification API call from the front end ( `uppy.on('upload-success')` makes a simulated call to this `preFinishHook`). It is not sending that `upload.Storage` parameter.

We had to all these jugglery because Tusd changes the uploaded filenames and there is no built-in way to reliably query the name mappings. Hence we are forced to maintain a DB and keeps the things in Sync.

TusD is a great library, if only it could address this file naming issue, or at least provide a way to query the file name mappings reliably (without the users having to maintain an external database and keep the sync in), it would be great. Just my 2 cents.

Acconut:
Thanks for the feedback, @KrishnaPG! We will take to our hearts and see how we can incorporate this into the future version. Hooks are not as simple and user friendly as we want them to be yet, so will try our best to improve.

PR #302, opened 2019-08-14
Whitelist hooks

arbakker:
Implemented cli option to for hooks whitelisting, based on this [comment](https://github.com/tus/tusd/pull/240#issuecomment-461571470). I named the cli option `hooks-enabled-events` so it is consistent with the other `hooks` options.  I added the following docs to the `docs/hooks.md` in the pull request:

<hr>
## Enabling/Disabling Hook Events

The `--hooks-enabled-events` option for the tusd binary works as a whitelist for hook events and takes a comma separated list of hook events. By default all hook events are enabled.

Possible values for the `--hooks-enabled-events` option:

- `*` - enable all hook events (default value)
- `""` - disable all hook events
- comma separated list of hook events (for instance: `pre-create,post-create`) - whitelist of hook events
<hr>

Feedback on the code quality is more than welcome, I am still quite new to Go.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,8 @@ Usage of tusd:
     	Use Google Cloud Storage with this bucket as storage backend (requires the GCS_SERVICE_ACCOUNT_FILE environment variable to be set)
   -hooks-dir string
     	Directory to search for available hooks scripts
+  -hooks-enabled-events string
+    	Comma separated list of enabled hook events, set to "-" to disable all (default "*")
   -hooks-http string
     	An HTTP endpoint to which hook events will be sent to
   -hooks-http-backoff int
--- a/cmd/tusd/cli/flags.go
+++ b/cmd/tusd/cli/flags.go
@@ -107,10 +107,18 @@ func ParseFlags() {
 }
 
 func SetEnabledHooks() {
+	var InvalidHooks []string
 	if Flags.EnabledHooks != "" {
 		slc := strings.Split(Flags.EnabledHooks, ",")
 		for i, h := range slc {
 			slc[i] = strings.TrimSpace(h)
+		}
+
+		// check if enabled hook strings are valid
+		for _, h := range slc {
+			if !hookTypeInSlice(hooks.HookType(h), hooks.AvailableHooks) {
+				InvalidHooks = append(InvalidHooks, h)
+			}
 		}
 
 		for _, h := range hooks.AvailableHooks {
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -110,6 +110,8 @@ Usage of tusd:
     	Use Google Cloud Storage with this bucket as storage backend (requires the GCS_SERVICE_ACCOUNT_FILE environment variable to be set)
   -hooks-dir string
     	Directory to search for available hooks scripts
+  -hooks-enabled-events string
+    	Comma separated list of enabled hook events, set to "-" to disable all (default "*")
```
Acconut: > set to "-" to disable all

Is that correct? Isn't "" used to disable all hooks?

cmd/tusd/cli/flags.go
```diff
@@ -89,3 +105,31 @@ func ParseFlags() {
...
 	}
 }
+
+func SetEnabledHooks() {
+	if Flags.EnabledHooks == "*" {
+		for _, h := range hooks.AvailableHooks {
+			EnabledHooks = append(EnabledHooks, h)
+		}
+	} else {
+		slc := strings.Split(Flags.EnabledHooks, ",")
```
Acconut: It would be nice to have a check here if the specified hook name actually exists, so users get notified early when they misspell `post-receive` as `post-recieve`, for example.

Acconut:
I am wondering if we should actually allow to disable hook events. If the user does not want hooks at all, they should not specify the `-hooks-path` or `-hooks-http` option, IMO. What do you think about this?

README.md
arbakker: Correct, I used `-` in a previous version and forgot to update it in the docs, I fixed that in this [commit - c93e37f](https://github.com/tus/tusd/pull/302/commits/c93e37f592fd9322924827d6eb79ef9dfe6768a8)  which I added to the pull request later.

arbakker:
Ha that did not occur to me, makes sense to not allow to disable all the hook events considering the `hook-path` and `hooks-http` options. I will change the behavior of the  `hooks-enabled-events` option to only accept a comma separated list of events. If the `hooks-enabled-events` option is missing all hook events are enabled.

arbakker:
Implemented both your suggestions:
- [commit](https://github.com/tus/tusd/pull/302/commits/75d69744af5bbd7447427482102a87fa1f73431c): removed capability to disable all events from hooks-enabled-events
- [commit](https://github.com/tus/tusd/pull/302/commits/f59eae6ac2deaa48c9ad0770a9d7e6cf2ab204e4): validate event strings in hooks-enabled-events option and show warning

Acconut:
Thank you very much! I cleaned up some code and merged this PR in https://github.com/tus/tusd/commit/f1fe5e26034504f084c97cd53cccc65271425095.

PR #315, opened 2019-10-10
disconnect context in a dedicated middleware

butonic:
Move disconnecting the context to a separate middleware. Deprecates https://github.com/tus/tusd/pull/309

This allows customizing the context that is used for the request handlers by implementing a custom middleware that can eg also add opentracing or other context information.

requires go 1.13 to do a deep copy of the request: https://golang.org/pkg/net/http/#Request.Clone see https://github.com/golang/go/issues/23544 for duscussion and https://github.com/golang/go/commit/f5c43b919499899fe006896643bbfebbea9d1995 for the commit

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -35,7 +35,9 @@ func NewHandler(config Config) (*Handler, error) {
 
 	mux := pat.New()
 
-	routedHandler.Handler = handler.Middleware(mux)
+	disconnectedHandler := handler.DisconnectMiddleware(mux)
+
+	routedHandler.Handler = handler.Middleware(disconnectedHandler)
 
 	mux.Post("", http.HandlerFunc(handler.PostFile))
 	mux.Head(":id", http.HandlerFunc(handler.HeadFile))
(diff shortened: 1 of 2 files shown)


Acconut:
To be honest, I am not a huge fan of since since it complicates the request flow even more. Instead, I could think of a creating a new background context which inherits the values from the request's context. Something like this:

```go
type contextWithValues struct {
  context.Context
  valueHolder context.Context
}

func (c contextWithValues) Value(key interface{}) interface{} {
  return c.valueHolder.Value(key)
}

ctxForStorage := contextWithValues{
  // Use background to not get cancel event
  Context: context.Background(),
  // Use request context to get stored values
  valueHolder: request.Context(),
}
```

> requires go 1.13 to do a deep copy of the request: https://golang.org/pkg/net/http/#Request.Clone see golang/go#23544 for duscussion and golang/go@f5c43b9 for the commit

This would also be a problem since we promised to support Go 1.12 until Go 1.14 has been released: https://github.com/tus/tusd/blob/master/docs/installation.md

Acconut:

PR #316, opened 2019-10-10
add gRPC hook

### LINKED ISSUE #311: add gRPC hooks
Add gRPC hooks

Just like the HTTP hooks, support gRPC+protobuf hooks

I can try to create a PR with a proposal

**Additional context**
Nowadays gRPC is a protocol widely used in cloud services

inigohu:
close #311


cmd/tusd/cli/flags.go
```diff
@@ -54,6 +55,7 @@ func ParseFlags() {
 	flag.StringVar(&Flags.HttpHooksEndpoint, "hooks-http", "", "An HTTP endpoint to which hook events will be sent to")
 	flag.IntVar(&Flags.HttpHooksRetry, "hooks-http-retry", 3, "Number of times to retry on a 500 or network timeout")
 	flag.IntVar(&Flags.HttpHooksBackoff, "hooks-http-backoff", 1, "Number of seconds to wait before retrying each retry")
+	flag.StringVar(&Flags.GrpcHooksEndpoint, "hooks-grpc", "", "An gRPC endpoint to which hook events will be sent to")
```
sagikazarmark: Shouldn't it have the same retry and backoff as HTTP?

internal/proto/v1/hook.proto
```diff
@@ -0,0 +1,66 @@
...
+// Response that contains data for sended hook
+message SendResponse {
+  string Api = 1;
+  int64 statusCode = 2;
+}
+
+// The hook service definition.
+service HookService {
+  // Sends a hook
+  rpc Send (SendRequest) returns (SendResponse) {}
```
sagikazarmark: I think it would make sense to play with the thought of making this a streaming RPC call. Given the content is actually an event stream (sort of).

Acconut:
What's the status on this PR? Is it still a work in progress?

inigohu:
> What's the status on this PR? Is it still a work in progress?

I pushed some changes following @sagikazarmark suggestions and I'm waiting for a new review.

internal/proto/v1/hook.proto
```diff
@@ -0,0 +1,66 @@
+syntax = "proto3";
```
Acconut: Can this file be located in the  cmd/tusd/cli/hooks/ folder where the other hook related functionality is located?

cmd/tusd/cli/hooks.go
```diff
@@ -58,6 +58,14 @@ func SetupPreHooks(config *handler.Config) error {
 			MaxRetries: Flags.HttpHooksRetry,
 			Backoff:    Flags.HttpHooksBackoff,
 		}
+	} else if Flags.GrpcHooksEndpoint != "" {
+		stdout.Printf("Using '%s' as the endpoint for hooks", Flags.GrpcHooksEndpoint)
```
Acconut: Please make it in the log statement clear that we are talking about gRPC hooks, e.g. "using ... as the endpoint for gRPC hooks"

cmd/tusd/cli/hooks/grpc.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	var data []byte
+	var resp *pb.SendResponse
+	ctx = stream.Context()
+	done := make(chan bool)
+
+	req := &pb.SendRequest{Hook: marshal(info)}
+	if err := stream.Send(req); err != nil {
+		log.Fatalf("can not send %v", err)
```
Acconut: Why is this error not returned but just logged?

cmd/tusd/cli/hooks/grpc.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	if err := stream.CloseSend(); err != nil {
+		log.Println(err)
```
Acconut: Why is this error not returned but just logged?

cmd/tusd/cli/hooks/grpc.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	go func() {
```
Acconut: Can you explain why goroutines are necessary here?

internal/proto/v1/hook.proto
```diff
@@ -0,0 +1,66 @@
...
+// Response that contains data for sended hook
+message SendResponse {
+  string Api = 1;
+  int64 statusCode = 2;
+}
+
+// The hook service definition.
+service HookService {
+  // Sends a hook
+  rpc Send (stream SendRequest) returns (stream SendResponse) {}
```
Acconut: What does the stream mean here? Could you explain it for a person that has never used gRPC before?

cmd/tusd/cli/hooks/grpc.go
inigohu: > When using streams, one must take care to avoid calling either SendMsg or RecvMsg multiple times against the same Stream from different goroutines. In other words, it's safe to have a goroutine calling SendMsg and another goroutine calling RecvMsg on the same stream at the same time. 
[link](https://github.com/grpc/grpc-go/blob/master/Documentation/concurrency.md#streams)

internal/proto/v1/hook.proto
inigohu: @sagikazarmark  commented:
> I think it would make sense to play with the thought of making this a streaming RPC call. Given the content is actually an event stream (sort of).

- [gRPC concepts](https://grpc.io/docs/guides/concepts/)

The most general case is Bidirectional Streaming where a single gRPC call establishes a stream in which both the client and the server can send a stream of messages to each other. 
Reviewing my approach, I think it's not well implemented for bidirectional streaming. We have two options:

1. Make a simple RPC call using the same patter as HTTP hook. Create a connection for each hook.
2. Use bidirectional streaming to use the same connection for each hook.

cmd/tusd/cli/flags.go
```diff
@@ -54,6 +57,9 @@ func ParseFlags() {
...
+	flag.IntVar(&Flags.GrpcHooksRetry, "hooks-grpc-retry", 3, "Number of times to retry on a 500 or network timeout")
```
sagikazarmark:
```suggestion
	flag.IntVar(&Flags.GrpcHooksRetry, "hooks-grpc-retry", 3, "Number of times to retry on a server error or network timeout")
```

cmd/tusd/cli/hooks/grpc.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	defer conn.Close()
+	client := pb.NewHookServiceClient(conn)
+	stream, err := client.Send(ctx)
+	if err != nil {
+		if e, ok := status.FromError(err); ok {
+			return nil, int(e.Code()), err
+		}
+		return nil, 0, err
+	}
```
sagikazarmark: I wonder if this part should go into `Setup` instead. I'm just guessing here, but this looks like a one time setup.

inigohu: Right, I moved gRPC options to Setup

Acconut:
Apologies for my late reply! I had the opportunity to get myself familiar with gRPC because of another project and I don't agree with @sagikazarmark's idea of a bidirectional stream:

> I think it would make sense to play with the thought of making this a streaming RPC call. Given the content is actually an event stream (sort of).

Yes, the hooks are a kind of event stream but they are all individual events and due to Go's nature, the ordering of these events is not guaranteed. So I do not see a benefit of using streams here. In fact, it seems like using streams makes the code more complex. We have a simple request/response structure for the HTTP hooks and it works quite well, so I am not convinced that streams are a good decision here. What do you think?

sagikazarmark:
Who said anything about *bidirectional* streaming? 🙂 

The upside of a streaming connection is that you don't have to open a new connection for every event. In case of a high throughput system this can make things faster, in case of an incident the system does not DOS itself, etc. A file upload system, with large size and small chunks could be such high throughput system.

Obviously, a streaming RPC version is more complex, but ultimately whether you want to use streaming or not depends on the context where you use tusd, and not on tusd itself. This is why I suggested to support both.

inigohu:
> We have a simple request/response structure for the HTTP hooks and it works quite well, so I am not convinced that streams are a good decision here. What do you think?

I agree. I think that for a first approach I would keep things simple and follow the same HTTP pattern with request/response structure.

Acconut:
> Who said anything about bidirectional streaming?

The PR currently implements bidirectional streaming, so that why I mentioned it. My attribution to you is wrong in that case, though :)

> The upside of a streaming connection is that you don't have to open a new connection for every event.

I would expect from gRPC to reuse the HTTP/TCP connection even in those cases. Is that not happening?

> Obviously, a streaming RPC version is more complex, but ultimately whether you want to use streaming or not depends on the context where you use tusd, and not on tusd itself. This is why I suggested to support both.

I agree with @inigohu that we should ship the simple approach first. If we see demand for a streaming endpoint in the future, we can still add it.

sagikazarmark:
> we should ship the simple approach first.

That I can agree with 🙂

inigohu:
@Acconut @sagikazarmark As commented, I've removed streaming gRPC.

cmd/tusd/cli/hooks/grpc.go
```diff
@@ -0,0 +1,76 @@
...
+	g.GrpcOpts = []grpc.DialOption{
+		grpc.WithInsecure(),
+		grpc.WithUnaryInterceptor(grpc_retry.UnaryClientInterceptor(opts...)),
+	}
+	return nil
+}
+
+func (g GrpcHook) InvokeHook(typ HookType, info handler.HookEvent, captureOutput bool) ([]byte, int, error) {
+	ctx := context.Background()
+	conn, err := grpc.Dial(g.Endpoint, g.GrpcOpts...)
```
Acconut: Is it necessary to recreate the connection and client for every hook invocation? If not, we can create them once in the Setup function and reuse them here (assuming their are safe to be used by multiple goroutines concurrently).

cmd/tusd/cli/hooks/proto/v1/hook.proto
```diff
@@ -0,0 +1,66 @@
...
+	// to be fired.
+	Upload upload = 1;
+	// HTTPRequest contains details about the HTTP request that reached
+	// tusd.
+	HTTPRequest httpRequest = 2;
+}
+
+// Request data to send hook
+message SendRequest {
+  string Api = 1;
```
Acconut: What purpose does the `string Api` property serve? I do not see it used anywhere.

cmd/tusd/cli/hooks/proto/v1/hook.proto
```diff
@@ -0,0 +1,66 @@
...
+  Hook hook = 2;
+}
+
+// Response that contains data for sended hook
+message SendResponse {
+  string Api = 1;
+  int64 statusCode = 2;
```
Acconut: HTTP and file hooks can also return data which is then sent to the client as the HTTP response body if the hook errors out. This feature is still missing here, isn't it?

cmd/tusd/cli/hooks/grpc.go
inigohu: > Is it necessary to recreate the connection and client for every hook invocation?

No, we can reuse the connection. I'll move it to Setup function.

> assuming their are safe to be used by multiple goroutines concurrently

Yes they are

cmd/tusd/cli/hooks/proto/v1/hook.proto
inigohu: It is not necessary. The idea was to add API version, but we can remove it.

cmd/tusd/cli/hooks/proto/v1/hook.proto
inigohu: What kind of response would be here?
Just an error string? Or any kind of response? If we want a general struct, we can use [google.protobuf.Any](https://developers.google.com/protocol-buffers/docs/proto3#any) type.

Acconut: The response is a byte array. Since those responses are usually rather small, I think it's fine to transfer them in one go without streaming or anything.

For reference, here is the code for getting the response body for the HTTP hooks: https://github.com/tus/tusd/blob/master/cmd/tusd/cli/hooks/http.go#L55

inigohu: > HTTP and file hooks can also return data which is then sent to the client as the HTTP response body if the hook errors out.

I don't understand why we need to return data if an error occurs. Any error message should go inside error parameter. In the HTTP hook no data is returned when the request gets an error for example:
https://github.com/tus/tusd/blob/master/cmd/tusd/cli/hooks/http.go#L51

My proposal is to leave just a response in the `SendResponse` message and get [Status](https://godoc.org/google.golang.org/grpc/status) data from the error

```
message SendResponse {
  google.protobuf.Any response = 1;
}

WDYT?
```

Acconut: What I basically want is that a hook should be able to provide an error message which is then sent to the client, if the hook fails. This is used for displaying a reason to the end user if the pre-create hook rejects an upload.

I haven't fully understood gRPC's error model, so maybe that is already possible with your PR and we don't need any modifications here.

inigohu: From tusd server we'are making a request to a hook service like this:

```
func (g GrpcHook) InvokeHook(typ HookType, info handler.HookEvent, captureOutput bool) ([]byte, int, error) {
  ...
  resp, err := g.Client.Send(ctx, req)
  ...
}
```

The hook service will have a service that implements `Send` method like this:
```
func (s *Service) Send(ctx context.Context, req *SendRequest) (*SendResponse, error) {
  ...
  return resp, error
}
```

If an error occurs, the hook service could send it back, and the same error is returned to the end user.

Acconut: Ok, then let's leave it like that.

Acconut:
Thanks for the feedback. So I think only following tasks remain:
- Remove the Api property
- Reuse the gRPC connections
- Add documentation to docs/hooks.md

Or am I missing something?

inigohu:
> Or am I missing something?

@Acconut I pushed proposed changes.

Acconut:
Amazing! Thank you very much  :100:

Issue #332, opened 2019-12-17
How to return  relative URL?


endeverchen:
**Question**
I use tusd as a lib in my server which is behind several proxies, and it's hard to configure to return absolute URL to the client. The prototol defines the URL may be absolute or relative, but I can't find a way to return relative URL.
Is there a way to return relative URL?

**Setup details**
Please provide following details, if applicable to your situation:
- Operating System: Linux
- Used tusd version: latest
- Used tusd data storage: 
- Used tusd configuration: 
- Used tus client library:

Acconut:
> Is there a way to return relative URL?

No, tusd has currently no way to return relative URLs. If that's a must for you, you would have to wrap a HTTP handler around tusd and modify the Location header to turn the absolute URL into a relative one.

endeverchen:
Thanks. I tried to wrap a HTTP handler and now it works well, but the code gets complicated. Since the protocol allows to return relative link, will it be a feature in the future?

Acconut:
We would be happy to accept a PR to support relative links but we currently do not have the resources to work on this on our own.

erlendvollset:
Need this aswell. Would be great to get this in!

Acconut:
@erlendvollset Feel free to open a PR or an issue to discuss implementation details if you are interested in helping. We are currently too low on resources to implement it on our own.

PR #337, opened 2019-12-24
feat: Enable returning relative URL to the client

endeverchen:
1. Add a parameter to indicate whether returning relative URL to the client.

### DIFF
--- a/pkg/handler/concat_test.go
+++ b/pkg/handler/concat_test.go
@@ -182,6 +182,84 @@ func TestConcat(t *testing.T) {
 			a.Equal("tada", req.Header.Get("X-Custom-Header"))
 		})
 
+		SubTest(t, "CreateWithRelativeURL", func(t *testing.T, store *MockFullDataStore, composer *StoreComposer) {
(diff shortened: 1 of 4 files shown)


endeverchen:
related issue: #332

Acconut:
Thank you for this PR! Before reviewing this properly, I do believe there is an edge case to cover: If one sends a POST request to `http://localhost/files` (no trailing slash) and the server returns `Location: vreuwghiewrhguirewh`, that would result in the upload URL being `http://localhost/vreuwghiewrhguirewh` which is not correct.

endeverchen:
Thank you for pointing out the problem. I had fixed it in the latest push: If the request URI does not end with trailing slash, the base path will be added as prefix to it.

innovate-invent:
Would you be interested in refactoring this to just swap out the callbacks at init for #341

Acconut:
Good point @innovate-invent. With #341 we might not need this explicit feature anymore as users can configure the relative URL on their own.

PR #341, opened 2020-01-03
Add callback to UnroutedHandler to remove restriction on upload URIs.

### LINKED ISSUE #340: Unrouted handler imposes unnecessary requirement on upload path
**Describe the bug**
The existence of [this function](https://github.com/tus/tusd/blob/b4db495cc65344770bcd26536022a660c3a05084/pkg/handler/unrouted_handler.go#L1193-L1200) imposes an unnecessary restriction on the URI path and software attempting to use the dependent library.

**Expected behavior**
`extractIDFromPath()` should be replaced with a callback:

```golang
type UnroutedHandler struct {
...
   GetID func(r *http.Request)(string, error)
}
```

GetID can default to the current functionality

**Setup details**
golang tusd handler library

innovate-invent:
Fixes #340

### DIFF
--- a/pkg/handler/unrouted_handler.go
+++ b/pkg/handler/unrouted_handler.go
@@ -9,6 +9,7 @@ import (
 	"math"
 	"net"
 	"net/http"
+	uri "net/url"
 	"os"
 	"regexp"
 	"strconv"
@@ -112,12 +113,10 @@ func newHookEvent(info FileInfo, r *http.Request) HookEvent {
 // such as PostFile, HeadFile, PatchFile and DelFile. In addition the GetFile method
 // is provided which is, however, not part of the specification.
 type UnroutedHandler struct {
-	config        Config
-	composer      *StoreComposer
-	isBasePathAbs bool
-	basePath      string
-	logger        *log.Logger
-	extensions    string
+	config     Config
+	composer   *StoreComposer
+	logger     *log.Logger
+	extensions string
 
 	// CompleteUploads is used to send notifications whenever an upload is
 	// completed by a user. The HookEvent will contain information about this
@@ -149,6 +148,10 @@ type UnroutedHandler struct {
 	CreatedUploads chan HookEvent
 	// Metrics provides numbers of the usage for this handler.
 	Metrics Metrics
+	// GetID is a customisable callback used by the handler to determine the id of an upload in a request
+	GetID func(r *http.Request) (string, error)
+	// GetURL is a customisable callback used by the handler to determine the url that uniquely identifies the resource id
+	GetURL func(r *http.Request, id string) (*uri.URL, error)
 }
 
 // NewUnroutedHandler creates a new handler without routing using the given
@@ -175,8 +178,6 @@ func NewUnroutedHandler(config Config) (*UnroutedHandler, error) {
 	handler := &UnroutedHandler{
 		config:            config,
 		composer:          config.StoreComposer,
-		basePath:          config.BasePath,
-		isBasePathAbs:     config.isAbs,
 		CompleteUploads:   make(chan HookEvent),
 		TerminatedUploads: make(chan HookEvent),
 		UploadProgress:    make(chan HookEvent),
@@ -184,6 +185,21 @@ func NewUnroutedHandler(config Config) (*UnroutedHandler, error) {
 		logger:            config.Logger,
 		extensions:        extensions,
 		Metrics:           newMetrics(),
+		GetID: func(r *http.Request) (string, error) {
+			return extractIDFromPath(r.URL.Path)
+		},
+		GetURL: func(r *http.Request, id string) (*uri.URL, error) {
+			if config.isAbs {
+				return uri.Parse(config.BasePath + id)
+			}
+
+			// Read origin and protocol from request
+			host, proto := getHostAndProtocol(r, config.RespectForwardedHeaders)
+
+			url := proto + "://" + host + config.BasePath + id
+
+			return uri.Parse(url)
+		},
 	}
 
 	return handler, nil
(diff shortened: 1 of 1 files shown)


innovate-invent:
I don't understand why the test is failing. The equivalence operation seems to be failing despite the reported values being equal.

Acconut:
Interesting approach, thanks for opening this PR. I understand the need for this feature but I disagree with the API design. Instead of the GetID and GetURL functions, I would advocate for this design:

```go
DeserializeUploadURL func(url string) (uploadId string, err error)
SerializeUploadURL func(uploadId string) (url string, err error)
```

In particular, I am not a fan of:
- the names GetID and GetURL, which are not very descriptive
- the http.Request argument is not need IMO (Could you explain why you added it?)
- the url.URL return value. Using a string in our situation is enough and makes it easier to handle

What do you think?

innovate-invent:
> Interesting approach, thanks for opening this PR. I understand the need for this feature but I disagree with the API design. Instead of the GetID and GetURL functions, I would advocate for this design:
> 
> ```go
> DeserializeUploadURL func(url string) (uploadId string, err error)
> SerializeUploadURL func(uploadId string) (url string, err error)
> ```
> 
> In particular, I am not a fan of:
> 
>     * the names GetID and GetURL, which are not very descriptive

I am ok with whatever name you want, although deserialize is not the best word to use given my response to the next point.

>     * the http.Request argument is not need IMO (Could you explain why you added it?)

tus does not specify how the ID is transmitted and could be influenced by extra header information (possibly added by a proxy). It is better to provide to the library user the entire request so as to not be unnecessarily restrictive.

>     * the url.URL return value. Using a string in our situation is enough and makes it easier to handle

URLs should be managed by the appropriate type, only converting to strings when strings are needed.

You will see most implementations do something along the lines of 
```golang
u := url.Parse(request.url)
paths := u.Path.split("/")
paths = append(paths, id)
u.Path = paths.join("/")
return u // Why discard information here by converting back to a string?
```

This is more for the sake of the interface, why make it restrictive?

Acconut:
Good points! The names GetID and GetUpload are too broad, so they must be changed to express their purpose more explicitly.

A bit different topic: Can you provide an example of what your URLs look like which cannot be used with the current tusd version?

innovate-invent:
> Good points! The names GetID and GetUpload are too broad, so they must be changed to express their purpose more explicitly.

How about RequestToID and IDToURL or BuildID and BuildURL or MapID and MapURL.
The argument signature is revealing to the purpose and can take some of the burden of the name.

> A bit different topic: Can you provide an example of what your URLs look like which cannot be used with the current tusd version?

My application lets the end user define the url path hierarchy so I can't provide a real example.
Here are some examples of possible hierarchy schemes:

Automatically routed based on user credential:
Request: POST /upload
Response: Location: /user/data/filename

A url scheme for concatenated uploads:
Request: POST /user/data/
Response: Location: /user/data/filename/1
Request: POST /user/data/
Response: Location: /user/data/filename/2

A 1:1 filesystem to url mapping that handles many many uploads:
Request POST /uploads/
Response: Location: /uploads/fi/le/filename

innovate-invent:
@Acconut Do you mind providing some more feedback on the proposed method names?

Acconut:
Thanks for reminding me about this PR, @innovate-invent!

>  How about RequestToID and IDToURL or BuildID and BuildURL or MapID and MapURL.
The argument signature is revealing to the purpose and can take some of the burden of the name.

I am a fan of descriptive method name and don't shy away from long names if necessary ;) What about ExtractID and GenerateUploadURL?

> My application lets the end user define the url path hierarchy so I can't provide a real example.
Here are some examples of possible hierarchy schemes:

Thanks for the example. However, in that case wouldn't it be better to add a middleware in front of the tusd handler tor rewrite URL? This has the benefit that you could possibly extract user information etc from the URL. With your proposed configurations, it would not be possible to forward this information to the end application, would it?

innovate-invent:
> Thanks for the example. However, in that case wouldn't it be better to add a middleware in front of the tusd handler tor rewrite URL? This has the benefit that you could possibly extract user information etc from the URL. With your proposed configurations, it would not be possible to forward this information to the end application, would it?

The end user provides the ExtractID and GenerateUploadURL callbacks, giving them full control over the urls. Using middleware could be a solution, although I feel it is a good idea to retain these callbacks. Hardcoded paths (or path schemes) are never a good idea, even if they are internal.

I will try and update the function names this weekend.

Acconut:
Since this PR was opened, a few things have changed which might make this PR not needed anymore:
- With tusd v2, hooks have the ability to change the upload ID.
- With https://github.com/tus/tusd/pull/1020, upload IDs will be able to contain slashes.
- In a future PR, we want to decouple the upload ID from the file path in the destination store.

Combining these changes should allow users to freely modify the upload URL however they like without the need for callback to translate between the URL and upload IDs.

ViRb3:
@Acconut with the current state of tusd, how can one implement relative upload URLs? It seems like you can change the upload ID in hooks, so one could go `../../../../../new/upload/path`, but that feels really hacky. Is there a better way? Thanks

Acconut:
@ViRb3 Can you expand an what you are trying to achieve? Do you want to have upload URLs that are not a "sub-path" of the upload creation URL? Something like this:

Upload creation url: https://example.com/api/create-upload
Upload url: https://example.com/api/upload/XXXXXX

I have never tried it, but this isn't really supported by tusd currently. If you want to do that right now, your best choice is doing the routing of the URL on your own and then calling the corresponding functions of the `UnroutedHandler` struct on your own. However, you then have to take care of rewriting the Location header in POST responses as well.

ViRb3:
@Acconut my use case is the upload URL being relative so that the tusd server does not need to know what base URL it's running behind. The client will use the relative URL with whatever base it wants. This makes it infinitely easier to deploy tusd behind reverse proxies and other more complicated topologies.

Acconut:
Responding with relative URLs for upload creation requests is currently not supported by tusd out of the box, no.

PR #342, opened 2020-01-03
Retain request context values while detaching to avoid cancellation.

innovate-invent:
Copypasta of @Acconut solution to #315

### DIFF
--- a/pkg/handler/concat_test.go
+++ b/pkg/handler/concat_test.go
@@ -38,14 +38,14 @@ func TestConcat(t *testing.T) {
 			upload := NewMockFullUpload(ctrl)
 
 			gomock.InOrder(
-				store.EXPECT().NewUpload(context.Background(), FileInfo{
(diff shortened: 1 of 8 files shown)


Acconut:
Thanks for the PR. Can you add a test to ensure that the context values from the HTTP request are actually accessible?

dschmidt:
This only makes the original `ctx` available for `DataStore`s, right? 

So there's still no way to access the context from `HookEvent`s right? Is that by design or just missing? If by design, what's the reason for that?

Acconut:
> This only makes the original ctx available for DataStores, right?

Yes, correct.

> So there's still no way to access the context from HookEvents right? Is that by design or just missing? If by design, what's the reason for that?

Yes, it's not really a design decision but grew naturally that way. The hook system was built before contexts existed in Go, so they were not included in the design of the hooks.

> So right now the only way to pass additional data to the HookEvent would probably be by messing with the headers?

You can do that. But it depends on what additional data you want to pass on and where it comes from.

Acconut:

PR #360, opened 2020-03-12
Handle FinishUpload/compose call with empty object names

tab1293:
This should fix index out of range panics we see when `FinishUpload` is called with an empty slice of object names to compose.

### DIFF
--- a/pkg/gcsstore/gcsservice.go
+++ b/pkg/gcsstore/gcsservice.go
@@ -129,6 +129,10 @@ const COMPOSE_RETRIES = 3
 
 // Compose takes a bucket name, a list of initial source names, and a destination string to compose multiple GCS objects together
 func (service *GCSService) compose(ctx context.Context, bucket string, srcs []string, dst string) error {
+	if len(srcs) < 1 {
+		return fmt.Errorf("empty srcs passed to compose for bucket: %s dest: %s", bucket, dst)
+	}
+
 	dstParams := GCSObjectParams{
 		Bucket: bucket,
 		ID:     dst,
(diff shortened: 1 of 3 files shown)


Acconut:
Thanks for this patch! Could you explain a bit more in what situation this error can occur? Does that mean that the GCSStore does not support uploading empty files?

bezi:
Our services have been seeing this exact issue crop up (basically a dupe of #419).  Preliminary testing on a fork with this fix in it does seem to be working correctly, but we'd love to get off of a custom fork of TUS.  Is there any chance this PR could get merged in?

Acconut:
Yes, this PR can be merged if it fixes the issue. I am just not sure what the actual problem is (see my question above). This PR replaces the panic with an error. That is an improvement but means that some uploads will still error out instead of completing. I would just love to know if we can improve it even further and avoid the issue at all.

bezi:
I also don't fundamentally understand what's going on here: even trying this PR was a Hail Mary attempt for us, since this bug has otherwise been rather baffling.  That said, putting into effect caused errors to basically vanish, and in particular it prevented the process from crashing out and taking down the TUS service.  Not fully sure what's going on

tab1293:
Sorry it has been awhile since I've worked with Tus. I believe this patch was in response to `FinishUpload` being called when there was actually no data uploaded to GCS. I can't quite recall why `FinishUpload` was being called [here ](https://github.com/tus/tusd/blob/master/pkg/handler/unrouted_handler.go#L699) when no chunks existed in GCS. Maybe its for the case for empty files?

Acconut (approved):

Acconut:
@tab1293 Thanks for chiming in again, greatly appreciated!

I am going to merge this then. An error is not the best behavior here, but still a lot better than a panic. @bezi, could you try to upload a 0-sized file using the gcsstore and see if it works or panics/errors out? Maybe that is the condition that causes empty object names.

PR #364, opened 2020-03-23
Make use of gRPC metadatas to pass on Hook-Name.

### LINKED ISSUE #362: gRPC Hooks headers does not contain Hook-Name
**Describe the bug**
Hook Name is not provided in the gRPC hook metadata

**To Reproduce**
1. Execute command 'tusd -hooks-grpc <grpc_service_address> -hooks-enabled-events pre-create,post-create,post-finish [...]'
2. Start upload using tus-js-client
3. gRPC call metadata does not contain the Hook-Name : 

```
Metadata(content-type=application/grpc,user-agent=grpc-go/1.29.0-dev)
``` 

**Expected behavior**
There should be a Hook-Name entry in the gRPC call metadata.

**Setup details**
- Operating System: Windows 10 pro - Docker version 19.03.5, build 633a0ea
- Used tusd version: v1.1
- Used tusd data storage: S3 - minio inside docker container
- Used tusd configuration: 'tusd -s3-bucket tusd.upload -s3-endpoint http://host.docker.internal:9000 -hooks-grpc <grpc_service_address> -hooks-enabled-events pre-create,post-create,post-finish'
- Used tus client library: tus-js-client

Clementp57:
Fixes #362

### DIFF
--- a/cmd/tusd/cli/hooks/grpc.go
+++ b/cmd/tusd/cli/hooks/grpc.go
@@ -9,7 +9,6 @@ import (
 	pb "github.com/tus/tusd/pkg/proto/v1"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/status"
-	"google.golang.org/grpc/metadata"
 )
 
 type GrpcHook struct {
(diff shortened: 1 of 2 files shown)


Acconut:
Thanks for the PR! Is there a reason why we would want to have the hook name in the metadata instead of the protobuf payload? This would avoid splitting the data over multiple places. I could imagine a hookName prototype here or so: https://github.com/tus/tusd/blob/master/cmd/tusd/cli/hooks/proto/v1/hook.proto#L43

Clementp57:
@Acconut I guess this is consistent with the HTTP implementation of hooks where the name of the hook is passed through an HTTP header. But I agree with you it would me much simpler to gather all the data at one place. So should we deviate from the HTTP implementation ?

Acconut:
Yes, that would be great!

Clementp57:
Ok I'll update the PR

PR #371, opened 2020-04-07
cli: add header forwarding in HTTP hooks

### LINKED ISSUE #369: Forward request headers in hooks
Currently request headers are included in the JSON body of a hook.
This is ok and solves most of our problems, however there are still some situations where forwarding them as actual headers is preferable.

For example, our gateway proxy injects Trace IDs as headers and all downstream infrastructure / proxies / apps / middleware / etc expects to find that value as a header.

Authorisation headers, this was touched on in #185 but doesn't appear to have been properly solved in the commit that closed that issue.
Ideally `Authorization` headers, JWT tokens in cookies etc etc should be forwarded as actual headers so that they can be evaluated in a standard manner instead of having to put custom exceptions and authentication code in just for tusd hooks.

A cli flag to specify a comma separated list of request headers to append to HTTP hooks

```
-hooks-http-forward-headers=Authorization,Cookie,TraceId
```

Forwarding all request headers is another option, downsides include
 * Backwards compatibility, suddenly hooks will have a bunch of extra headers that weren't there in previous versions. A config flag means this would be an opt-in change only.
* Blacklisting headers. Some headers should not be forwarded, `Content-Length`, `Host`, `Transfer-Encoding`, etc. Maintaining this is a chore and will likely have odd corner-cases that break.

Yep, if this is a feature that would likely be accepted I can take a look and raise a PR

hamishforbes:
Closes #369

### DIFF
--- a/cmd/tusd/cli/flags.go
+++ b/cmd/tusd/cli/flags.go
@@ -56,7 +56,7 @@ func ParseFlags() {
 	flag.StringVar(&Flags.EnabledHooksString, "hooks-enabled-events", "", "Comma separated list of enabled hook events (e.g. post-create,post-finish). Leave empty to enable all events")
 	flag.StringVar(&Flags.FileHooksDir, "hooks-dir", "", "Directory to search for available hooks scripts")
 	flag.StringVar(&Flags.HttpHooksEndpoint, "hooks-http", "", "An HTTP endpoint to which hook events will be sent to")
-	flag.StringVar(&Flags.HttpHooksForwardHeaders, "hooks-http-forward-headers", "", "List of HTTP request headers to include in hook request")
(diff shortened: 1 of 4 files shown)


cmd/tusd/cli/flags.go
```diff
@@ -55,6 +56,7 @@ func ParseFlags() {
 	flag.StringVar(&Flags.EnabledHooksString, "hooks-enabled-events", "", "Comma separated list of enabled hook events (e.g. post-create,post-finish). Leave empty to enable all events")
 	flag.StringVar(&Flags.FileHooksDir, "hooks-dir", "", "Directory to search for available hooks scripts")
 	flag.StringVar(&Flags.HttpHooksEndpoint, "hooks-http", "", "An HTTP endpoint to which hook events will be sent to")
+	flag.StringVar(&Flags.HttpHooksForwardHeaders, "hooks-http-forward-headers", "", "List of HTTP request headers to include in hook request")
```
Acconut, line 59: Please reword the description so it's apparent that the header values are forwarded from the client's requests.

PR #373, opened 2020-04-15
Rework Hook sending

### LINKED ISSUE #350: Add a synchronous CompleteUpload hook
The current `CompleteUpload` hook is asynchronous, this makes it difficult to synchronize any completion operations performed by it with the uploading client and will require an out-of-band mechanism to do so. 

Add a synchronous `CompleteUpload` hook blocking the response to the final request that complete the upload until it is done.

Possibly adding a mechanism/extension to the TUS protocol that can be used to wait for the completion of such post upload post processing as additional requests integrating it with the existing async hook.

Possibly. But the exact interface needs to be discussed.

hamishforbes:
Change the tusd standalone cli to use callback functions to issue hooks, rather than channels, for all events.
Previously only `pre-create` used a callback function which allowed it to be blocking.

Now all events trigger a callback and can be optionally configured as synchronous / blocking. (Fixes #350)

The event channels are left in place for anyone using tusd programatically but the standalone cli disables them.

### DIFF
--- a/cmd/tusd/cli/flags.go
+++ b/cmd/tusd/cli/flags.go
@@ -24,6 +24,8 @@ var Flags struct {
 	EnabledHooksString  string
 	FileHooksDir        string
 	HttpHooksEndpoint   string
+	SyncHooksString     string
+	SyncHooks           []hooks.HookType
 	HttpHooksRetry      int
 	HttpHooksBackoff    int
 	GrpcHooksEndpoint   string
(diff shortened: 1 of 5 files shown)


Acconut:
First of all, thank you for the effort you put into this. However, I am not sure if this is the best idea for a few reasons:
1. For the post-receive event is does not make much sense to execute it synchronously. Since the callback is executed in a separate goroutine, there is nothing that will wait for it. Offering a flag to make the post-receive event synchronous would then have no effect, which is not a good design IMO.
2. The actual hook code (e.g. your HTTP endpoint) does not know whether it is being executed in a synchronous or asynchronous manner. Right now, the hook code can just check the event name to see if tusd is waiting for a response but with this PR, a hook cannot make that decision anymore.
3. Making more hooks synchronous decreases performance. For example, if the post-create hook is synchronous, the data cannot be transferred until that hook is finished.
4. I always recommend people to implement the hooks in a way where they can be executed in a non-guaranteed order. It is harder to implement but makes your service more resilient against failures in my experience. By adding this feature, it would tell people that they can rely on the hooks always being delivered in the same order, which I don't think is a good system design.

Are these points understandable and make sense? In #350, I was open to adding a synchronous upload-complete event but I thought about an additional hook event and not a general switch as you propose in this PR.

hamishforbes:
1) That is absolute true, hadn't picked up on that.
2) Yep, although if you as the operator have chosen to configure the hook to be synchronous that is on you imo
3) Yes but it is a non-default option
4) That's fair.

I originally implemented this as a flag that made `post-finish` synchronous.
It then seemed a little odd to have these 2 callback functions as well as the channels, and then this odd flag to toggle synchronous behaviour on just 1 event.
It seemed nicer to make all of them configurable in a generic fashion.

I only have a requirement for a synchronous hook after the final chunk of an upload but before the response is returned to the client.

So I'm quite happy to rework this as a single new synchronous callback event.
Couple questions there though.
Is `upload-complete` what you want as the event name? It makes sense but doesn't fit the naming pattern of the existing events. Perhaps `pre-finish`?
Then it's clear that `pre-*` events are blocking and `post-*` events are not?

I assume there would not need to be a channel for this event as well?

Acconut:
> So I'm quite happy to rework this as a single new synchronous callback event.

I'm very happy to hear that :)

> Is upload-complete what you want as the event name? It makes sense but doesn't fit the naming pattern of the existing events. Perhaps pre-finish?

Exactly, `pre-finish` is also what I had in mind. This callback should be invoked after the storage is done will all the work (WriteChunk and FinishUpload) but before the response is sent to the client. One additional note: This hook should not be enabled by default or otherwise it would be breaking change. tusd has a flag which is used to specify the enabled hooks.

> I assume there would not need to be a channel for this event as well?

Exactly.

hamishforbes:
Closing in favour of #382

PR #375, opened 2020-04-17
Add missing interfaces and docs

butonic:
When trying to embed the tusd handler these interfaces are missing, AFAICT.
Also see https://github.com/cs3org/reva/pull/661/files#diff-1661b4643c6e31dcd318877ae6181470 for a working version, which is a fork of the handler I'd like to get rid of, especially seeing https://github.com/tus/tusd/pull/342

### DIFF
--- a/pkg/handler/composer.go
+++ b/pkg/handler/composer.go
@@ -6,6 +6,8 @@ package handler
 type StoreComposer struct {
 	Core DataStore
 
+	UsesCreator        bool
+	Creator            CreatorDataStore
 	UsesTerminater     bool
 	Terminater         TerminaterDataStore
 	UsesLocker         bool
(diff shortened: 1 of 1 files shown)


butonic:
ok, I'll add https://github.com/cs3org/reva/pull/661/files#diff-dede5f91913ed47007ebbf653f65f2a2 and come back ...

Acconut:
The `CreatorDataStore` inteface does not exist anymore since the NewUpload function has been merged in the main `DataStore` interface: https://github.com/tus/tusd/blob/master/pkg/handler/datastore.go#L86

Are you aware of that situation?

butonic:
No, I was not aware of that ... sorry for the noise.

PR #377, opened 2020-04-21
Send errors instead of 100 Continue...

hamishforbes:
... when requests have the `Expect: 100-continue` header.

Go's HTTP server automatically returns `100 Continue` if the Expect header is there, unless you use the `http.Error()` method.
Also handles the content length and body automagically.

### DIFF
--- a/pkg/handler/unrouted_handler.go
+++ b/pkg/handler/unrouted_handler.go
@@ -915,15 +915,8 @@ func (handler *UnroutedHandler) sendError(w http.ResponseWriter, r *http.Request
 		statusErr = NewHTTPError(err, http.StatusInternalServerError)
 	}
 
-	reason := append(statusErr.Body(), '\n')
-	if r.Method == "HEAD" {
-		reason = nil
-	}
-
 	w.Header().Set("Content-Type", "text/plain; charset=utf-8")
-	w.Header().Set("Content-Length", strconv.Itoa(len(reason)))
-	w.WriteHeader(statusErr.StatusCode())
-	w.Write(reason)
+	http.Error(w, string(statusErr.Body()), statusErr.StatusCode())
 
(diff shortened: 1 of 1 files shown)


pkg/handler/unrouted_handler.go
```diff
@@ -915,15 +915,8 @@ func (handler *UnroutedHandler) sendError(w http.ResponseWriter, r *http.Request
 		statusErr = NewHTTPError(err, http.StatusInternalServerError)
 	}
 
-	reason := append(statusErr.Body(), '\n')
-	if r.Method == "HEAD" {
-		reason = nil
```
Acconut, line 920: Do you know how `http.Error` handles HEAD requests? Does it include the body or not?

pkg/handler/unrouted_handler.go
```diff
@@ -915,15 +915,8 @@ func (handler *UnroutedHandler) sendError(w http.ResponseWriter, r *http.Request
...
-	}
-
 	w.Header().Set("Content-Type", "text/plain; charset=utf-8")
```
Acconut, line 918: The `http.Error` method automatically sets the same Content-Type header, so could remove this here. See https://golang.org/src/net/http/server.go?s=62104:62156#L2011

Acconut:
Also, please have a look at the failing tests: https://github.com/tus/tusd/runs/604523101

hamishforbes:
I actually can't reproduce the problem now, I think I must have been testing through nginx in our dev env and getting tricked by their hardcoded 100-continue support!

`http.Error()` does not handle HEAD requests so this change in it's current state is a regression.
I'll just close.

If you have a chance though, could you please take a look over my other 2 PRs? Would be great to get those upstreamed and not have to maintain a fork!

Acconut:
:+1: Thanks a lot for your efforts! I will have a look at your other PRs as soon as possible.

Issue #379, opened 2020-04-23
S3-Store: Slow transfer-speed in some environments


runar-indico:
**Question**
For a customer running Netapp StorageGrid as the s3-solution, we see slow transfer-speeds, around 2.5 Mb/sec. Same speed both for chunked uploads, as well as non-chunked uploads.

This is within the local network, so we would expect a lot faster transfers. Other applications running in the same network and using the same s3-bucket achieve the expected speeds. 

In other environments, we see great speed with tusd, so I am not really sure what the problem might be.

Do you have any tips for how we might go about debugging the slow transfer-speeds? Or any hunches about what might be the cause?

In #344 it was mentioned that there were some ideas as to how to increase the efficiency of the transfers, without going into any details.

If you are able to share some of the ideas, I might be able put in some time to implement them and contribute to this project.

**Setup details**

Please provide following details, if applicable to your situation:
- Operating System: Container-less destribution (linux), Kubernetes-environment.
- Used tusd version: 1.1.0
- Used tusd data storage: S3-compatible (specifically a [NetApp StorageGrid](https://www.netapp.com/us/products/data-management-software/object-storage-grid-sds.aspx)
- Used tusd configuration: custom-build, but here is the relevant s3-config used:
```golang
s3 := s3store.S3Store{
		Bucket:            bucket,
		Service:           service,
		ObjectPrefix:      objectPrefix,
		MaxPartSize:       5 * 1024 * 1024 * 1024,
		MinPartSize:       5 * 1024 * 1024,
		MaxMultipartParts: 10000,
		MaxObjectSize:     5 * 1024 * 1024 * 1024 * 1024,
		BaseConfig:        base,
	}
```

Acconut:
Thanks for bringing this up. Apologies for the delay as it took me some time to write this together.

> Other applications running in the same network and using the same s3-bucket achieve the expected speeds.

Just as a reference, what are the "expected speeds"?

> In other environments, we see great speed with tusd, so I am not really sure what the problem might be.

What do you mean with "other environments"? Do they use different storage backends?

> Do you have any tips for how we might go about debugging the slow transfer-speeds? Or any hunches about what might be the cause?

[...]

PR #402, opened 2020-06-22
S3Store: Concurrently write upload parts to S3 while reading from client

acj:
## Overview

As discussed in #379, the throughput performance of S3Store can degrade when the S3 (or compatible) API has high or unpredictable latency. Currently, S3Store synchronously reads bytes from the client, writes them to disk, and then writes the file on disk to the S3 API. That approach is simple and reliable, but it blocks the client from transmitting any new bytes while tusd is writing the current part to S3. This blocking causes the client to transmit more slowly due to standard TCP behavior.

This PR introduces `s3ChunkProducer`, a struct that is responsible for consuming bytes from the source reader (connected to the client) and converting them into `*os.File` instances that are pushed into a channel. The original request goroutine then consumes files from this channel until it's empty (or there's an error), uploading each file as a new part in S3.

Decoupling reading client bytes from writing them into S3 means that the two processes happen in parallel, which should allow the client to upload at a consistent speed even if the S3 API is intermittently slow. Actual performance will depend on details of the deployment, e.g. whether the tusd server is near its S3 endpoint on the network, the available bandwidth between them, how much bandwidth the client has along the path to tusd, etc. I've introduced a `MaxBufferedParts` field on S3Store to allow the server to store multiple complete parts on disk before they're written into S3. If the client is able to upload to the tusd server faster than tusd can upload to S3, then a larger buffer size may be a good way to boost client upload performance.

## Notes

Introducing parallelism in the `WriteChunk` workflow means that many of the S3Store tests run nondeterministically when uploading parts. To help with this, the first commit in this branch relaxes the use of `gomock.InOrder` so that we don't require a strict ordering of function calls -- but we still require each individual call to happen at some point. I tried to keep the strict ordering in places where it mattered. I'm not an expert on gomock, so please let me know if there's a better way to do this.

The consumer will drain the files channel and write all available parts into S3, even if the client has gone away. I think this is mostly a good thing, but it may be possible for a high-throughput client to upload several parts into the buffer and then get disconnected because of a network issue. If the server isn't using a locker and the client tries to resume uploading while tusd is still uploading those buffered parts, then we may run into a conflict. I think this is already a possibility on tusd master, but the part buffer could make such a scenario more likely.

To help tusd operators tune the size of the file channel buffer, I'd like to add a metric for how much time we spend blocking the client. Maybe this could be measured in the main WriteChunk for loop and reported at the end of an upload. I'm not sure how to pass the time data from the S3Store (which knows about part upload latency) back to the handler (which owns the Metrics struct), though. Open to suggestions.

This code has not received any production testing yet, so I would consider it unsafe until we're confident that the test coverage is sufficient and that there are no logic errors.

### DIFF
--- a/pkg/s3store/s3store.go
+++ b/pkg/s3store/s3store.go
@@ -140,6 +140,10 @@ type S3Store struct {
 	// can help improve throughput by not blocking the client while tusd is
 	// communicating with the S3 API, which can have unpredictable latency.
 	MaxBufferedParts int64
+	// TemporaryDirectory is the path where S3Store will create temporary files
+	// on disk during the upload. An empty string ("", the default value) will
+	// cause S3Store to use the operating system's default temporary directory.
+	TemporaryDirectory string
 }
 
 type S3API interface {
@@ -158,13 +162,14 @@ type S3API interface {
 // New constructs a new storage using the supplied bucket and service object.
 func New(bucket string, service S3API) S3Store {
 	return S3Store{
-		Bucket:            bucket,
-		Service:           service,
-		MaxPartSize:       5 * 1024 * 1024 * 1024,
-		MinPartSize:       5 * 1024 * 1024,
-		MaxMultipartParts: 10000,
-		MaxObjectSize:     5 * 1024 * 1024 * 1024 * 1024,
-		MaxBufferedParts:  20,
+		Bucket:             bucket,
+		Service:            service,
+		MaxPartSize:        5 * 1024 * 1024 * 1024,
+		MinPartSize:        5 * 1024 * 1024,
+		MaxMultipartParts:  10000,
+		MaxObjectSize:      5 * 1024 * 1024 * 1024 * 1024,
+		MaxBufferedParts:   20,
+		TemporaryDirectory: "",
 	}
 }
 
@@ -280,6 +285,7 @@ func (upload *s3Upload) writeInfo(ctx context.Context, info handler.FileInfo) er
 
 // s3PartProducer converts a stream of bytes from the reader into a stream of files on disk
 type s3PartProducer struct {
+	store *S3Store
 	files chan<- *os.File
 	done  chan struct{}
 	err   error
@@ -309,7 +315,7 @@ func (spp *s3PartProducer) produce(partSize int64) {
 
 func (spp *s3PartProducer) nextPart(size int64) (*os.File, error) {
 	// Create a temporary file to store the part
-	file, err := ioutil.TempFile("", "tusd-s3-tmp-")
+	file, err := ioutil.TempFile(spp.store.TemporaryDirectory, "tusd-s3-tmp-")
 	if err != nil {
 		return nil, err
 	}
@@ -334,8 +340,7 @@ func (spp *s3PartProducer) nextPart(size int64) (*os.File, error) {
 	// io.Copy returns 0 since it is unable to read any bytes. In that
 	// case, we can close the s3PartProducer.
 	if n == 0 {
-		os.Remove(file.Name())
-		file.Close()
+		cleanUpTempFile(file)
 		return nil, nil
 	}
 
@@ -378,8 +383,7 @@ func (upload s3Upload) WriteChunk(ctx context.Context, offset int64, src io.Read
 		return 0, err
 	}
 	if incompletePartFile != nil {
-		defer os.Remove(incompletePartFile.Name())
-		defer incompletePartFile.Close()
+		defer cleanUpTempFile(incompletePartFile)
 
 		if err := store.deleteIncompletePartForUpload(ctx, uploadId); err != nil {
 			return 0, err
@@ -401,6 +405,7 @@ func (upload s3Upload) WriteChunk(ctx context.Context, offset int64, src io.Read
 	}()
 
 	partProducer := s3PartProducer{
+		store: store,
 		done:  doneChan,
 		files: fileChan,
 		r:     src,
@@ -723,7 +728,7 @@ func (upload *s3Upload) concatUsingDownload(ctx context.Context, partialUploads 
 	uploadId, multipartId := splitIds(id)
 
 	// Create a temporary file for holding the concatenated data
-	file, err := ioutil.TempFile("", "tusd-s3-concat-tmp-")
+	file, err := ioutil.TempFile(store.TemporaryDirectory, "tusd-s3-concat-tmp-")
 	if err != nil {
 		return err
 	}
@@ -868,7 +873,7 @@ func (store S3Store) downloadIncompletePartForUpload(ctx context.Context, upload
 	}
 	defer incompleteUploadObject.Body.Close()
 
-	partFile, err := ioutil.TempFile("", "tusd-s3-tmp-")
+	partFile, err := ioutil.TempFile(store.TemporaryDirectory, "tusd-s3-tmp-")
 	if err != nil {
 		return nil, 0, err
 	}
--- a/pkg/s3store/s3store.go
+++ b/pkg/s3store/s3store.go
@@ -396,8 +396,7 @@ func (upload s3Upload) WriteChunk(ctx context.Context, offset int64, src io.Read
 	// we may leak file descriptors. Let's ensure that those are cleaned up.
 	defer func() {
 		for file := range fileChan {
-			os.Remove(file.Name())
-			file.Close()
+			cleanUpTempFile(file)
 		}
 	}()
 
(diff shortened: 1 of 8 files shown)


pkg/s3store/s3store.go
```diff
@@ -135,6 +135,11 @@ type S3Store struct {
 	// MaxObjectSize is the maximum size an S3 Object can have according to S3
 	// API specifications. See link above.
 	MaxObjectSize int64
+	// MaxBufferedParts is the number of additional parts that can be received from
+	// the client and stored on disk while a part is being uploaded to S3. This
+	// can help improve throughput by not blocking the client while tusd is
+	// communicating with the S3 API, which can have unpredictable latency.
+	MaxBufferedParts int
```
Acconut, line 142: The addition of MaxBufferedParts is a very good one! Only to things: Could we make it an int64, just to align with the rest of the configuration files? Also, can we set a default value for this (maybe 20?) in the New function (just as a safety net)?

pkg/s3store/s3store.go
```diff
@@ -272,6 +277,70 @@ func (upload *s3Upload) writeInfo(ctx context.Context, info handler.FileInfo) er
 	return err
 }
 
+type s3ChunkProducer struct {
```
Acconut, line 280: Can you add a brief sentence here documenting what this structure does? Also, what about naming it s3PartProducer to align with the naming of MaxBufferedParts?

pkg/s3store/s3store.go
```diff
@@ -272,6 +277,70 @@ func (upload *s3Upload) writeInfo(ctx context.Context, info handler.FileInfo) er
...
+	// prevent the remaining chunk to be stored on S3.
+	if err == io.ErrUnexpectedEOF {
+		err = nil
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	// io.Copy returns (0, nil) when it reaches EOF
```
Acconut, line 331: I am aware that this comment has already been there before your PR, but I think it is wrong. io.Copy returns (0, nil) if it is reaches EOF *and* did not read an bytes (i.e. the input buffer is empty).

I think in that case, a better comment would be:
> If the entire request body is read and no more data is available, io.Copy returns 0 since it is unable to read any bytes. In that case, we can close the s3PartProducer.

What do you think?

pkg/s3store/s3store.go
```diff
@@ -373,7 +430,16 @@ func (upload s3Upload) WriteChunk(ctx context.Context, offset int64, src io.Read
 		offset += n
 		bytesUploaded += n
 		nextPartNum += 1
+
+		if err := os.Remove(file.Name()); err != nil {
```
Acconut, line 434: If the function returns before os.Remove or file.Close is invoked, we leak file descriptions and temporary files. We also cannot use defer here, since we want to remove files as soon as they are not needed anymore. Do you have an idea how to improve this? Maybe refactor the uploading part into it's own function and use defer in there?

pkg/s3store/s3store_chunk_producer_test.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	safelyDrainChannelOrFail(fileChan, t)
+
+	if cp.err == nil {
+		t.Error("expected an error but didn't get one")
+	}
+}
+
+func safelyDrainChannelOrFail(c chan *os.File, t *testing.T) {
```
Acconut, line 144: Very nice tests! Do we also need a "safety drain" for the producer in the WriteChunk method on its own? I.e. when an error appears and WriteChunk has to return, we still make sure to drain the file channel and close all file descriptors.

pkg/s3store/s3store.go
acj: > Could we make it an int64

👍

> can we set a default value for this (maybe 20?) in the New function (just as a safety net)?

Sure. I had left it at zero (unset) because it was closest to the existing behavior, but maybe it's better to take advantage of the buffering by default.

pkg/s3store/s3store.go
acj: Ah, yes, I meant to ask for feedback about this specifically because it's not safe as-is for the reason you mentioned. I think your refactor suggestion is a good one. I'll give that a try and report back. If we're not happy with it, I'm fine with reverting to the current `defer` approach for now.

pkg/s3store/s3store_chunk_producer_test.go
acj: Yes, I think so. I'll look into this and will try to write a test for it as well.

acj:
> We had thoughts about storage-specific metrics (e.g. to measure latency to S3 or other services) but this would require some more work on how we gather metrics. Maybe that is better reserved for its own PR. What do you think?

👍

Acconut:
Thanks for your feedback! Looking forward to your adjustments :)

acj:
Thanks for your patience. I'm still thinking about how/whether to test the file draining in d014a3ef91193d21d1e698c89f0f0c5dbb677cb9, but this felt like a good time to stop and gather more feedback. I should have time to do a bit of stress testing later this week.

pkg/s3store/s3store.go
```diff
@@ -388,12 +392,21 @@ func (upload s3Upload) WriteChunk(ctx context.Context, offset int64, src io.Read
 	doneChan := make(chan struct{})
 	defer close(doneChan)
 
-	chunkProducer := s3ChunkProducer{
+	// If we panic or return while there are still files in the channel, then
+	// we may leak file descriptors. Let's ensure that those are cleaned up.
+	defer func() {
+		for file := range fileChan {
+			os.Remove(file.Name())
+			file.Close()
```
Acconut, line 400: Closing the file descriptor after removing the file should be fine on Linux. But, IIRC, on Windows you need to close all file descriptors before you can remove the file. So please change the order of these two lines, unless I am missing something.

FYI, deferred functions call are execute in the opposite order they are declared: https://tour.golang.org/flowcontrol/13. That's why we call `defer os.Remove` before `defer file.Close`

acj: Right, good catch. In each place where we do this file cleanup, I was thinking about using this same closure pattern instead of using two separate defers. Something like:

```go
defer func() {
  file.Close()
  os.Remove(file)
}()
```

Then we don't need to remember defer's stack behavior when we read this code. WDYT?

Acconut: Good idea. Maybe we can also refactor the file.Close and os.Remove into its own function.

Acconut:
@acj If you want to, I can also take care of the last minor tasks to free you up and get this PR finished a bit sooner :)

pkg/s3store/s3store.go
```diff
@@ -724,9 +727,7 @@ func (upload *s3Upload) concatUsingDownload(ctx context.Context, partialUploads
 	if err != nil {
 		return err
 	}
-	fmt.Println(file.Name())
```
acj, line 650: This looked like debugging/testing residue, so I removed it. Let me know if that's not the case.

acj:
> If you want to, I can also take care of the last minor tasks

Yes, much appreciated! My past few weeks have been unexpectedly busy. It sounds like you have a clear idea of what you want for that last test, so I'll leave it in your able hands.

I will be doing more stress testing and will post here if I find any issues.

pkg/s3store/s3store.go
Acconut: Yes, you are right! Thanks!

Acconut:
> Yes, much appreciated! My past few weeks have been unexpectedly busy. It sounds like you have a clear idea of what you want for that last test, so I'll leave it in your able hands.

No worries, please make sure to not overwork yourself :) I will have a look at the tests.

> I will be doing more stress testing and will post here if I find any issues.

I also did some testing and your PR improve the upload speed from 90Mbit/s to 150Mbit/s! With a bigger part size (50MB instead of 5MB), I even got up to 250MBit/s. So that's likely the next step in terms of optimizations. Very impressive work here!

Acconut:
@acj I just pushed the test. Please have a look at it and let me know what you think!

pkg/s3store/s3store_test.go
```diff
@@ -1317,3 +1319,99 @@ func TestConcatUploadsUsingDownload(t *testing.T) {
...
+	}).Return(&s3.ListPartsOutput{
+		Parts: []*s3.Part{
+			{
+				Size: aws.Int64(100),
+			},
+			{
+				Size: aws.Int64(200),
+			},
+		},
+	}, nil).Times(2)
```
acj, line 1394: > .Times(2)

👍

It's out of scope here, but I think there are a number of places where we could combine duplicate `EXPECT()` calls now that the call ordering is less strict.

pkg/s3store/s3store.go
```diff
@@ -334,8 +340,7 @@ func (spp *s3PartProducer) nextPart(size int64) (*os.File, error) {
 	// io.Copy returns 0 since it is unable to read any bytes. In that
 	// case, we can close the s3PartProducer.
 	if n == 0 {
-		os.Remove(file.Name())
```
acj: Thanks for catching these 👍

PR #411, opened 2020-07-28
Enable S3 transfer acceleration

netaisllc:
This PR address the following issue: https://github.com/tus/tusd/issues/392

But, I'm not sure this is the best way to go about the feature.  Please advise if this should reside behind a feature flag or other ideas.  Thanks.

### DIFF
--- a/cmd/tusd/cli/composer.go
+++ b/cmd/tusd/cli/composer.go
@@ -26,7 +26,7 @@ func CreateComposer() {
 	if Flags.S3Bucket != "" {
 		s3Config := aws.NewConfig()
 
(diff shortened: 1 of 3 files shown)


Acconut:
Thank you very much for this PR! 

> Please advise if this should reside behind a feature flag or other ideas. 

Yes, I think transfer acceleration should only be enabled when a feature flag is set. Otherwise an error will occur for users if the bucket has not transfer acceleration enabled or maybe even if S3-compatible storage other than AWS S3 is used.

netaisllc:
Please advise any more adjustments.

netaisllc:
Issue #392

Issue #395, opened 2020-06-12
Gracefully terminate on SIGTERM


jrtcppv:
**Is your feature request related to a problem? Please describe.**
We are currently running tusd in Kubernetes. During normal operation (upgrades, autoscaling, etc), Kubernetes shuts down tusd, first by sending a SIGTERM, waiting for a grace period, then sending a SIGKILL. We ran into issues because tusd uses the default behavior for Go programs, which is to immediately ungracefully terminate upon receiving a SIGTERM. Occasionally this causes files to become locked, and tusd returns consistent 423 responses until the upload is restarted (we are using the python client in this case).

**Describe the solution you'd like**
It would be great if tusd could gracefully shutdown upon receiving a SIGTERM, ie flush out any pending requests then close.

**Describe alternatives you've considered**
We have a fork of tusd at https://github.com/cvisionai/tusd in which SIGTERM is simply ignored. This resolves our issue, but the correct behavior is to gracefully terminate.

**Can you provide help with implementing this feature?**
Yes we can if this is a desired feature in tusd.

Acconut:
You are absolutely right, tusd should be able to gracefully shut down. We would be very happy to assist you with a PR!

rbastic:
Hi,

I have a very rough cut of a fix for this issue here: https://github.com/rbastic/tusd/commit/90a8d992ca6701a0bea0a1d81acdaa5b02f3f1e0

Would you care to "pre-review" this and give me your feedback?

Thanks!
Ryan

mehtadev17:
+1

Acconut:
@rbastic Would you mind opening a PR with your change? Then we can talk about your commit and provide easy advice when needed :)

rbastic:
Hey! As requested, here is the PR: https://github.com/tus/tusd/pull/415 Thanks!

Acconut:
#963 has been merged, so this will be included in the upcoming v2 release!

PR #415, opened 2020-08-20
graceful shutdown - preliminary attempt 

rbastic:
As requested, here is my PR to resolve https://github.com/tus/tusd/issues/395

I tested this last night and the behavior is as follows:
- when uploading, the server waits until the active handler being executed is complete (so a chunk can be written) before shutting down
- if tusd is then started back up (and I resume the upload) it picks back up where it left off

In my testing, I killed tusd with an active upload at 34%, started tusd back up, and it picked right back up at 35% like nothing had happened. Then it completed successfully. This was just a small test but exactly the behavior desired/expected, I think? 

Main questions: what do you think about the TODO markers? The shutdown timeout, etc.? I'm guessing these should become CLI flags.

### DIFF
--- a/cmd/tusd/cli/serve.go
+++ b/cmd/tusd/cli/serve.go
@@ -1,15 +1,23 @@
 package cli
 
 import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
+
 	"net"
 	"net/http"
 	"strings"
 	"time"
 
+	"github.com/oklog/run"
 	"github.com/tus/tusd/pkg/handler"
 )
 
-// Setups the different components, starts a Listener and give it to
+// Serve sets up the different components, starts a Listener and give it to
 // http.Serve().
 //
 // By default it will bind to the specified host/port, unless a UNIX socket is
@@ -51,8 +59,20 @@ func Serve() {
 
 	SetupPostHooks(handler)
 
+	mux := http.NewServeMux()
+	s := &http.Server{
+		Addr:		address,
+		Handler:        mux,
+		// TODO(rbastic): what to do about these?
+		// ReadTimeout:    5000 * time.Second,
+		// WriteTimeout:   5000 * time.Second,
+		// MaxHeaderBytes: 1 << 20,
+		// END TODO(rbastic)
+	}
+
+
 	if Flags.ExposeMetrics {
-		SetupMetrics(handler)
+		SetupMetrics(mux, handler)
 		SetupHookMetrics()
 	}
 
@@ -61,18 +81,19 @@ func Serve() {
 	if basepath == "/" {
 		// If the basepath is set to the root path, only install the tusd handler
 		// and do not show a greeting.
-		http.Handle("/", http.StripPrefix("/", handler))
+
+		mux.Handle("/", http.StripPrefix("/", handler))
 	} else {
 		// If a custom basepath is defined, we show a greeting at the root path...
-		http.HandleFunc("/", DisplayGreeting)
+		mux.HandleFunc("/", DisplayGreeting)
 
 		// ... and register a route with and without the trailing slash, so we can
 		// handle uploads for /files/ and /files, for example.
 		basepathWithoutSlash := strings.TrimSuffix(basepath, "/")
 		basepathWithSlash := basepathWithoutSlash + "/"
 
-		http.Handle(basepathWithSlash, http.StripPrefix(basepathWithSlash, handler))
-		http.Handle(basepathWithoutSlash, http.StripPrefix(basepathWithoutSlash, handler))
+		mux.Handle(basepathWithSlash, http.StripPrefix(basepathWithSlash, handler))
+		mux.Handle(basepathWithoutSlash, http.StripPrefix(basepathWithoutSlash, handler))
 	}
 
 	var listener net.Listener
@@ -92,7 +113,42 @@ func Serve() {
 		stdout.Printf("You can now upload files to: http://%s%s", address, basepath)
 	}
 
-	if err = http.Serve(listener, nil); err != nil {
-		stderr.Fatalf("Unable to serve: %s", err)
+	var g run.Group
+
+	g.Add(func() error {
+		if err = s.Serve(listener); err != nil {
+			stderr.Fatalf("Unable to serve: %s", err)
+			return err
+		}
+		return nil
+	}, func(error) {
+		// TODO(rbastic): externalize shutdown timeout? for now just 30 mins? i don't know.
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*600)
+		defer cancel()
+		stderr.Fatalf("httpserver shutting down %s", s.Shutdown(ctx))
+	})
+
+	cancel := make(chan struct{})
+	g.Add(func() error {
+		return interrupt(cancel)
+	}, func(error) {
+		close(cancel)
+	})
+
+	err = g.Run()
+	if err != nil {
+		stderr.Fatalf("error", err)
+		os.Exit(-1)
+	}
+}
+
+func interrupt(cancel <-chan struct{}) error {
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
+	select {
+	case sig := <-c:
+		return fmt.Errorf("received signal %s", sig)
+	case <-cancel:
+		return errors.New("canceled")
 	}
 }
--- a/cmd/tusd/cli/serve.go
+++ b/cmd/tusd/cli/serve.go
@@ -117,7 +117,6 @@ func Serve() {
 		}
 		return nil
 	}, func(error) {
-		// TODO(rbastic): externalize shutdown timeout? for now just 30 mins? i don't know.
 		ctx, cancel := context.WithTimeout(context.Background(), time.Second*time.Duration(Flags.ShutdownTimeout))
 		defer cancel()
 		stderr.Printf("httpserver shutting down %s\n", s.Shutdown(ctx))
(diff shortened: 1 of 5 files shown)


cmd/tusd/cli/serve.go
```diff
@@ -51,8 +59,20 @@ func Serve() {
 
 	SetupPostHooks(handler)
 
+	mux := http.NewServeMux()
+	s := &http.Server{
+		Addr:		address,
+		Handler:        mux,
+		// TODO(rbastic): what to do about these?
```
Acconut, line 66: We have a custom timeout implementation because Go's timeout operate on the entire HTTP request and not a single read operation. You can find it here: https://github.com/tus/tusd/blob/master/cmd/tusd/cli/listener.go So I am not sure if we need the timeouts here TBH.

cmd/tusd/cli/serve.go
```diff
@@ -92,7 +113,42 @@ func Serve() {
 		stdout.Printf("You can now upload files to: http://%s%s", address, basepath)
 	}
 
-	if err = http.Serve(listener, nil); err != nil {
-		stderr.Fatalf("Unable to serve: %s", err)
+	var g run.Group
```
Acconut, line 116: Can you explain what the run.Group is? Do we need it?

cmd/tusd/cli/serve.go
```diff
@@ -92,7 +113,42 @@ func Serve() {
...
+
+	g.Add(func() error {
+		if err = s.Serve(listener); err != nil {
```
Acconut, line 119: The Golang documentation says following:

> When Shutdown is called, Serve, ListenAndServe, and ListenAndServeTLS immediately return ErrServerClosed. Make sure the program doesn't exit and waits instead for Shutdown to return. 

So, I think we need a waiting mechanism here, don't we?

cmd/tusd/cli/serve.go
```diff
@@ -92,7 +113,42 @@ func Serve() {
...
+			stderr.Fatalf("Unable to serve: %s", err)
+			return err
+		}
+		return nil
+	}, func(error) {
+		// TODO(rbastic): externalize shutdown timeout? for now just 30 mins? i don't know.
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*600)
+		defer cancel()
+		stderr.Fatalf("httpserver shutting down %s", s.Shutdown(ctx))
```
Acconut, line 128: Using Fatalf seems a bit harsh here since it forces a process to exit using os.Exit if I recall correctly. Can we use less aggressive logging statement here?

rbastic: Small followup: changed to stderr.Printf and chose 'shutdown-timeout' for the CLI parameter name.

cmd/tusd/cli/serve.go
Acconut: > I interpret that as avoid calling os.Exit() and just call Shutdown() instead

I don't think that is right. We must make sure that the main goroutine does not end before the call to Shutdown() has completed. I am not sure if run.Group would guarantee that.

cmd/tusd/cli/serve.go
Acconut: To be honest, the run.Group usage makes the code a lot harder for me and other contributors to understand. This package might be very helpful for larger application with many moving parts but I don't think it's necessary in this case. There are some very simple and easy to understand examples about stopping a HTTP server from a signal. For example, https://stackoverflow.com/a/40041517. What do you think about switching to that one?

Acconut:
I left two comments on your changes. Please check them out :)

On another topic: How does this behave with long running requests (e.g. a PATCH request transferring multiple GBs for hours)? Is this request interrupted when the shutdown begins or only after the timeout has been reached? Is the request interrupted "nicely" by closing the request body or violently by killing the TCP connection? Did you try that out and can shine light on it?

cmd/tusd/cli/serve.go
rbastic: My understanding is that the call to Shutdown aborts if it does not complete within the deadline associated from the supplied context.

The context deadline should likely be configured in accordance with the interval for an individual chunk request, so that restarting the server may happen with seemingly minimal interruption.

rbastic:
> I left two comments on your changes. Please check them out :)
> 
> On another topic: How does this behave with long running requests (e.g. a PATCH request transferring multiple GBs for hours)? Is this request interrupted when the shutdown begins or only after the timeout has been reached? Is the request interrupted "nicely" by closing the request body or violently by killing the TCP connection? Did you try that out and can shine light on it?

The only test I performed was the originally specified one in the PR description. The data set size is sub-gigabyte and performed locally.

The request is never interrupted. Shutdown fails if in-flight requests are not completed prior to the shutdown's context timeout (see my other comments.)

Acconut:
> The request is never interrupted. Shutdown fails if in-flight requests are not completed prior to the shutdown's context timeout

Does that mean that the server cannot shutdown while requests are still open to it? I think that this would be a good decision. I would expect tusd, once asked to shutdown, to wait a few seconds for all requests to end while not accepting new requests. If the shutdown timeout is reached it should end all still active requests by force. Is this behavior implemented by your PR or am I missing something?

rbastic:
> > The request is never interrupted. Shutdown fails if in-flight requests are not completed prior to the shutdown's context timeout
> 
> Does that mean that the server cannot shutdown while requests are still open to it? I think that this would be a good decision. I would expect tusd, once asked to shutdown, to wait a few seconds for all requests to end while not accepting new requests. If the shutdown timeout is reached it should end all still active requests by force. Is this behavior implemented by your PR or am I missing something?

This is not how Go's built-in Shutdown API works. Please see the documentation for the best clarification on this matter. Active requests will not be ended by force if the timeout is reached - that would defeat the purpose of graceful shutdown.

cmd/tusd/cli/serve.go
rbastic: I'm not really interested in rewriting the patch. Feel free to rewrite it if you understand how to do so.

However, is there something in particular that you and the other contributors you've discussed this with don't understand about oklog/run groups? Peter Bourgon (the oklog/run author) is also accessible on the gophers slack and may be able to clarify regarding any specific questions you have.

Acconut:
> Active requests will not be ended by force if the timeout is reached - that would defeat the purpose of graceful shutdown.

Ok, that's good to know. However, it's not optimal if an active request (which might be open for hours) prevents tusd from shutting down. Does you PR already handle that in some case? The usage of run.Group makes those question very hard for me to answer by looking at the code.

rbastic:
> > Active requests will not be ended by force if the timeout is reached - that would defeat the purpose of graceful shutdown.
> 
> Ok, that's good to know. However, it's not optimal if an active request (which might be open for hours) prevents tusd from shutting down. Does you PR already handle that in some case? The usage of run.Group makes those question very hard for me to answer by looking at the code.

Remember that part where I was setting ReadTimeout and WriteTimeout and you mentioned you already have timeout handlers in place? What do those timeout handlers do and/or can you point out where they are in the code?

Thanks,
Ryan

rbastic:
Closed by accident. No coffee yet this morning, hahaha.

rbastic:
Hi again,

Just following up. I see ReadTimeout and WriteTimeout in the net.Listener code that tusd uses. This answers your last question, right? There's nothing to do on my end here?

Acconut:
> Remember that part where I was setting ReadTimeout and WriteTimeout and you mentioned you already have timeout handlers in place? What do those timeout handlers do and/or can you point out where they are in the code?

Our timeout handling is located here: https://github.com/tus/tusd/blob/master/cmd/tusd/cli/listener.go Its purpose is to reset the read/write timeouts whenever a read/write operation occurs. Go does not support that kind of timeouts on its own, so we had to build it for ourselves. Let me know if you need more in-depth explanation.

> This answers your last question, right? There's nothing to do on my end here?

To be honest, I am still very much confused on how this patch behaves when requests are still open. Even with your explanations I am not able to understand it from just looking at the code. I would have to try it on my own and see how it works.

Also, there is still the usage of `run.Group`. You said that you are not interested in removing it (https://github.com/tus/tusd/pull/415#discussion_r485020714) which is absolutely in your rights. It will take some time until I can tackle this on my own.

All in all, this PR is not in a state where I can confidently merge it into master since it has a lot of potential for breakage. I'm sorry that all your efforts lead to such a situation.

rbastic:
That's fair. Let me see if I can come up with something else on my end to try to address some of the issues remaining. If I can come up with a solid explanation for how the run.Group functionality works, would you be comfortable with merging based on that? Or are there other blockers?

I think there's a reason why rewriting it away from oklog/run may be less than beneficial, but my research will hopefully figure that out. I don't want to be attached to it unnecessarily either if it's introducing additional complexity. Time is tight on my end though, hence why I'm hoping to avoid rewriting the patch if I can.

Acconut:
> That's fair. Let me see if I can come up with something else on my end to try to address some of the issues remaining.

That's great, thank you for understanding our point of view as well :)

> If I can come up with a solid explanation for how the run.Group functionality works, would you be comfortable with merging based on that?

The chances are low, I would say. The best code is still the one which does not need explanation for a reasonably experienced Go developer. If it's obvious that the code without run.Group is more complex, I am happy to accept it as a dependency. But unless that has been shown, I would want to avoid it.

> I think there's a reason why rewriting it away from oklog/run may be less than beneficial, but my research will hopefully figure that out. 

If there is such a reason, I am open to accept run.Group :)

ssikyou:
Any update on this feature?

Acconut:
> Any update on this feature?

Sadly no. This PR is still missing a feature where the currently open upload requests get gracefully closed down (in addition to the already implemented stop of accepting new upload requests). Also, I disagree with the author about the use of a dependency and they clearly indicated that they are not interested in updating the patch (which they completely have the right to).

However, maybe in the near future I will attempt to implement this on my own, as gracefully closing open requests is a somewhat harry thing to implement without more knowledge about the internals of tusd.

PR #425, opened 2020-09-11
Heroku deployment

ifedapoolarewaju:
This PR enables auto deployment of tusd to heroku

wasn't sure what part of the docker/kube build should be removed, so I just left them be for now.


kvz (approved):
Looking good, but we should be cleaning up all the k8s traces in this PR too i feel. Publishing the docker container can stay. Perhaps @kiloreux can take a swing at that

kvz:
okay!

PR #426, opened 2020-09-11
build: remove k8s related build

ifedapoolarewaju:
deployment worked so ...


kiloreux (approved):
Looking good Sir. Thank you very much.

PR #430, opened 2020-10-01
Mention https explicitly in http hooks page.

RubenGarcia:


RubenGarcia:
Mentioned in https://github.com/tus/tusd/pull/423
It is not obvious that https hooks are supported.

Acconut (approved):
Thank you very much!

PR #433, opened 2020-10-11
Fix unrouted handler forwarded header

zue666:
should use `;` instead of `,` in https://github.com/tus/tusd/blob/f863189009bb9e8a279c7d2a11f3b7503e3a9c7e/pkg/handler/unrouted_handler.go#L24
which causes the returned `host` (when `handler.config.RespectForwardedHeaders` is set to `true`) https://github.com/tus/tusd/blob/f863189009bb9e8a279c7d2a11f3b7503e3a9c7e/pkg/handler/unrouted_handler.go#L1032 to be `upload.example.tld;proto=https;proto-version=` instead of `upload.example.tld`.


Acconut:
Thank you very much! Could you explain what those tests are supposed to do? Did you intentionally commit these since they contain the erroneous regular expression?

zue666:
Well, yes. I intentionally committed these tests. It test the original implementation and the changes. and since the `getHostAndProtocol` function is not exported, I copied it, as well as the regex.

Acconut:
I think it's enough to add a test case which ensure that the patched function now parses the headers correct. There should be no need to keep incorrect functions around.

zue666:
Ok, I'll update the tests, and recommit the changes.

Acconut (approved):
Thank you. I merged your test with an existing one, so we don't have to duplicate code

PR #436, opened 2020-10-17
gcsstore: Perform multiple API calls for compose and delete operations concurrently

justinruggles:
GCS only allows composing up to 32 objects at once with a single API call. However, with small chunks, the final compose required at the end of uploading can have hundreds of objects. Performing all of the required API calls sequentially is unnecessarily slow.

Additionally, the temporary objects are deleted using a single API call per object. All of these deletes can be performed concurrently to improve performance.

An optional parameter, MaxConcurrency, is added to GCSService to control how much concurrency is used per operation. The default is 256.


justinruggles:
Sorry, wrong repo. I will send a cleaned-up version after merging it into a fork.

Issue #443, opened 2020-12-15
Reusing provided HTTP context in UnroutedHandler


bastien-seqone:
**Is your feature request related to a problem? Please describe.**
We use tusd with our own store to use the SSE-C functionality of the S3 standard and manage encryption. In order to manage our encryption keys, we have middleware that does a lot of business verification and we would like to pass information through the HTTP context.

But in the UnroutedHandler, the handler methods create a new context (ctx.Background()) and do not reuse the HTTP request context provided (r.Context()).

By example, 
```golang
// PostFile creates a new file upload using the datastore after validating the
// length and parsing the metadata.
func (handler *UnroutedHandler) PostFile(w http.ResponseWriter, r *http.Request) {
	ctx := context.Background() // <- Here
...
```

**Describe the solution you'd like**
The [Http request context](https://golang.org/pkg/net/http/#Request.Context) could be used: this would allow middleware to add their own information. I don't believe there is any reason not to use it, but I may have missed something.

Using the same example, it would be : 
```golang
// PostFile creates a new file upload using the datastore after validating the
// length and parsing the metadata.
func (handler *UnroutedHandler) PostFile(w http.ResponseWriter, r *http.Request) {
	ctx := r.Context()
...
```

**Describe alternatives you've considered**
Using wide available Hashmap, or something like that… but it's not very nice :(

**Can you provide help with implementing this feature?**
Yep, I think we can manage this little change :)

Acconut:
Thank you for the detailed report. There were already some attempts to reuse the request context but we encountered some issues which are explained in https://github.com/tus/tusd/pull/361 and https://github.com/tus/tusd/pull/309. I am not sure if these would be resolved by now but I am more than happy to help you try it out! :)

bastien-seqone:
Sorry I missed the attempts already made.

I've read the solution you propose in the comments in PR https://github.com/tus/tusd/pull/315. I will try to propose something in this way.

bastien-seqone:
I have just seen that this that this PR https://github.com/tus/tusd/pull/342 seems to be doing the job.

> Can you add a test to ensure that the context values from the HTTP request are actually accessible?

It is this missing test that is blocking its approval?

Acconut:
Yes, this PR is just missing a proper test. However, I would also like to try reusing the request context in such a way that the data storages still work. But that may require some internal refactoring. This would definitely be a better approach instead of such a work around, if you understand what I mean.

Acconut:
[...]

PR #455, opened 2021-01-19
Retains HTTP Context values

bastien-seqone:
As discussed in #443, revival of the pull request #342 , to which I have added a unit test as requested in the comments.

The unit test ensure that value in request context is accessible in datastore. The original PR is a Copypasta of @Acconut solution to #315 by @innovate-invent

I rebase previous PR onto master.

It could solve issues #443, and related to #361 and #309.

### DIFF
--- a/pkg/handler/concat_test.go
+++ b/pkg/handler/concat_test.go
@@ -38,14 +38,14 @@ func TestConcat(t *testing.T) {
 			upload := NewMockFullUpload(ctrl)
 
 			gomock.InOrder(
-				store.EXPECT().NewUpload(context.Background(), FileInfo{
+				store.EXPECT().NewUpload(WrapsContext(context.Background()), FileInfo{
 					Size:           300,
 					IsPartial:      true,
 					IsFinal:        false,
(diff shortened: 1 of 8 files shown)


svett:
Any plans this PR to be accepted?

pcfreak30:
Ive just ran into this issue as well. Im delegating to tusd `Handler` from the iris framework which has to do auth via a jwt middleware, and need to ensure the tus store is aware of the auth state data via context to save to the database properly.

Given the current situation, will likely have to fork.

Acconut:
I somehow forgot this PR, but I will revisit it soon. Thanks for reminding me.

NickNeedsAName:
ping

This seems to be the "go" way to do things.  Is there a reason to not use the request context when doing things?

Acconut:
> Is there a reason to not use the request context when doing things?

The request context is cancelled if the request is aborted. But we want to continue performing tasks after the request is aborted, such as saving the last received bytes to S3. This cannot be done if we use the request context.

Issue #401, opened 2020-06-20
Azure Blob Storage


Benehiko:
Hi there,

**Is your feature request related to a problem? Please describe.**
I saw the library does not have support for the Azure Blob Storage as a cloud store. I had implemented tusd in a freelancing project recently and was forced by my employer to use Azure which I in the end coded on top of the tusd library as a custom store which worked out quite well.

**Describe the solution you'd like**
Have Azure Blob Storage support out of the box.

**Can you provide help with implementing this feature?**
I thought since this isn't currently in the tusd library that I'd integrate the code I wrote and do a pull request. The code might not be up to the library's standard and there are some things I would test before adding it - I just thought I'd get some feedback first.

What are your thoughts?

Kind Regards
Benehiko

wregen:
I would love to see  Azure Blob Storage supported by tus.io out of the box.

Acconut:
Yes, we would be very much interested in such a PR! Thanks for the offer :heart:

Benehiko:
I will see if I can have a PR by this weekend :)

AKlaus:
Can we extend the feature request by adding:

1. Support _Azure File Storage_ along with _Blob Storage_.
2. Compute the MD5 hash of the file and put it base64 representation in the `Content-MD5` property of the blob.

It seems that the linked PR doesn't have it if I'm not mistaken (I'm terrible in Go).

omBratteng:
@AKlaus may I ask what benefit computing the MD5 hash of the file has? (I've never used it, so I am curious)
Would you compute the MD5 hash of the block of the blob, or the entire file?

AKlaus:
The hash is used to verify the integrity of the blob/file during transport.
Form experience, the likelihood of broken transmission is 1/1,000,000 and may appear negligible, but my users transfer many million files.

Acconut:
I am going to close this issue since the main feature request (general support for Azure has been implemented in #458). @AKlaus If you have some additional ideas or feature requests, please feel free to open a new issue. Having separate issues helps with avoiding confusion :)

PR #407, opened 2020-07-05
Azure store implementation

Benehiko:
Azure store implementation involves the usage of Append Blobs and Block Blobs.
Append Blobs are used for files under 195GB 
Block Blobs are used for files above 195GB
Info file is by default an Append Blob.

Issue #401

Please note, this implementation still needs automated tests. As far as I have tested manually (with go-tus and tus-js-client) file can be paused, restarted, "crash" from server side and "crash" from client side still continued the file upload.

PR #458, opened 2021-02-02
Implement Azure store

omBratteng:
Related issue #401

This store uses only Block Blobs. Therefore supports files up to 190.7TiB.
Uses the latest version of https://github.com/Azure/azure-storage-blob-go and supports [Microsoft Service Version `2019-12-12`](https://docs.microsoft.com/en-us/rest/api/storageservices/version-2019-12-12)

Includes automated test for the store, but not the service. As I have not been able to use gock to mock the HTTP responses, I think this is because `azure-storage-blob-go` has used their own http.Client through `azure-pipeline-go`.
But uploads have been tested with [uppy](https://uppy.io/), [tusc](https://github.com/jackhftang/tusc) and [tuspy](https://github.com/vayam/tuspy). With chunk sizes above 4MiB

Additional features allows you to specify the access type of the container upon creating (if it does not already exist, limited by the storage account support).
https://docs.microsoft.com/en-us/azure/storage/blobs/anonymous-read-access-configure

And the blob access tier; inferred (inherited from container default), hot, cool and archive.
https://docs.microsoft.com/en-us/azure/storage/common/storage-account-overview#access-tiers-for-block-blob-data


omBratteng:
I just realised the tests uses Go v1.12.x and v1.13.x. Which causes some trouble.
By downgrading `azure-storage-blob-go` to `v0.11.0`, it would pass in go v1.13.x
But

omBratteng:
@Acconut I did a test in our repo, by changing the test targets to match your https://github.com/tus/tusd/blob/master/docs/installation.md, i.e. v1.14 and v1.15, tests are passing fine.
https://github.com/arkivverket/tusd/actions/runs/530638133

The branch I tested on is based on this branch, only change I did was the testing matrix, and I removed the build stage, as it would never run in my test branch.
https://github.com/arkivverket/tusd/commit/fa264a35565f2366cd5052384081b1f1cb118250

Please let me know if you have any questions, or need some changes done.
As stated earlier, we have assigned resources to be able to resolve this implementation.

Acconut:
First of all, thank you for this amazing PR and apologies for my late reply! 

> But uploads have been tested with [uppy](https://uppy.io/), [tusc](https://github.com/jackhftang/tusc) and [tuspy](https://github.com/vayam/tuspy). With chunk sizes above 4MiB

Have you also tested this without specifying any chunk size at all? For example, in Uppy the `chunkSize` option is not mandatory and rarely necessary.

> Which means 1.14 and 1.15 should be the test targets?

Yes, feel free to adjust the Go versions in your branch!

> Please let me know if you have any questions, or need some changes done.

I wanted to have a look at this today but noticed that @Benehiko also updated their PR https://github.com/tus/tusd/pull/407 in the meantime. Maybe you can head over there quickly and see which PR is better and on which we should move our focus :)

omBratteng:
> > But uploads have been tested with [uppy](https://uppy.io/), [tusc](https://github.com/jackhftang/tusc) and [tuspy](https://github.com/vayam/tuspy). With chunk sizes above 4MiB
> 
> Have you also tested this without specifying any chunk size at all? For example, in Uppy the `chunkSize` option is not mandatory and rarely necessary.

Yes, I have tried without specifying any chunksize, it it too, works nicely.

> > Which means 1.14 and 1.15 should be the test targets?
> 
> Yes, feel free to adjust the Go versions in your branch!

I have adjusted the go versions in the workflow to be 1.15 and 1.16, as they're the latest versions now.
And I changed the test strategy to not fail fast, so it allows all tests to run, even if other fails. That way, in my opinion, it is easier to trace any testing issues related specifically to an OS version or go version being tested. But that commit can be reverted, let me know what you think.

> > Please let me know if you have any questions, or need some changes done.
> 
> I wanted to have a look at this today but noticed that @Benehiko also updated their PR #407 in the meantime. Maybe you can head over there quickly and see which PR is better and on which we should move our focus :)

I might be biased and prefer my PR.
But I have implemented the possibility to change the blob tier. E.g. you only upload for long term archiving, you don't need it to be in the hot tier, but perhaps cool or archive tier.
Same goes with the container (bucket) access type.

omBratteng:
@Acconut In the future, it might be nice to look at adding concurrent support for this implementation. BlockBlobs lets you upload blocks in any order you want, and when you're done, commit the order of the blocks (block list).

Benehiko (commented):
I have reviewed this and it looks really similar to my last commit before my major refactor which can be found here https://github.com/Benehiko/tusd/commit/df249230ad0d79b1952b8b52658a82006cc7730c

I'm also not really sure what is implemented better here than from mine. Sure there are 2 flags to customise when creating a container and a blob. I also see here it only implements block blobs, which is fine i guess. According to the microsoft docs "An append blob is comprised of blocks and is optimized for append operations." - so i guess the tusd server might benefit from this when tons of "smaller" files are uploaded, which can be the size of 195GiB.

erlendvollset:
Really nice work here! Any updates on getting this or #407 merged? I have a use case requiring gcp, s3, _and_ azure blob backends, so I would really like to use this feature.

omBratteng:
@Acconut I see there's an upcoming v2 release, I've tested my PR against the latest v2 branch, and so far, it is working fine. Although the tests are failing on s3store, but I see they're doing the same here (https://github.com/tus/tusd/actions/runs/885415253)

Acconut:
Thank you very much for the hard work and the patience with us! Since the other PR has been closed by its maintainer, I am going to merge this beast and so how it holds up! :tada:

Acconut (approved):

PR #460, opened 2021-02-23
Customisable FileStore information storage engine for handler.FileInfo

ChrisNapthine:
Add ability to make use of an alternative mechanism for storing/retrieving/removing handler.FileInfo on behalf of a FileStore.

The original implementation is used by default, or when no alternative is provided.

This provides the ability to move away from using '{file_name}.info' files if required (or even just to locate them under an alternative folder using the NewFileInfoStore(path string) factory method), and use another 'information storage backend', eg. DB.

PS: My first PR so apologies if not relevant, feel free to incorporate as you wish :)

### DIFF
--- a/cmd/tusd/cli/composer.go
+++ b/cmd/tusd/cli/composer.go
@@ -97,7 +97,7 @@ func CreateComposer() {
 			stderr.Fatalf("Unable to ensure directory exists: %s", err)
 		}
 
-		store := filestore.New(dir)
+		store := filestore.NewFileStore(dir)
 		store.UseIn(Composer)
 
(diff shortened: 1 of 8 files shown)


Acconut:
Thank you very much for this large PR, Christopher! We have thought about adding the capability to change the storage of the upload information, but it has not been a priority yet. Although it does open up some interesting functionalities!

That being said, we would prefer such an approach to be more global and work with all other storages as well. It should be natively integrated into the tusd binary instead of just working for the FileStore.

So, all in all, I do not think this PR is likely to land in its current form but we are always happy to talk about how we can implement this better, if you are interested. I hope you understand :)

PR #461, opened 2021-03-11
Compile tusd for arm64

omBratteng:
Add arm64 binaries for linux.
https://github.com/tus/tusd/pull/458 adds arm64 binaries for darwin (macOS)


Acconut (approved):
Thanks!

PR #462, opened 2021-03-11
S3Store: Fix data inconsistency bug when using Upload-Defer-Length

### LINKED ISSUE #396: When using deferred length with S3 compatible service, last part of multi-upload is ignored
**Describe the bug**
I use tusd with an S3 compatible service (Scaleway). When I upload a file by chunks with Upload-Defer-Length to 1, the last part is not concatenated with the previous parts.

See following screenshot, obtained with chunk size of 5 Mb.

sceenshot-s3-multi-upload

Here is the content of .info file: 

```json
{
 "SizeIsDeferred": false,
 "PartialUploads": null,
 "Storage": {
  "Key": "d484497fc33fbd87bf6d698a01a14e69",
  "Bucket": "essentials-test",
  "Type": "s3store"
 },
 "Offset": 10485760,
 "IsFinal": false,
 "ID": "d484497fc33fbd87bf6d698a01a14e69+NmQxZTY2NDUtYWQxYi00ZWMyLThlMmMtN2M4OThiZDU2Zjli",
 "IsPartial": false,
 "Size": 10691054,
 "MetaData": {}
}
```

**Setup details**
- Operating System: Linux
- Used tusd version: 1.3.0 docker
- Used tusd data storage: AWS S3
- Used tusd configuration: -s3-endpoint https://s3.fr-par.scw.cloud -s3-bucket essentials-test --hooks-dir /srv/tusd-hooks
- Used tus client library: tus-js-client

acj:
Fixes #396, which has a rich discussion of the bug. There is a subtle interaction between UnroutedHandler and S3Store when using deferred length. The result is that the last chunk of an upload, if it's less than 5MB, is ignored.

### DIFF
--- a/pkg/s3store/s3store.go
+++ b/pkg/s3store/s3store.go
@@ -895,7 +895,7 @@ func (upload *s3Upload) concatUsingMultipart(ctx context.Context, partialUploads
 	return upload.FinishUpload(ctx)
 }
 
-func (upload s3Upload) DeclareLength(ctx context.Context, length int64) error {
+func (upload *s3Upload) DeclareLength(ctx context.Context, length int64) error {
 	info, err := upload.GetInfo(ctx)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


Acconut (approved):
Looks good, thank you!

PR #481, opened 2021-05-20
S3Store: fix GCS delete failure with single deletes fallback

AndyHovingh:
Deleting an upload in Google Cloud Storage via the S3 SDK fails due to multi-delete not being supported (https://cloud.google.com/storage/docs/migrating#methods-comparison). This adds a fallback to single deletes in that case.


Acconut:
Closing this issue due to inactivity and because I am not sure if this is necessary. One could also use GCS via the gcsstore directly, instead of going through the S3 API. Feel free to leave a comment if you want to continue the discussion :)

Issue #431, opened 2020-10-07
Authentication example using hooks


wizardlevel9:
Can you please provide a detailed example of how i would set up authentication (username and password) using hooks.

The documentation outlines that this is possible, but does not provide any guidance or examples.

I would like to take advantage of this fantastic capability, but I'm not sure how to do it correctly.

Any documentation or assistance with examples would be greatly appreciated.

Thanks in advance.

Acconut:
I can't provide you with a "detailed example" right now but the basic gist is that you configure the tus client to put the authentication credentials (e.g. password or token) into an header, such as `Authentication`. The tusd hooks have access to these headers (see https://github.com/tus/tusd/blob/master/docs/hooks.md#the-hooks-environment) and can validate and check the credentials. If authentication fails, the hook should error out and tusd will reject the upload. I hope that makes sense.

benitogf:
I was able to implement authentication using the http hook:

```
tusd -hooks-http="http://localhost:8081/tusd/hook"
```

```golang
  server.Router.HandleFunc("/tusd/hook", func(w http.ResponseWriter, r *http.Request) {
    // https://github.com/tus/tusd/blob/master/docs/hooks.md#usage
    // log.Println("HOOK: ", r.Header.Get("hook-name"))
    hookName := r.Header.Get("hook-name")
    _, account, _ := auth.Audit(r)
    if account == "" && hookName == "pre-create" {
	    log.Println("file upload not authorized")
	    w.WriteHeader(http.StatusUnauthorized)
	    return
    }
    w.WriteHeader(http.StatusOK)
  }).Methods(http.MethodPost)
```

but I would like to limit file reading as well, is there a way to do this? I think that adding a read hook would be useful

Acconut:
> I would like to limit file reading as well, is there a way to do this?

I am not sure what you mean by this. However, tusd has a `MaxSize` setting, which controls the maximum allowed upload size.

benitogf:
> I am not sure what you mean by this

Sorry I meant read access limit, a way to prevent unwanted file reads

Acconut:
> Sorry I meant read access limit, a way to prevent unwanted file reads

This is not possible using hooks at the moment. When using the `tusd` binary the only option right now is to put a proxy in front of tusd to handle authentication for GET requests. There is no hook emitted for GET requests.

Acconut:
[...]

PR #483, opened 2021-05-27
add a pre-get hook

benitogf:
following up on https://github.com/tus/tusd/issues/431

not sure if this feature would be welcomed, please advice @Acconut

### DIFF
--- a/cmd/tusd/cli/hooks.go
+++ b/cmd/tusd/cli/hooks.go
@@ -43,6 +43,10 @@ func preFinishCallback(info handler.HookEvent) error {
 	return hookCallback(hooks.HookPreFinish, info)
 }
 
+func preGetCallback(info handler.HookEvent) error {
+	return hookCallback(hooks.HookPreGet, info)
+}
+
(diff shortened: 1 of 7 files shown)


Acconut:
Thank you for the PR and apologies for my delayed response! I can see the value of a `pre-get` hook. You are probably using it for authorization/authentication as you mentioned in #431, aren't you? For those cases, I think it would make more sense to implement a proper authentication system using hooks where a hook is always invoked before a user tries to access an upload. The hook can then decide if this access should be granted or not.

Such a system is more complex to implement but would serve more use cases. What do you think?

benitogf:
sounds good! thanks, we can close this one then and create an issue to document the requirements when possible

> This avoids invoking the hook if the upload does not exist

not sure about this one, if there's no authorization it should not return a 404 just not authorized imho

Acconut:
Superseeded by https://github.com/tus/tusd/pull/1077, which implements a similar mechanism, which is not restricted to downloads.

PR #489, opened 2021-06-09
Change timestamp to microseconds precision in log

jjrodrig:
Currently log config is writing timestamp with second  precision but usually this is not enough to correctly track the requests. 
This PR proposes to use the microsecond precision for the logging timestamps. 

Loking at the standar log package (https://golang.org/pkg/log/#pkg-constants), there is no alternative to the microsecond precision.


Acconut:
Looks good! Let's try that out, thanks!

PR #495, opened 2021-07-27
Only determine object type based on name after last separator

tinco:
Ran into a problem when Store.ObjectPrefix is set to a path that has an underscore in it. I think this solves it, unless you need to support other separators than "/", which I believe is configurable.

### DIFF
--- a/pkg/gcsstore/gcsservice.go
+++ b/pkg/gcsstore/gcsservice.go
@@ -351,7 +351,11 @@ loop:
 		if strings.HasSuffix(objAttrs.Name, "info") {
 			continue
 		}
-		split := strings.Split(objAttrs.Name, "_")
+
+		fileNameParts := strings.Split(objAttrs.Name, "/")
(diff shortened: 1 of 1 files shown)


Acconut:
Thanks for the PR, looks very helpful!

Would you mind adding a test case for this, so we can ensure that there is no regression in the future regarding the handling of underscores? Something similar to https://github.com/tus/tusd/blob/master/pkg/gcsstore/gcsservice_test.go#L446, but testing with a prefix containing an underscore.

In addition, could you update the description of the `-gcs-object-prefix` flag, which now allows underscores: https://github.com/tus/tusd/blob/master/cmd/tusd/cli/flags.go#L72

tinco:
I modified the existing test case to have an underscore, verified that the test fails without my fix, and passes with my fix. I also updated the readme :)

Acconut (approved):
Thank you very much for the updates and apologies for the delay on my side!

Issue #386, opened 2020-05-19
Option to disable CORS headers


sripberger:
**Is your feature request related to a problem? Please describe.**
I'm seeing that you guys are blindly setting `Access-Control-Allow-Origin` to match the `Origin` header of the request, regardless of what that Origin header is. That's nice for ease of use, but it isn't exactly a good idea on the public internet.

You don't provide any options for changing CORS behavior, and I'm seeing in #135 that the response to this is that in production you're going to want to put `tusd` behind a proxy of some kind, anyway, and should probably have the proxy handle the CORS headers instead. This is totally reasonable, but it can present some challenges and pitfalls.

In my case I'm using [haproxy][1], and I can have it conditionally set `Access-Control-Origin` only for Origins that match a Regex contained in a file with something like this:

```
capture request header origin len 128
http-response add-header Access-Control-Allow-Origin %[capture.req.hdr(0)] if { capture.req.hdr(0) -m reg -f /etc/haproxy/cors-origins.lst 
```

The problem here is that `add-header` appends to what is already there, so this header will be set to `${Origin}, ${Origin}` for requests that hit `tusd`, which is obviously will cause the request to be blocked by the browser.

If I change this configuration to use `set-header` instead, which overwrites what may or may not already be there, it will get rid of this problem. However, having this collision go undetected creates a more serious problem, which is that `tusd` is still setting the header for non-matching origins, and the proxy is not clearing this. So even though haproxy is checking against the regexes, I am still allowing requests from all origins if they terminate at `tusd`.

To fix this, I have to actively delete the header in my configuration for non-matching origins. This isn't too complicated, since it's just a matter of starting by deleting whatever might be there:

```
capture request header origin len 128
http-response delete-header Access-Control-Allow-Origin
http-response set-header Access-Control-Allow-Origin %[capture.req.hdr(0)] if { capture.req.hdr(0) -m reg -f /etc/haproxy/cors-origins.lst
http-
```

It would be nice if it were cleaner, though. Having to explicitly delete something that shouldn't even be there in the first place is a bit of a kludge, to the point where I feel the need to leave a comment in the config file making note of `tusd`'s behavior that causes this `delete header` line to be required.

**Describe the solution you'd like**
To handle this problem-- and other confusion arising from multiple processes being responsible for CORS headers-- it would be nice to have an option of disabling CORS headers completely in `tusd`. That way I can easily just say that CORS is handled completely by my proxy.

Something  like `tusd --no-cors`.

[...]

PR #504, opened 2021-09-04
Add flag to control CORS Origin header

### LINKED ISSUE #135: Support custom headers for CORS
I'm using the oficial JS client with the oficial go server, the problem is that I need to send a custom header when it's JS code, and the [tusd middleware](https://github.com/tus/tusd/blob/master/unrouted_handler.go#L140) has fixed CORS headers.

It would be good to have a way to set custom CORS headers.

sean9999:
I added a flag to allow the user to set their own Access-Control-Allow-Origin header. I feel it's proper to be able to set this header because CORS was designed with this use-case in mind. You need to be able explicitly say what origin can access your API. Using a reverse-proxy felt kludgy, because the app already sets CORS headers. It is not offloading that responsibility. This resolves #135 and #386.

I developed this for personal use because I am running tusd in a Kubernetes environment, and because of the way k8s handles network routing via Ingress controller, setting the proper Origin header was infeasible (even with --behind-proxy turned on). Using a reverse proxy felt wrong because k8s already does a lot of network indirection.

I hope you find this useful

### DIFF
--- a/pkg/handler/unrouted_handler.go
+++ b/pkg/handler/unrouted_handler.go
@@ -217,15 +217,15 @@ func (handler *UnroutedHandler) Middleware(h http.Handler) http.Handler {
 
 		header := w.Header()
 
-                var origin = handler.config.CorsOrigin
-                if origin == "" {
-                        origin = r.Header.Get("Origin")
-                }
+		var origin = handler.config.CorsOrigin
+		if origin == "" {
+			origin = r.Header.Get("Origin")
+		}
 
-                if origin != "" {
+		if origin != "" {
 
-                        header.Set("Access-Control-Allow-Origin", origin)
-                        header.Set("Vary", "Origin")
+			header.Set("Access-Control-Allow-Origin", origin)
+			header.Set("Vary", "Origin")
 
 			if r.Method == "OPTIONS" {
 				// Preflight request
--- a/pkg/handler/unrouted_handler.go
+++ b/pkg/handler/unrouted_handler.go
@@ -217,8 +217,15 @@ func (handler *UnroutedHandler) Middleware(h http.Handler) http.Handler {
 
 		header := w.Header()
 
-		if origin := r.Header.Get("Origin"); origin != "" {
-			header.Set("Access-Control-Allow-Origin", origin)
+                var origin = handler.config.CorsOrigin
+                if origin == "" {
+                        origin = r.Header.Get("Origin")
+                }
+
+                if origin != "" {
+
+                        header.Set("Access-Control-Allow-Origin", origin)
+                        header.Set("Vary", "Origin")
 
 			if r.Method == "OPTIONS" {
 				// Preflight request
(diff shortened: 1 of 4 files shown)


pkg/handler/unrouted_handler.go
```diff
@@ -217,8 +217,15 @@ func (handler *UnroutedHandler) Middleware(h http.Handler) http.Handler {
...
-		if origin := r.Header.Get("Origin"); origin != "" {
+		var origin = handler.config.CorsOrigin
+		if origin == "" {
+			origin = r.Header.Get("Origin")
+		}
+
+		if origin != "" {
+
 			header.Set("Access-Control-Allow-Origin", origin)
+			header.Set("Vary", "Origin")
```
sean9999, line 228: This is just for correctness ( https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin#cors_and_caching )

DeepDiver1975:
> : Is it not necessary to validate the the Origin header in the request matches `config.CorsOrigin`?



> Right now, I don't see such a test but only that `config.CorsOrigin` is always set as the header value.

The most simplistic setup on apache/nginx is also just adding the origin with no further logic applied. I'd say this is okay to go for the very first iteration. refs https://ubiq.co/tech-blog/enable-cors-apache-web-server/ (I just grabbed to first hit on google search :see_no_evil: )

@sean9999  @Acconut is there anything I can help out with to get this PR going? THX a lot

patrickjahns:
Currently also experimenting with tusd in a kubernetes setup. Adding this functionality would gladly improve the setup.

@sean9999 @Acconut  - is there anything we can do to move this PR forward?

G2G2G2G:
Yea adding this would be nice.. For now I use `proxy_hide_header Access-Control-Allow-Origin;` on nginx which disables their bad header.

sean9999:
> Is it not necessary to validate the the Origin header in the request matches `config.CorsOrigin`? If they are not equal, we could return an error. Or is this not necessary? Right now, I don't see such a test but only that `config.CorsOrigin` is always set as the header value.

You're right. I checked the spec, and a CORS compliant server is supposed to refuse a connection or return an error in that case. I'll try adding that in. https://www.w3.org/TR/2020/SPSD-cors-20200602/#resource-requests

Acconut:
> I'll try adding that in

Sounds good! After that, I think we can merge this :)

DeepDiver1975:
Took the freedom to rebase this PR -> #942

Acconut:
> > Is it not necessary to validate the the Origin header in the request matches `config.CorsOrigin`? If they are not equal, we could return an error. Or is this not necessary? Right now, I don't see such a test but only that `config.CorsOrigin` is always set as the header value.
> 
> You're right. I checked the spec, and a CORS compliant server is supposed to refuse a connection or return an error in that case. I'll try adding that in. https://www.w3.org/TR/2020/SPSD-cors-20200602/#resource-requests

For anyone interested. This would be the last point that needs to be done. If this check is implement, we can move towards merging such a feature :)

DeepDiver1975:
can be closed as well - merged as part of https://github.com/tus/tusd/pull/987

Acconut:
This is superseeded by #997. Thank you for all your help!

Issue #506, opened 2021-09-21
integration with digitalocean spaces


benitogf:
currently I'm not able to use tusd with digitalocean (s3 compatible service) spaces
I'm able to upload files, but not getting them through tusd

**To Reproduce**

I'm using this systemd service file to run tusd:

```
[Install]
WantedBy=default.target

[Unit]
Description=tusd service
After=syslog.target network.target multi-user.target

[Service]
Environment="AWS_ACCESS_KEY_ID={DO_ACCESS_KEY}"
Environment="AWS_SECRET_ACCESS_KEY={DO_SECRET_KEY}"
Environment="AWS_REGION=us-east-1"
ExecStart=/usr/bin/tusd -host="127.0.0.1" -s3-bucket="{DO_SPACE_NAME}" -s3-endpoint="{DO_SPACE_ENDPOINT}"
Restart=always
RestartSec=1
```

**Expected behavior**

I tested getting the uploaded files through `s3cmd` and it works, but the tusd api to fetch them cannot find them

**Setup details**
Please provide following details, if applicable to your situation:
- Operating System: Ubuntu 20.04
- Used tusd version: v1.6.0
- Used tusd data storage:  digital ocean spaces (s3)
- Used tusd configuration: s3-bucket and s3-endpoint
- Used tus client library: tus-js-client

Issue #465, opened 2021-03-18
Could not download file using ceph


bluven:
**Describe the bug**
After uploading file, I tried to download file, tusd reports error like this:
```
NoSuchKey: 
	status code: 404, request id: tx000000000000000022c0f-00605304fa-296a6-my-store, host id: 
```

**To Reproduce**
1. `tusc client http://127.0.0.1:18080/files kibana-7.6.2-linux-x86_64.tar.gz -r --chunk-size=5242880`
2. Use browser to download file using the link returned by tusc

**Expected behavior**
File should be downloaded.

**Setup details**
Please provide following details, if applicable to your situation:
- Operating System: Ubuntu
- Used tusd version: I built myself
- Used tusd data storage: Ceph
- Used tusd configuration: 
```
export AWS_ACCESS_KEY_ID=X9OGMOAFERCT6Y09GYV0
export AWS_SECRET_ACCESS_KEY=ALg6uCkPdM8PbqVlVsIuFN23ENCeJ2Ta1woVtW0F
export AWS_REGION=us-east-1
go build -o e-tusd  ./cmd/e-tusd
./e-tusd -s3-bucket=bluven \
    -s3-endpoint=http://10.20.8.169:32222 \
    -port=18080 \
```
PS: I add an extra endpoint to implement resumable download. 

- Used tus client library: [tusc](https://github.com/jackhftang/tusc)

---------------------------------------------------------

I checked s3store's source code  and found that:
```
        parts, err := store.listAllParts(ctx, id)
	if err != nil {
		if isAwsError(err, "NoSuchUpload") {
			info.Offset = info.Size
			return info, nil
		} else {
			return info, err
		}
	}
```

This snippet was found in **fetchInfo** method which called **listAllParts** to get parts info, since upload was finished, an error was returned.
But here **fetchInfo** just checked **NoSuchUpload** error, while Ceph for some reason return an **NoSuchKey** error and this caused download failure. 

Here is what I got from wireshark for listParts request:
```
Minio(yeah, I also tried on minio)

<?xml version="1.0" encoding="UTF-8"?>
[...]

PR #507, opened 2021-09-24
fix digitalocean spaces fetch

benitogf:
related to https://github.com/tus/tusd/issues/465

https://github.com/tus/tusd/issues/506


Acconut:
Thank you for this PR, but I don´t think I fully understand its purpose. What was the error from DigitalOcean Spaces? How does this PR fix this and why is the `info.Size > 0` check needed?

benitogf:
Hello, the error from DigitalOcean spaces is that it returns `NoSuchKey` instead of `NoSuchUpload` on the condition described by the comment above the check `when the multipart upload has already been completed or aborted`

Will remove the `info.Size > 0` check since its already here https://github.com/tus/tusd/blob/0ad435b4c80bbd8fbaae784206e427d6b1316688/pkg/handler/unrouted_handler.go#L751

Acconut (approved):
Thank you very much for the explanation and the PR!

Issue #511, opened 2021-10-13
Update Docker build process


Acconut:
Previously we used the Docker Build service to build the Docker images by Docker hub themselves. Since they changed their pricing, this feature is no longer available and the Docker images there are outdated: https://hub.docker.com/repository/docker/tusproject/tusd/tags?page=1&ordering=last_updated

Either we try to get a free subscription as an open-source project or build the Docker images on our own and push them to the registry.

omBratteng:
Let me know if you need any help with this @Acconut.

Acconut:
> Let me know if you need any help with this @Acconut.

Thank you very much for your offering, @omBratteng! That would be very helpful. I don't use Docker on a daily basis, so my knowledge regarding Docker is a bit rusty :)

If I understand our current CI setup correctly, we already build the Docker image on GHA and push it to the registry using the last commit as the tag: https://github.com/tus/tusd/blob/16a3747ec421c041587f995ef4a00dc3e4ca004d/.github/workflows/main.yml#L54-L61

This part seems to work fine, as I can see those tags, e.g. https://hub.docker.com/layers/tusproject/tusd/dd44267c082c98e99f08ca88ad85eb0b7fe93ef6/images/sha256-bf372cfa53e6c0314104cc9dcf66b28f7b6957a4276259a527239cfc507366ea?context=explore

In that case, the only thing we need to do, is make the same image available under the `latest` tag and the corresponding version tag, e.g. `v1.7.1`. Is that correct? Do you what the commands for that would be?

> what about gitub registery ?

The GitHub registry is a possible improvement in the future. But as the Docker registry is the default for many developers, I would like to see an updated image on Docker Hub first :)

omBratteng:
@Acconut take a look at #511, reach out to me if you want to discuss it

PR #515, opened 2021-10-18
New CI and release workflow

omBratteng:
CI and build/deploy is now a separate workflow


.github/workflows/release.yaml
```diff
@@ -0,0 +1,113 @@
...
+      -
+        name: Docker meta
+        id: docker_meta
+        uses: docker/metadata-action@v3.5.0
+        with:
+          images: |
+            ghcr.io/tus/tusd
+            tusproject/tusd
+          flavor: |
+            latest=${{ startsWith(github.ref, 'refs/tags/') }}
```
Acconut, line 30: Does this set the `latest` tag for the current image?

Dockerfile
```diff
@@ -1,37 +1,43 @@
-FROM golang:1.16-alpine AS builder
+FROM golang:1.17.2-alpine AS builder
```
Acconut, line 1: Could you explain why you made these the changes in the Dockerfile? Do they enable multiarch builds, as you mentioned in https://github.com/tus/tusd/issues/511#issuecomment-945778847?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,113 @@
...
+          tags: |
+            type=sha
+            type=semver,pattern=v{{version}}
+            type=semver,pattern=v{{major}}.{{minor}}
+            type=semver,pattern=v{{major}}
```
Acconut, line 35: Wow, that metadata action looks awesome :heart_eyes:

.github/workflows/release.yaml
Acconut: One more thing: We would like to only set the latest tag if the tag is no pre-release (i.e. not of the form `v1.2.3-pre1`). In that case, the tags v1 and v1.2 should also not be set. Is that easily possible or should we use an entirely different tag format for pre-releases?

.github/workflows/release.yaml
omBratteng: It is a pretty sweet action, does some great metadata magic

Dockerfile
Acconut: I was talking about all changes in the Dockerfile :) However, you second comment already explains what I wanted to know!

.github/workflows/release.yaml
omBratteng: If you check the runs I did, under the step Docker meta, and the Docker tags section, you see what tags it generates

`v300.0.0-pre1`
image

`v200.1.2`
image

`push to master` (in the test run it's to branch called `local-testing` for a quick test)
image

PR #598, opened 2021-12-16
s3store: stricter character replacement regexp for x-amz-metadata

marodere:
What's up: some upload-metadata values will be successfully delivered (thanks to b64encode) but then cause an error communicating with S3.

How to reproduce:
```
$ curl -X POST -H 'Upload-Metadata: filename dGVzdC5/Lm1wNA==' [...]
[...]
500 Internal Server Error: s3store: unable to create multipart upload:
RequestError: send request failed
caused by: Post "https://s3.example.net/tusd/d167476518c4ac7e01d3a9bd7beaa0ce?uploads=": net/http: invalid header field value "test.\u007f.mp4" for key X-Amz-Meta-Filename
```

Tracing back lead us here https://cs.opensource.google/go/go/+/master:src/net/http/transport.go;l=528
then here https://pkg.go.dev/golang.org/x/net/http/httpguts#ValidHeaderFieldValue
and eventually refers to a few HTTP RFCs.

From `net/http` implementation (as well as from the RFCs) could be seen that some characters (in particular 0x7f) should not be considered valid, that's why the regexp should be changed.

### DIFF
--- a/pkg/s3store/s3store.go
+++ b/pkg/s3store/s3store.go
@@ -94,7 +94,7 @@ import (
 
 // This regular expression matches every character which is not
 // considered valid into a header value according to RFC2616.
-var nonPrintableRegexp = regexp.MustCompile(`[^\x09,\x20-\x7E]`)
+var nonPrintableRegexp = regexp.MustCompile(`[^\x09\x20-\x7E]`)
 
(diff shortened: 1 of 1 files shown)


Acconut (approved):
Thank you very much for the fix!

PR #664, opened 2022-02-21
Fix leakage of info when no azure blob is found

### LINKED ISSUE #661: Azure storage backend leaks information
**Describe the bug**
Azure storage backend leaks information:
* The used client library including version
* The underlying error
* The Storage Account Name / URL
* X-Ms-Client-Request-Id
* X-Ms-Request-Id

Example:
```
-> github.com/Azure/azure-storage-blob-go/azblob.newStorageError, github.com/Azure/azure-storage-blob-go@v0.14.0/azblob/zc_storage_error.go:42
===== RESPONSE ERROR (ServiceCode=BlobNotFound) =====
Description=The specified blob does not exist.
RequestId:(*redacted for github issue*)
Time:2022-02-18T10:47:40.9444083Z, Details: 
   Code: BlobNotFound
   GET https://somecontainername.blob.core.windows.net/tusd/foo.info?timeout=61
   Authorization: REDACTED
   User-Agent: [Azure-Storage/0.14 (go1.17.2; linux)]
   X-Ms-Client-Request-Id: [(*redacted for github issue*)]
   X-Ms-Date: [Fri, 18 Feb 2022 10:47:40 GMT]
   X-Ms-Version: [2020-04-08]
   --------------------------------------------------------------------------------
   RESPONSE Status: 404 The specified blob does not exist.
   Content-Length: [215]
   Content-Type: [application/xml]
   Date: [Fri, 18 Feb 2022 10:47:40 GMT]
   Server: [Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0]
   X-Ms-Client-Request-Id: [(*redacted for github issue*)]
   X-Ms-Error-Code: [BlobNotFound]
   X-Ms-Request-Id: [(*redacted for github issue*)]
   X-Ms-Version: [2020-04-08]
```

**To Reproduce**
1. Setup a storage account in Azure
2. Start tusd using Azure as Storage Backend
`./tusd -azure-storage=tusd`
3. Send HTTP GET to file upload endpoint
curl 'http://tusd:1080/files/foo'
4. Inspect the delivered response

**Expected behavior**
The response to the client should be general 404, or 500 Error.
It should not contain details about the used storage backend.
No details should be included.
Consistently with the local file upload it should return status 404 with string `upload not found`.

**Setup details**
Please provide following details, if applicable to your situation:
- Operating System: Linux
- Used tusd version:
```
./tusd -version
Version: v1.8.0
Commit: 48ffebec56fcf3221461b3f8cbe000e5367e2d48
Date: Fri Oct 29 09:32:00 UTC 2021
```
- Used tusd data storage: Azure Blob Storage
- Used tusd configuration: `./tusd -azure-storage=tusd`
- Used tus client library: curl

omBratteng:
Fixes #661

### DIFF
--- a/pkg/azurestore/azureservice.go
+++ b/pkg/azurestore/azureservice.go
@@ -308,3 +308,10 @@ func blockIDBase64ToInt(blockID string) int {
 	blockIDBase64ToBinary(blockID)
 	return int(binary.LittleEndian.Uint32(blockIDBase64ToBinary(blockID)))
 }
+
+func isAzureError(err error, code string) bool {
+	if err, ok := err.(azblob.StorageError); ok && string(err.ServiceCode()) == code {
+		return true
+	}
+	return false
+}
--- a/pkg/azurestore/azureservice.go
+++ b/pkg/azurestore/azureservice.go
@@ -201,7 +201,7 @@ func (blockBlob *BlockBlob) GetOffset(ctx context.Context) (int64, error) {
 
 	getBlock, err := blockBlob.Blob.GetBlockList(ctx, azblob.BlockListAll, azblob.LeaseAccessConditions{})
 	if err != nil {
-		if err.(azblob.StorageError).ServiceCode() == azblob.ServiceCodeBlobNotFound {
+		if isAzureError(err, "BlobNotFound") {
 			return 0, nil
 		}
 
--- a/pkg/azurestore/azureservice.go
+++ b/pkg/azurestore/azureservice.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 
 	"github.com/Azure/azure-storage-blob-go/azblob"
+	"github.com/tus/tusd/pkg/handler"
 )
 
 const (
@@ -261,6 +262,9 @@ func (infoBlob *InfoBlob) Download(ctx context.Context) ([]byte, error) {
 		return nil, fmt.Errorf("File %s does not exist", infoBlob.Blob.ToBlockBlobURL())
 	}
 	if err != nil {
+		if isAzureError(err, "BlobNotFound") {
+			err = handler.ErrNotFound
+		}
 		return nil, err
 	}
 
--- a/pkg/azurestore/azureservice.go
+++ b/pkg/azurestore/azureservice.go
@@ -202,7 +202,7 @@ func (blockBlob *BlockBlob) GetOffset(ctx context.Context) (int64, error) {
 	getBlock, err := blockBlob.Blob.GetBlockList(ctx, azblob.BlockListAll, azblob.LeaseAccessConditions{})
 	if err != nil {
 		if isAzureError(err, "BlobNotFound") {
-			return 0, nil
+			err = nil
 		}
 
 		return 0, err
--- a/pkg/azurestore/azureservice.go
+++ b/pkg/azurestore/azureservice.go
@@ -206,7 +206,7 @@ func (blockBlob *BlockBlob) GetOffset(ctx context.Context) (int64, error) {
 	getBlock, err := blockBlob.Blob.GetBlockList(ctx, azblob.BlockListAll, azblob.LeaseAccessConditions{})
 	if err != nil {
 		if isAzureError(err, "BlobNotFound") {
-			err = nil
+			err = handler.ErrNotFound
 		}
 
 		return 0, err
--- a/pkg/azurestore/azureservice.go
+++ b/pkg/azurestore/azureservice.go
@@ -176,9 +176,13 @@ func (blockBlob *BlockBlob) Download(ctx context.Context) (data []byte, err erro
 
 	// If the file does not exist, it will not return an error, but a 404 status and body
 	if downloadResponse != nil && downloadResponse.StatusCode() == 404 {
-		return nil, fmt.Errorf("File %s does not exist", blockBlob.Blob.ToBlockBlobURL())
+		return nil, handler.ErrNotFound
 	}
 	if err != nil {
+		// This might occur when the blob is being uploaded, but a block list has not been committed yet
+		if isAzureError(err, "BlobNotFound") {
+			err = handler.ErrNotFound
+		}
 		return nil, err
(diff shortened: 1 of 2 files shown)


omBratteng:
@Acconut 
I've ran into a possible error, or unintended behaviour from what tus expects.
When I run a `GET` on a file that does not exist, it does not return any content, but does not throw an error.
```console
❯ curl -i http://localhost:1080/files/does_not_exist
HTTP/1.1 204 No Content
Content-Disposition: attachment;filename="does_not_exist"
Content-Type: application/octet-stream
Tus-Resumable: 1.0.0
X-Content-Type-Options: nosniff
Date: Mon, 21 Feb 2022 16:55:02 GMT
```

It happens because in https://github.com/omBratteng/tusd/blob/fix-azure-store-leak/pkg/azurestore/azurestore.go#L114-L117, I try to fetch the offset of the blob, which will [return a `nil`](https://github.com/omBratteng/tusd/blob/fix-azure-store-leak/pkg/azurestore/azureservice.go#L204-L206) in the error, if no blob is found.
If I change it to `handler.ErrNotFound`, the above curl will return a 404 as expected
```console
❯ curl -i http://localhost:1080/files/does_not_exist
HTTP/1.1 404 Not Found
Content-Length: 17
Content-Type: text/plain; charset=utf-8
Tus-Resumable: 1.0.0
X-Content-Type-Options: nosniff
Date: Mon, 21 Feb 2022 17:01:12 GMT

upload not found
```

_however_, uploads no longer work, because when sending the patch request, it also uses the `GetOffset` called within `GetUpload`.
If there was a way to differentiate when `GetUpload` was called, I could add a check when getting the offset to disregard not found error on uploads.

omBratteng:
Another thing I noticed, if I upload an empty file, it also returns a 204 No content error. (tested this on the demo site too https://tus.io/demo.html with [204 No Content](https://tusd.tusdemo.net/files/57edb3434bbe98a916936b988658e53c+D2grXoKtIdVY1Zro.ltZSYbSNHAE8LYpymzbUrrbjPSUURmfIioLuCEHrZzY8_6ljI2v2XkNxs_yirB9dF70HuixLfgbdJY7xl3oZiS8Jp8PMcv6rRO._PZK09Bzyzo7)).

Tested it just by uploading a empty file

omBratteng:
Doing this change fixes downloads by not throwing a `204 No Content` but instead a `404 Not found`

```console
❯ curl -i http://localhost:1080/files/does_not_exist
HTTP/1.1 404 Not Found
Content-Length: 17
Content-Type: text/plain; charset=utf-8
Tus-Resumable: 1.0.0
X-Content-Type-Options: nosniff
Date: Mon, 21 Feb 2022 17:29:38 GMT

upload not found
```

But uploads breaks
```console
```

omBratteng:
This happens when the info file blob exists, but not the uploaded blob file

Acconut:
Thank you very much for working so quickly on this!

> _however_, uploads no longer work, because when sending the patch request, it also uses the `GetOffset` called within `GetUpload`. If there was a way to differentiate when `GetUpload` was called, I could add a check when getting the offset to disregard not found error on uploads.
> 
> This happens when the info file blob exists, but not the uploaded blob file

I don't think a differentiation to see why GetUpload is called is a good design. Instead, could we add a condition to ignore the error from `GetOffset` if an info file blob exists? This way, new upload should still be possible, but we still can get the 404 if both, info file and uploaded blob, do not exist.

> Another thing I noticed, if I upload an empty file, it also returns a 204 No content error.

This does not sound like an error to me. 204 is the correct response code when downloading an empty file. Or am I missing your point?

omBratteng:
> I don't think a differentiation to see why GetUpload is called is a good design.

I agree

> Instead, could we add a condition to ignore the error from GetOffset if an info file blob exists? This way, new upload should still be possible, but we still can get the 404 if both, info file and uploaded blob, do not exist.

That might work.
So we should not expect a file to exist, if the info file does not exist?

> > Another thing I noticed, if I upload an empty file, it also returns a 204 No content error.
> 
> This does not sound like an error to me. 204 is the correct response code when downloading an empty file. Or am I missing your point?

The header is the correct header. I just see in my browser it tries to download a file because of the `Content-Disposition` header, but browser says `Failed - no file`. So tusd is probably doing the correct thing. I just expected it to download a file with no content.

omBratteng:
I [changed GetOffset](https://github.com/tus/tusd/pull/664/commits/96d0a7b4ed5996b1c4fdc1f20af1c4b7d3256673) to return `handler.ErrNotFound`, but in GetUpload when getting the offset, I check if the returned error is not `handler.ErrNotFound`.

omBratteng:
Uploading still works, and when trying to download a file that only has a info file but not main file, you get a 204 No Content
```console
❯ curl -i http://localhost:1080/files/only_info_file
HTTP/1.1 204 No Content
Content-Disposition: attachment;filename="no_file"
Content-Type: application/octet-stream
Tus-Resumable: 1.0.0
X-Content-Type-Options: nosniff
Date: Tue, 22 Feb 2022 16:35:58 GMT
```

And if no info file exists
```console
❯ curl -i http://localhost:1080/files/no_info_file
HTTP/1.1 404 Not Found
Content-Length: 17
Content-Type: text/plain; charset=utf-8
Tus-Resumable: 1.0.0
X-Content-Type-Options: nosniff
Date: Tue, 22 Feb 2022 16:37:04 GMT

upload not found
```

Acconut:
> So we should not expect a file to exist, if the info file does not exist?

Yes, for the sake of uploads, we can make this reasoning.

> I [changed GetOffset](https://github.com/tus/tusd/pull/664/commits/96d0a7b4ed5996b1c4fdc1f20af1c4b7d3256673) to return `handler.ErrNotFound`, but in GetUpload when getting the offset, I check if the returned error is not `handler.ErrNotFound`.

:+1: Awesome, thank you!

Acconut (approved):

PR #666, opened 2022-02-22
fix: set dependabot to check weekly instead of daily

omBratteng:
I see you also suffer from daily dependabot updates regarding AWS SDK, which they release a new version daily. They do the same in their python SDK, so we've adjusted it down to weekly checks instead of daily.
That way we still get dependabot updates, but not spammed down with it.


Acconut (approved):
Thank you very much! That is very helpful!

PR #667, opened 2022-02-24
v2: Implement cancelable lock mechanism

Acconut:
The rationale behind this PR is discussed in https://github.com/tus/tusd/wiki/Thoughts-on-handling-connection-issues

The goal is following: If an upload request cannot be processed because the upload is still locked from a previous upload request, the new upload request can ask the lock holder to release the lock in a controlled manner. Then the new upload request is able to take over the upload without forwarding a locking error to the client.

TODO:
- [ ] Implement functionality for filelocker
- [ ] Adapt implementation at tusd-etcd-locker
- [ ] Add tests for new functionality in handler
- [ ] Increase default read timeout to reduce their likelihood


oliverpool:
Hi, this is brilliant!

I investigated yesterday suspicious cases of `upload is currently locked` and my conclusion is that "half-closed" TCP connections are very likely to be the culprit.

It is mentioned on https://github.com/tus/tusd/wiki/Thoughts-on-handling-connection-issues that the timeout is 6s, but we have seen up to 2 minutes in the log, where the lock was held:
- t0: WriteStart
- t0+37s: second attempt to patch (likely because the client has closed its side of the connection), resulting in http 423
- t0+112s: response to first request (with error 500), the server closed its side of the connection

So our current workaround would be to retry at least for 2 minutes.

Anyway, the mechanism described here would be very helpful for us!

---

https://github.com/tus/tusd/blob/368ea5c413db6eddda0ebcc1f1c978f64a80b6d1/pkg/handler/unrouted_handler.go#L1102-L1104
Maybe the `releaseLock` function could be wrapped in `sync.Once`, so that the `releaseLock` function is called at most once:
```go
        var once sync.Once
	if err := lock.Lock(ctx, func(){once.Do(releaseLock)}); err != nil {
		return nil, err
	}
```

That way the lock implementations don't have to take this into account (maybe this should be documented in the `Lock` interface)

Acconut:
> Anyway, the mechanism described here would be very helpful for us!

Thanks! I hope it works out as intended in the end. Feel free to download one of the v2 release candidates and test it out for yourself: https://github.com/tus/tusd/releases/tag/v2.0.0-rc8 I would love to hear some feedback on how it performs!

> Maybe the `releaseLock` function could be wrapped in `sync.Once`, so that the `releaseLock` function is called at most once

I thought of this during the development, but so far I haven't seen a need for it yet, so I didn't use sync.Once. However, if it's reasonable we can still add it later before v2 is finished :)

oliverpool:
Unfortunately I can't import the `v2` as a go module yet (see https://github.com/tus/tusd/pull/672#issuecomment-1082737523)

---

The current executable uses a file locker (to handle program restart), which doesn't support `requestRelease`.

A workaround would be to provide a `DoubleLocker` which locks in memory and as a file.
This way the `requestRelease` works at least when the lock is simply held by another goroutine:

```go
doubleLock := DoubleLocker{
	First:  memorylocker.New(),
	Second: filelocker.New("path"),
}
```
```go
type DoubleLocker struct {
	First  handler.Locker
	Second handler.Locker
}
type doubleLock struct {
	First  handler.Lock
	Second handler.Lock
}

func (locker DoubleLocker) NewLock(id string) (handler.Lock, error) {
	first, err := locker.First.NewLock(id)
	if err != nil {
		return nil, err
	}
	second, err := locker.Second.NewLock(id)
	if err != nil {
		return nil, err
	}
	return doubleLock{first, second}, nil
}

//Lock tries to obtain the exclusive lock.
func (lock doubleLock) Lock(ctx context.Context, requestRelease func()) error {
	if err := lock.First.Lock(ctx, requestRelease); err != nil {
		return err
	}

	if err := lock.Second.Lock(ctx, requestRelease); err != nil {
		if errUnlock := lock.First.Unlock(); errUnlock != nil {
			// this unlocking error is more critical than the lock error
			return errUnlock
		}
		return err
	}
	return nil
}

func (lock doubleLock) Unlock() error {
	// in reverse order
	errSecond := lock.Second.Unlock()
	errFirst := lock.First.Unlock()
	if errSecond != nil {
		return errSecond
	}
	return errFirst
}
```

Shall I make a PR with this? As a new `doublelocker` package?

Or should all the locker be gathered in a new `lock` package (`lock.InMemory`, `lock.OnDisk`, `lock.Double`) to reduce the number of packages in `v2`?

Acconut:
> A workaround would be to provide a `DoubleLocker` which locks in memory and as a file.

I don't fully understand how this should work. Do you want to use the support for `requestRelease` from the `memorylocker` to notify `filelocker` to release its lock? I am not sure if the code you have provided provides this functionality. In general, having two different types of locks at the same time also seems dangerous to me. There is not single source of truth anymore and a lot more effort is required to ensure the memorylock and filelocker are synchronized.

If a package, such as filelocker, does not support the new `requestRelease` method, then I do not see this as a problem. The locker will just use to the old behavior of returning `ErrUploadLocked` errors. `requestRelease` is a progressive enhancement that does not break compatibility. Wrapping a memorylocker around a filelocker seems not good to me in that context.

oliverpool:
> Do you want to use the support for requestRelease from the memorylocker to notify filelocker to release its lock?

No. I just want to use the `requestRelease` logic (provided by the `memorylocker`) when the upload is handled by another goroutine, while preventing multiple processes to write the same file (provided by `filelocker`).

---

We currently use a `filelocker` because we sometimes start a new version of the upload server (the ongoing uploads are still handled by the old process, but all new requests go to the new process). However the `filelocker` does not support the `requestRelease` at all (its a no-op).

Hence to benefit from the `requestRelease` mechanism I thought of using a `memorylocker`. However I still need to be sure, that no other process is writing the file: hence a mix of memory and file lockers:

Acquiring the lock would work like this:
1. lock in memory (`requestRelease` will work at some point, if another goroutine of the same process had the lock)
2. lock file (if it fails, release the memory lock and return the error: no inter-process communication is implemented for now)

Releasing the lock would then be in reversed order:
1. release the file lock
2. release the memory lock

=> If there is only one process (most of the time), the `requestRelease` will act (and the `filelocker` will always succeed, since acquiring the memory lock is a pre-requisite)
=> If there are two processes (when upgrading for instance), the `lock file` will fail (so same behavior as we have now) and we will release the memory lock before returning the error.

---

I thought of doing a `file+memory-locker` but it turns out that it is simpler to have a generic `DoubleLocker` implementation:
```go
doubleLock := DoubleLocker{
	First:  memorylocker.New(),
	Second: filelocker.New("path"),
}
lock := doubleLock.NewLock("123")
err := lock.Lock(ctx, requestRelease) // memory and then file lock
if err != nil {} // could not acquire the lock

_ = lock.Unlock() // file and then memory unlock
```

We have a "**opportunistic improvement**": if the lock was held in the same process (but another goroutine), the `requestRelease` will work. Otherwise we get a failure (current behavior).

PR #668, opened 2022-02-24
Add a flag to disable downloads of uploaded files

### LINKED ISSUE #223: Prevent Download
Is there a way to disable the download in the tusd binary, I've looked at all the cli flags and don't see anything in particular about how to disable the download serving. Looking through the code when you're using S3 it looks like it overwrites anything I could do to disable it as it just passes the GetReader on.

(I'm using Minio as the s3 storage)

omBratteng:
Closes #223

### DIFF
--- a/cmd/tusd/cli/flags.go
+++ b/cmd/tusd/cli/flags.go
@@ -22,6 +22,7 @@ var Flags struct {
 	Basepath                string
 	ShowGreeting            bool
 	DisableDownload         bool
+	DisableDelete           bool
 	Timeout                 int64
 	S3Bucket                string
 	S3ObjectPrefix          string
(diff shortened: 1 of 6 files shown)


Akkarine:
I'm also interested in this feature. Use-case: I'm loading files to S3 and serve them with CDN. So I don't want to load my backend with location link usage. Can anyone look on this PR? @tim-kos?

Akkarine (dismissed):
Merged to #685, checked, works.

Acconut:
Thank you for this PR! It looks good, but there are two things:

1. It's better to call the flag `DisableTermination` instead of `DisableDelete` because for tus we use the terminology "to terminate an upload".
2. Your PR removes the routes from the router, but I wonder if one can still access the DELETE/GET handlers by supplying a X-HTTP-Method-Override header: https://github.com/tus/tusd/blob/master/pkg/handler/unrouted_handler.go#L211 Can you test if this would circumvent the disabling of these endpoints?

omBratteng:
> 1. It's better to call the flag `DisableTermination` instead of `DisableDelete` because for tus we use the terminology "to terminate an upload".

Done

> 2. Your PR removes the routes from the router, but I wonder if one can still access the DELETE/GET handlers by supplying a X-HTTP-Method-Override header: [https://github.com/tus/tusd/blob/master/pkg/handler/unrouted_handler.go#L211](https://github.com/tus/tusd/blob/master/pkg/handler/unrouted_handler.go?rgh-link-date=2022-03-21T10%3A56%3A28Z#L211) Can you test if this would circumvent the disabling of these endpoints?

```console
❯ curl -X DELETE -H 'X-HTTP-Method-Override: DELETE' -H 'Tus-Resumable: 1.0.0' http://localhost:1080/files/d62477b5406a4d2c7725d0d8d498ec81
Method Not Allowed

❯ curl -X GET -H 'X-HTTP-Method-Override: DELETE' -H 'Tus-Resumable: 1.0.0' http://localhost:1080/files/d62477b5406a4d2c7725d0d8d498ec81
Method Not Allowed

❯ curl -X PATCH -H 'X-HTTP-Method-Override: GET' -H 'Tus-Resumable: 1.0.0' http://localhost:1080/files/d62477b5406a4d2c7725d0d8d498ec81
Method Not Allowed
```

Seems like you can't bypass it by using the `X-HTTP-Method-Override` header

Acconut (approved):
Thank you very much!

omBratteng:
Awesome @Acconut, do you think you could do a patch or minor release soon?

Acconut:
This is included in v1.9.0: https://github.com/tus/tusd/releases/tag/v1.9.0

PR #734, opened 2022-05-23
feat: allow user to change fileinfo in callbacks

halconel:
#goal allow user to change file info in callbacks
In my case, I use this in order to be able to independently generate `Upload.ID`
which allow me to change folder in s3 store

Example:
```
	handler, err := tusd.NewUnroutedHandler(tusd.Config{
		BasePath:                "api/v1/files/",
		StoreComposer:           composer,
		NotifyCreatedUploads:    true,
		NotifyUploadProgress:    true,
		NotifyTerminatedUploads: true,
		NotifyCompleteUploads:   true,
		PreUploadCreateCallback: defineFolders,
	})
	if err != nil {
		panic(fmt.Errorf("Unable to create handler: %s", err))
	}
```
```
// defines path and key id for new upload
// {some}/{path}/{uid}
func defineFolders(hook *tusd.CallbackEvent) error {
	folder := "folder_example"
	hook.Upload.ID = fmt.Sprintf("%s/%s", folder, uid())

	return nil
}
```

### DIFF
--- a/pkg/handler/config.go
+++ b/pkg/handler/config.go
@@ -50,11 +50,11 @@ type Config struct {
 	// property is supplied. If the callback returns nil, the upload will be created.
 	// Otherwise the HTTP request will be aborted. This can be used to implement
 	// validation of upload metadata etc.
-	PreUploadCreateCallback func(hook HookEvent) error
+	PreUploadCreateCallback func(hook *CallbackEvent) error
 	// PreFinishResponseCallback will be invoked after an upload is completed but before
 	// a response is returned to the client. Error responses from the callback will be passed
 	// back to the client. This can be used to implement post-processing validation.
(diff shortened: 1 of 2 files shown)


Acconut:
Thank you for this PR, but I don't think approach is the optimal. As laid out in https://github.com/tus/tusd/pull/685, we have plans to allow users to change the location of files. However, we want to implement this using the hook system that will be launched in tus v2.

Acconut:
@sdhull We unfortunately do not have any concrete dates or ETAs available. I know it is frustrating to see PRs like this being closed when no other option is available yet. However, we would like to implement such features in a proper manner, even if this takes time. I hope you understand this.

If you feel the need, to switch to one of these forks, feel free too. You can always switch back to the mainline tusd version. But please be aware that we are obviously not maintaining them :)

PR #739, opened 2022-06-03
Return ReadCloser in getReader

stefan-scheidewig:
After uploading big files to an azure storage, a call to getReader would load the complete file into memory leading to out-of-memory issues. This pr fixes that and unnecessary check for Close by just returning the ReadCloser calling getReader.


Acconut:
The two changes look sensible to me, unless @omBratteng has something else to say for the AzureStore. Otherwise, please have a look at the failing tests. If they pass, we can merge this.

stefan-scheidewig:
@Acconut somehow mockgen generates an invalid `handler_mock_test.go` for me. the datastore.go types are not correctly imported.

Acconut:
Fixed the mocks manually and merged in https://github.com/tus/tusd/commit/1038298a7999922f09a925e2f59edc813fa8deca. Thank you.

PR #672, opened 2022-03-01
v2: Next major version

Acconut:
Tasks to be done:

**General**
- [x] change go.mod to ../v2 and update internal references to .../v2 (see https://github.com/tus/tusd/pull/672#issuecomment-1082737523)
- [x] Switch to new structured logging package: https://pkg.go.dev/log/slog@master, https://thedevelopercafe.com/articles/logging-in-go-with-slog-a7bb489755c2
- [x] Pass context values down to storages and hooks while retaining context values (Closes https://github.com/tus/tusd/issues/713, https://github.com/tus/tusd/issues/443, https://github.com/tus/tusd/pull/455, https://github.com/tus/tusd/pull/315, https://github.com/tus/tusd/pull/740, https://github.com/tus/tusd/pull/342)
- [x] Look through TODOs in code
- [x] Look through lint errors in code

**Hooks**
- [x] docs: Update documentation about hooks
- [x] docs: Add plugin hooks to documentation
- [x] cli: Consider moving hooks in `pkg/`
- [x] cli: Add tests for hook implementations and system
- [x] handler: Allow modifying the HTTP response when stopping an upload
- [x] grpc: Add headers to hook request
- [x] grpc: Add documentation to proto files
- [x] grpc: Move proto file into better, public location
- [x] plugin: Stop zombie processes when tusd terminates
- [x] plugin: Move into public package. The plugin should not import for the cmd directory
- [x] plugin: Clean up hooks/plugin.go

**Locking**
- [x] Implement functionality for filelocker
- [x] ~~Adapt implementation at tusd-etcd-locker~~
- [x] Add tests for new functionality in handler
- [x] Increase default read timeout to reduce their likelihood
- [x] Add documentation

**S3**
- [x] Upgrade to AWS SDK v2
- [ ] Fix `file already closed` error (Close is already called by HTTP transport)

**Testing**
- [ ] Thorough testing of filestore and s3store under bad network conditions

**Release**
- [ ] Release notes (breaking changes, new features)
- [ ] Blog post
- [ ] Tweet

PR #740, opened 2022-06-03
Passing the context to sync callbacks

stefan-scheidewig:
In a scenario when tusd is used as library it is most probably integrated into an application having some kind of authentication context attached to the actual tus requests. To access that and to also cancel operations when the client request is cancelled it is necessary to use the request's context in tusd and to pass the context to the callback functions to access these information in the application's domain.


pkg/handler/context.go
```diff
@@ -19,8 +19,7 @@ type httpContext struct {
 
 func newContext(w http.ResponseWriter, r *http.Request) *httpContext {
 	return &httpContext{
-		// TODO: Try to reuse the request's context in the future
-		Context: context.Background(),
+		Context: r.Context(),
```
Acconut, line 22: This is not possible so easily as it breaks the S3Store (and possibly also gcsstore etc). The problem is that these stores run code using this context, which should also run even if the HTTP request has been aborted and thus the request context has been cancelled. The code in question is responsible for storing all of the data up to the moment that the request got stopped. And that is not possible if the context has been cancelled.

We haven't found a way so far to work around this problem but that is why the request context is not used for now. Let us know if you have any idea on how to solve this.

stefan-scheidewig:
But isn't that behavior wanted. If the client cancels a patch request the underlying store should drop it, too. So the upload offset of an upload would actually be the old offset before the cancelled operation.

Acconut:
> But isn't that behavior wanted. If the client cancels a patch request the underlying store should drop it, too. So the upload offset of an upload would actually be the old offset before the cancelled operation.

No that is not what we want. If the upload offset after a PATCH request is the same as before the PATCH request, then no data was stored and the data transfer was useless. If you then try to resume, then you have to start at the previous offset again. A tus server should always try to store as much received data as possible, regardless of whether the PATCH request is completed properly or not.

stefan-scheidewig:
But what about the last PATCH request. If the client cancels the request, it will not get the response anymore by the application's callback. Actually the upload is completed. Can the callback be called by the client if the upload offset pointer stands at the upload length?

Acconut:
> If the client cancels the request, it will not get the response anymore by the application's callback

That is correct. That is why the responses for the PATCH requests should not include any critical information that must not be lost. Such information should be relied through another, application-specific system, such as another RESTful endpoint.


> Can the callback be called by the client if the upload offset pointer stands at the upload length?

No, the callback is only invoked once, when the upload is completed. Never a second time.

stefan-scheidewig:
Made a change that passes at least the parent context's values but decouples the contexts again: https://github.com/tus/tusd/commit/ab78498b23645d284578c773c6f0473eed0018cd

I wanted to change the godoc of the synchronous callback methods in config.go but unsure how to phrase it correctly because it is contradictory if the callback can be used for validation but it is not guaranteed that the caller ever gets the result.

Acconut:
> it is contradictory if the callback can be used for validation but it is not guaranteed that the caller ever gets the result.

I don't understand that. Can you elaborate? Who is the caller? The tus client?

stefan-scheidewig:
> > it is contradictory if the callback can be used for validation but it is not guaranteed that the caller ever gets the result.
> 
> I don't understand that. Can you elaborate? Who is the caller? The tus client?

right.

Acconut:
Can you explain what the problem with the documentation is then?

stefan-scheidewig:
> Can you explain what the problem with the documentation is then?

the outcome of the validation is crucial to the uploading client. if the connection drops and the request is thus cancelled the client is forced to reupload the file and it will get the outcome after that second upload attempt because the callback is called just once.

Acconut:

stefan-scheidewig:
cool. thanks.

PR #741, opened 2022-06-03
Setting Upload-Offset directly to response

stefan-scheidewig:
The Upload-Offset header is a mandatory response header within the tus protocol. In a scenario using tusd as library by uploading and processing big files the client's request can be kept alive by writing to the ResponseWriter. Doing so would lead to an invalid response without the Upload-Offset header as the response was already written. For that reason it is crucial to directly write mandatory headers like this one to the response.

In general it would also be better for the using application to have a chance to directly write headers into the response in the synchronous callbacks.

### DIFF
--- a/pkg/handler/unrouted_handler.go
+++ b/pkg/handler/unrouted_handler.go
@@ -649,7 +649,7 @@ func (handler *UnroutedHandler) writeChunk(c *httpContext, resp HTTPResponse, up
 
 	// Send new offset to client
 	newOffset := offset + bytesWritten
-	resp.Headers["Upload-Offset"] = strconv.FormatInt(newOffset, 10)
+	c.res.Header().Set("Upload-Offset", strconv.FormatInt(newOffset, 10))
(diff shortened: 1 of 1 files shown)


Acconut:
I do not fully understand your situation. How exactly are you handling requests? Do you capture them before they are passed to tusd and then use them to send additional data to the client? tusd is not designed for such request handling outside of its logic.

Are you not able to use the PreFinishResponseCallback for processing files? The response will not be send until this callback returns.

stefan-scheidewig:
We are using the blocking PreFinishResponseCallback to process the uploaded file, which can take longer. As it may can take longer we prevent the closing of the client connection by sending empty bytes in a certain interval to keep the connection alive. But when you write to the response the currently set headers will be sent. They cannot be set after that. So the best would be to set headers directly to the response. Also for the application called by the callback.

Acconut:
https://github.com/tus/tusd/pull/455 implements a solution with provides the values from the context, but does not cause the context to get cancelled.

PR #746, opened 2022-06-10
fix: avoiding `writeChunk` no-op

andrestone:
Some implementations make a `PATCH` request to inform the server about the upload size after the last byte is sent:

```javascript
// tus-js-client/lib/upload.js
// ...
        // If the upload length is deferred, the upload size was not specified during
        // upload creation. So, if the file reader is done reading, we know the total
        // upload size and can tell the tus server.
        if (this.options.uploadLengthDeferred && done) {
          this._size = this._offset + (value && value.size ? value.size : 0)
          req.setHeader('Upload-Length', this._size)
        }

        if (value === null) {
          return this._sendRequest(req)
        }
        this._emitProgress(this._offset, this._size)
        return this._sendRequest(req, value)
// ...
```

This PR simply adds a check to `PatchFile` so that `writeChunk` isn't called if there are no bytes to write.

### DIFF
--- a/pkg/handler/unrouted_handler.go
+++ b/pkg/handler/unrouted_handler.go
@@ -577,12 +577,6 @@ func (handler *UnroutedHandler) PatchFile(w http.ResponseWriter, r *http.Request
 		info.Size = uploadLength
 		info.SizeIsDeferred = false
 
-		// Request was made simply to inform the server of the upload length, no need to call handler.writeChunk
-		if uploadLength == offset {
-			handler.sendResp(w, r, http.StatusNoContent)
-			return
-		}
-
 	}
 
 	if err := handler.writeChunk(ctx, upload, info, w, r); err != nil {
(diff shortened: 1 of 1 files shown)


Acconut:
Thank you for this PR, but `writeChunk` actually must be called after Upload-Length has been received for length-deferred uploads because it contains logic for calling `FinishUpload` and invoking pre-finish and post-finish hooks. You can also see this in the failing test case: https://github.com/tus/tusd/runs/6934303723?check_suite_focus=true

Is there a need for you to avoid calling writeChunk?

andrestone:
Sorry I missed the tests. I've changed it so it avoids calling `FileUpload`'s `writeChunk` (tests are passing).

> Is there a need for you to avoid calling writeChunk?
Yes, I watch the file system and work on the uploaded data on real time, the no-op call to the `os.OpenFile` was interfering in the process. This is not a blocker though, feel free to close without merging.

Thanks!

Acconut:
> Yes, I watch the file system and work on the uploaded data on real time

Any reasons why you decided to use this approach instead of using the hooks that tusd exposes to notify you about uploads?



> Sorry I missed the tests. I've changed it so it avoids calling `FileUpload`'s `writeChunk` instead (tests are passing).

Thanks for the quick change! An even easier approach might be to add the condition `&& maxSize > 0` to line https://github.com/tus/tusd/blob/63830b35d11ab0259f31ede6b7cd28645d41c714/pkg/handler/unrouted_handler.go#L626. Could you test that as well?

andrestone:
> Could you test that as well?

Done!

> Any reasons why you decided to use this approach instead of using the hooks that tusd exposes to notify you about uploads?

I need more granular control over the chunk sizes / frequency.

Acconut:
Thanks! Merged!

> I need more granular control over the chunk sizes / frequency.

If you have any idea for improvement or feature request, let us know :)

andrestone:
> Thanks! Merged!

Great! Thank you. 

> If you have any idea for improvement or feature request, let us know :)

Sure thing!

Acconut:
Unfortunately, this change caused a bug to appears, as reported in #798. We could work around this by adding an additional check for partial objects in the FinishUpload method of the S3Store. But as a quick fix, I will have to revert this PR.

PR #758, opened 2022-06-28
fix mimeInlineBrowserWhitelist typo

pescew:
fixed typo to allow video/ogg content-disposition inline.
fixed comment typos.

In an offline branch I also added 'video/mp4' to be served inline, but I know the MP4 container supports more video+audio codecs than browsers are capable of displaying.  Would it be possible to add whitelisted codecs?

such as:
'video/mp4; codecs="avc1.4d002a"'

https://developer.mozilla.org/en-US/docs/Web/Media/Formats/codecs_parameter#general_syntax

### DIFF
--- a/pkg/handler/unrouted_handler.go
+++ b/pkg/handler/unrouted_handler.go
@@ -782,10 +782,10 @@ func (handler *UnroutedHandler) GetFile(w http.ResponseWriter, r *http.Request) 
 
 // mimeInlineBrowserWhitelist is a map containing MIME types which should be
 // allowed to be rendered by browser inline, instead of being forced to be
-// downloadd. For example, HTML or SVG files are not allowed, since they may
+// downloaded. For example, HTML or SVG files are not allowed, since they may
 // contain malicious JavaScript. In a similiar fashion PDF is not on this list
 // as their parsers commonly contain vulnerabilities which can be exploited.
-// The values of this map does not convei any meaning and are therefore just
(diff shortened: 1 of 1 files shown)


Acconut (approved):
Thank you very much!

Acconut:
> In an offline branch I also added 'video/mp4' to be served inline, but I know the MP4 container supports more video+audio codecs than browsers are capable of displaying. Would it be possible to add whitelisted codecs?

We can add video/mp4 to the whitelist, but I am not very keen of whitelisting specific codecs. tusd is not the place to handle such stuff.

pescew:
Understood, thanks. You may want to hold-off on adding "video/mp4" by itself, because any codecs other than h264+AAC won't render in most browsers.

Another mime type I added in my offline branch is "image/avif".  It's for the new AV1 image format, which is surprisingly widely supported across major browsers for inline viewing already.

https://en.wikipedia.org/wiki/AVIF
https://caniuse.com/avif

Acconut:
The goal behind mimeInlineBrowserWhitelist is not to have a whitelist of which media types can be displayed in a browser or not. Instead, it is concerned about which files can be inlined without causing a security issue. For example, HTML must not be inlined or otherwise people can host arbitrary pages on tusd and compromise user's data. As such, every file format that is safe to be displayed inline can be added to this list, regardless of what the browser support looks like. It is up to the browsers to check whether they support the file format or not.

Issue #456, opened 2021-01-19
s3store: NoSuchKey error


lucas007007:
**Question**
Hi All
I try to use tusd to  upload large file( ~1G) 
After  100% completed  was shown in  browser console , http 500 error returned. (and upload failed)

```
[tusd] 2021/01/19 11:36:45 event="UploadCreated" id="98754d0a85dd5a81d3cb839c34c803ef+g_ZoA05ZqjmE87ULQPuXk4VgkjyO1di6oddh9IsPYlda8j4PYNfVPeH6u2_zExxsM2vnmUbpM4FVqJJXikfeahcW71RjDCTAlXj8I7zlRiU-" size="751566392" url="http://localhost:8005/api/v1/files/98754d0a85dd5a81d3cb839c34c803ef+g_ZoA05ZqjmE87ULQPuXk4VgkjyO1di6oddh9IsPYlda8j4PYNfVPeH6u2_zExxsM2vnmUbpM4FVqJJXikfeahcW71RjDCTAlXj8I7zlRiU-"
2021/01/19/ 11:36:45.108741537  [Info] vvv/pkg/upload.invokeHookSync 157 id:98754d0a85dd5a81d3cb839c34c803ef+g_ZoA05ZqjmE87ULQPuXk4VgkjyO1di6oddh9IsPYlda8j4PYNfVPeH6u2_zExxsM2vnmUbpM4FVqJJXikfeahcW71RjDCTAlXj8I7zlRiU-,size:%!s(int64=751566392),name:post-create
[tusd] 2021/01/19 11:36:45 event="ResponseOutgoing" status="500" method="POST" path="" error="Multiple errors occurred:
        NoSuchKey: The specified key does not exist.
        status code: 404, request id: BFA2563C1B12B816, host id: m1ORcd4vo91gwKa0nQUSIHdJ5YDMGQaz0Qk6nUpWkwMC5/HeZ8k+uDWQ9z23d12lIOQiBQTuQLQ=
        NoSuchKey: The specified key does not exist.
        status code: 404, request id: 856AEEF3BF64CEC4, host id: Nu3Kog6n1emh+jCxsGeD7rthWKCu+tg8Lhd+LanY98eGlrYvdFy1Z8hqFGySeUOOPV0HQomnb7E=
        NoSuchKey: The specified key does not exist.
[...]

PR #767, opened 2022-07-12
Add prefix to parallel upload concat source

RobertVillalba:
Fixes: #456 

Add the store prefix to the sources when calling concatUsingMultipart. This fixes the bug mentioned in the above PR which was being triggered whenever a parallel upload with parts > 5MB was being concatenated.


Acconut (approved):
Looks good! Thank you for the fix!

Issue #793, opened 2022-08-17
Is it possible to add timestamp with microseconds to tusd log?


ShunNaito:
**Question**
Is it possible to add timestamp with microseconds to tusd log?

I want change following line, like this.
https://github.com/tus/tusd/blob/69c567b4faa6ecb3518d123bd0725ef610e0231b/pkg/handler/config.go#L62
↓
```
config.Logger = log.New(os.Stdout, "[tusd] ", log.Ldate|log.Lmicroseconds)
```

**Setup details**
Please provide following details, if applicable to your situation:
- Operating System: Linux
- Used tusd version: v1.9.1
- Used tusd data storage: AWS S3
- Used tusd configuration: [-base-path, -hooks-dir, -port, -behind-proxy, -timeout, -s3-bucket]
- Used tus client library: tus-js-client

ShunNaito:
Can I send you a pull request?

Acconut:
Closing because it is implemented in #794.

PR #794, opened 2022-08-18
Change timestamp to microseconds in log

ShunNaito:
It is useful to correctly track the requests if log of events other than hooks are written timestamps with microseconds.


ShunNaito:
Closes #793

Acconut (approved):
Looks helpful, thanks!

Issue #799, opened 2022-08-24
[FEATURE REQUEST] Docker Secrets to read sensitive information from


renepardon:
**Is your feature request related to a problem? Please describe.**
I would like to use Docker Secrets to store my MINIO/AWS credentials. Therefore I created a Docker Secret at **/run/secrets/aws-access-key** and **/run/secrets/aws-key-id**
But there are no environment variables available as for other Docker images to load those secretes from corresponding file. E.g. https://hub.docker.com/r/bitnami/redis has **REDIS_PASSWORD_FILE** in addition to **REDIS_PASSWORD** to load this Docker Secret.

**Describe the solution you'd like**
I would like to have the environment variable **AWS_ACCESS_KEY_ID_FILE** and **AWS_SECRET_ACCESS_KEY_FILE** available to load the secret from Docker rather than from environment variables. Which makes it more secure against accidental exposure.

**Can you provide help with implementing this feature?**
If you tell me where and how to implement it, I would be happy to provide a pull request.

Acconut:
Interesting approach. I have never heard of this before.

> E.g. https://hub.docker.com/r/bitnami/redis has **REDIS_PASSWORD_FILE** in addition to **REDIS_PASSWORD** to load this Docker Secret.

Do you know how this is implemented? Is this a feature in the Redis binary or a feature of the used Docker image (i.e. the corresponding file is read before starting Redis)?

renepardon:
@Acconut i saw this already in multiple images like rabbitmq, MariaDB etc.

More details about Docker secrets: https://docs.docker.com/engine/swarm/secrets/

More example images using this approach already:
- https://hub.docker.com/_/rabbitmq
- https://hub.docker.com/_/mongo

And lot more. 

It’s part of the init script to read the secret and pass it to the correct destination.

Acconut:
Ok. Then feel free to open a PR, if you want to. We don't use docker, so I cannot work on this on my own.

renepardon:
@Acconut done :)

PR #810, opened 2022-09-12
Docker: add docker secret env parsing

renepardon:
This pull request resolves the requested feature #799 to read mounted Docker Secrets from /run/secrets folder into corresponding ENV vars. The ENV vars are defined inside **docker/load-env.sh** script. If a _FILE suffix is added to an ENV var, the script tries to read the corresponding file and sets the environment variable without _FILE suffix to the same value.

grafik
grafik

### DIFF
--- a/examples/docker-compose.yml
+++ b/examples/docker-compose.yml
@@ -1,0 +1,29 @@
+version: "3.9"
+services:
+    tusd:
+      image: tusproject/tusd:custom # tusproject/tusd:v1.9
+      command: -verbose -s3-bucket mybucket -s3-endpoint http://minio:9000
+      volumes:
+        - tusd:/data
+      environment:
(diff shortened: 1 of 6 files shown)


examples/docker-compose.yml
```diff
@@ -0,0 +1,29 @@
+version: "3.9"
+services:
+    tusd:
+      image: tusproject/tusd:v1.9
+      command: -verbose -s3-bucket mybucket -s3-endpoint http://minio:9000
+      volumes:
+        - tusd:/data
+      environment:
+        - AWS_REGION=us-east-1
+        - AWS_URL=http://minio:9000
```
Acconut, line 10: tusd does not use the AWS_URL environment variable. Is it used by some other component?

Acconut (approved):
Thank you for this PR!

PR #820, opened 2022-09-28
filestore: keep ID passed to NewUpload, if available

genofire:
`info.ID` could be set external e.g. by `config.PreUploadCreateCallback`

is already set in
- s3store (complicated)
- gcsstore (get from there)
- azurestore (get from there)


Acconut:
In theory, I am open to merge this PR, but the CI is failing. Please have a look at the build failures.

genofire:
@Acconut of course - i have overseen an "bug" in the code - here you get it

Acconut (approved):
Thank you!

PR #884, opened 2023-01-19
Recommend cloning using HTTPS

zain-naveed:


Acconut:
Why was it necessary to change the clone command for you?

Your version does not work for me:

```
$ git clone git@github.com/tus/tusd.git
fatal: repository 'git@github.com/tus/tusd.git' does not exist
```

zain-naveed:
i have to try to clone this repo but it not working yet then I try to clone this repo with this command

git clone https://github.com/tus/tusd.git

Acconut:
Could you explain your rationale behind this change? Is HTTPS better than SSH?

zain-naveed:
so for SSH, we need to generate the keys because most of the developers do
not know about it SSH, so HTTPS is recommended.

Acconut:
Makes sense. Some developers might not have set up cloning over SSH.

Issue #897, opened 2023-02-06
Is there any documentation related how to configure tusd server with azure blob storage?


Nippucsharptek:
What is the process of configuring Tusd server with azure blob storage, i have created a storage account and a container inside it what else do i need to do to configure tusd server with the container.

**Setup details**
Please provide following details, if applicable to your situation:
- Operating System: [e.g. Linux, Windows, macOS]
- Used tusd version: [can be obtained by running `tusd -version`]
- Used tusd data storage: Azure]
- Used tusd configuration: [what arguments are you using to invoke `tusd`?]
- Used tus client library: [e.g. tus-js-client]

Acconut:
You are correct, the docs at https://github.com/tus/tusd/blob/main/docs/usage-binary.md do not yet mention Azure. The information is however available thorugh `tusd -help` or in the source code for the flags: https://github.com/tus/tusd/blob/aa2fc1ea6e4a987e28932d3f8032fab9d3556871/cmd/tusd/cli/flags.go#L84-L88

@omBratteng Would you have time to add a brief section about using tusd with Azure to the above-mentioned docs?

omBratteng:
See #906

Acconut:
Closed in #906, thanks to @omBratteng :tada:

PR #906, opened 2023-02-13
Add usage docs for using azure storage backend

omBratteng:
Adds documentation for using the Azure Blob Storage backend, as requested in issue #897


Acconut (approved):
Thank you very much!

PR #936, opened 2023-04-11
fix(docker): correct multi platform build

### LINKED ISSUE #922: arm64 container fails with "Exec format error"
**Describe the bug**
When trying to run `ghcr.io/tus/tusd:v1.10.1` on an ARM Amphere system, the container fails to start and displays the error message:
```
root@oracle-arm01:/home/justin# docker run -it --rm ghcr.io/tus/tusd:v1.10.1
/usr/local/share/docker-entrypoint.sh: line 9: /usr/local/bin/tusd: cannot execute binary file: Exec format error
/usr/local/share/docker-entrypoint.sh: line 9: /usr/local/bin/tusd: No error information
```

**To Reproduce**
On an arm64 system with docker or an equivalent container runtime, execute `docker run -it --rm ghcr.io/tus/tusd:v1.10.1`

**Expected behavior**
Running this command would start tusd.

**Setup details**
Please provide following details, if applicable to your situation:
- Operating System: Linux 
- Used tusd version: v1.10.1
- Used tusd data storage: N/A
- Used tusd configuration: No configuration
- Used tus client library: N/A

I suspect specifying the GOARCH in the Dockerfile is the root cause of the issue: https://github.com/tus/tusd/blob/main/Dockerfile#L23

The github action here should do the right thing with no GOARCH in the Dockerfile: https://github.com/tus/tusd/blob/main/.github/workflows/release.yaml#L71

dargmuesli:
Resolves #922 

cc @PatrickMi @jrcichra for testing

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.20.2-alpine AS builder
+FROM --platform=$BUILDPLATFORM golang:1.20.2-alpine AS builder
 WORKDIR /go/src/github.com/tus/tusd
 
 # Add gcc and libc-dev early so it is cached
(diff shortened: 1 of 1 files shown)


Acconut:
Thank you for the PR. Let's merge it, so others can easily test if the produced image is usable on ARM :)

Acconut (approved):

PR #945, opened 2023-05-26
fix typo

abetomo:
Terminater ->
Terminator


Acconut:
Thank you for this PR, but the name `Terminater` was chosen on purpose to follow the naming conventions for Go interfaces: https://go.dev/doc/effective_go#interface-names Of course, could also have gone with `Terminator` at that time, but we now have this interface name and I think it is fine to keep it and avoid unneeded breaking changes.

Nevertheless, thank you very much for your effort here :)

abetomo:
Thanks for the review!
Understood.

PR #963, opened 2023-06-12
Graceful shutdown

### LINKED ISSUE #395: Gracefully terminate on SIGTERM

Acconut:
This PR implements a graceful shutdown for tusd. Once it receives a shutdown instruction, it will stop listening and try to close down all uploads properly, so that data stores and locker can properly clean up. It waits for a specific period after the first shutdown instructions. If the requests are not all closed down at this point, it will exit regardless.

Closes #395 

Tasks:
- [x] Consider doing this for SIGTERM on Linux as well
- [x] Make wait period configurable
- [x] Allow to disable this behavior
- [x] Test this on Windows
- [x] Documentation

### DIFF
--- a/cmd/tusd/cli/serve.go
+++ b/cmd/tusd/cli/serve.go
@@ -1,9 +1,13 @@
 package cli
 
 import (
+	"context"
 	"crypto/tls"
+	"errors"
 	"net"
 	"net/http"
+	"os"
+	"os/signal"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


cmd/tusd/cli/serve.go
```diff
@@ -115,62 +119,113 @@ func Serve() {
...
+		stdout.Println("Received interrupt signal. Shutting down tusd...")
+
+		// Wait for second interrupt signal, while also shutting down the existing server
+		go func() {
+			<-c
+			stdout.Println("Received second interrupt signal. Exiting immediately!")
+			os.Exit(1)
+		}()
+
+		// Shutdown the server, but with a user-specified timeout
```
craigpastro, line 213: I know this is still a draft, but just in case perhaps it can be a nice reminder: the timeout below is not user-specified.

Acconut: Thank you! You are right and this is also in the list of tasks in my comment above :)

Acconut:
@kvz Maybe you can help us with one decision here. Basically this PR adds a functionality to tusd that a graceful shutdown is started if it receives a SIGINT. Graceful shutdown means that it will wait a few seconds for all requests to be shut down properly before exiting. The shutdown will also be performed immediately if a second SIGINT is received. So far, so good.

Now I am wondering if this graceful shutdown should also be performed when a SIGTERM is received. AFAIK SIGINT and SIGTERM are both intended to be a kind request for the program to shut down. SIGKILL would still stop tusd immediately.

What do you think? Should SIGTERM also start this graceful shutdown? Could there be any issues with process monitors like systemd when tusd does not exit immediately after SIGINT/SIGTERM?

Acconut:
Thank you very much, Kevin! Your insights helped a lot and tusd should then behave as you described.

PR #962, opened 2023-06-09
hooks: Allow pre-create to change upload ID, meta data and storage details

Acconut:
This PR adds a method for v2 to overwrite the upload ID and meta data from the pre-create hook. No other hooks can perform these changes later on and only a selected set of properties can be overwritten. We also add the ability to pass storage details to data stores to influence file names, for example, although no data store implements this yet.

Context and previous work, which this PR is based on:
- https://github.com/tus/tusd/pull/814
- https://github.com/tus/tusd/pull/916
- https://github.com/tus/tusd/issues/811

PR #977, opened 2023-07-05
Enhance File Path Customization

MoritzGruber:
Hello, tusd maintainers and community,

In our organization, we employ a tenant structure akin to GitHub or GitLab, where we have organizations containing multiple projects. Each project can upload assets using tusd.

To efficiently manage permissions and quotas on files, we would like to incorporate the `project_id` into the file path of each asset. For example, the desired format is `assets/<project_id>/<asset_id>/filename.png`.

In order to achieve this, we needed to modify the file ID within `fileInfo` beforehand. Furthermore, we had to override the `ExtractIDFromPath` to align with our use case.

We have initially implemented this in a fork. However, we believe that integrating this enhancement into the core library would be beneficial for other users who might have similar use cases requiring flexible path customization.

## Proposed Changes:

1. Allow passing a custom function for `ExtractIDFromPath` in the configuration, which can be used to customize how file IDs are extracted from URLs.

2. Update references to `ExtractIDFromPath` to use the custom function if provided, else default to the existing implementation.

These changes are backward compatible and allow for greater flexibility in accommodating various file path structures.

We kindly ask for your review and consideration in integrating these changes into the core library.

## Special Thanks:

Big shoutout to the tusd maintainers and contributors! You guys rock for keeping this library on point. It's great that we can just plug in your library and rely on it working rock solid. Keep up the awesome work! 👏

Best regards,
Moritz

## Summary of Code Changes:

1. `pkg/handler/config.go`: Added `ExtractIDFromPath` to the Config struct.
2. `pkg/handler/unrouted_handler.go`: Various changes to reference `ExtractIDFromPath` from Config and to pass file information by reference.


Acconut:
Thank you for your efforts on this and the kind words! We have received requests about more detailed customization like this for a long time and finally managed to implement it a few weeks ago: https://github.com/tus/tusd/pull/962 It has not been (properly) released yet because it is still part of the ongoing development on tusd v2. There are pre-releases available and I would recommend you to check them out. Documentation is still lacking, but the examples from the PR might help you enough already.

Given that we already have implemented something similar in v2, I sadly don't think we can accept this PR. However, please provide any feedback for our implementation in v2, so can we improve it!

MoritzGruber:
Thanks for sharing. Makes sense!. We will check v2, thanks much!

Issue #1017, opened 2023-10-10
Close unfinished file


mskayali:
Hi, Im working on a device and I get binary data by a stream and I wouldn't know the total stream size, I just able to understand while stream is done.

Is it possible to add a protocol feature to close specific unfinished files by a request? That would be better if I able to give first post request header as "uploadLengthHeader" : "infinit" to finish file by a featured request.

mskayali:
I solved this, please take a look if you still read this https://github.com/tus/tusd/pull/1018

Acconut:
The tus protocol already supports this using the `Upload-Defer-Length` header. It works basically as you implemented it: The initial POST request includes `Upload-Defer-Length: 1` to indicate the absence of a finite length and a future PATCH request can then send the actual length via `Upload-Length`. tusd also implements this extension by default.

PR #1018, opened 2023-10-13
Close unfinished file and set max file size

mskayali:
If you get binary data by a stream and you wouldn't know the total stream size but you just able to understand this is the end of file when stream is done. You need to finalize the file.

This kind of IoT issues can be solved with this commit 
first set file-length as "max" to maximize the file length 
then upload stream bytes
finally send "Finish-Incomplate-File" header as "?1" by a head request


Acconut:
Thank you for the PR, but the tus protocol and tusd already support a very similar approach for handling unknown file sizes. I described this in my comment in #1017

Issue #1010, opened 2023-09-28
S3 headIncompletePartForUpload function error exclusions seem insufficient


jimydavis:
In `s3store.go`, there's a number of errors that are ignored as below. I presume this is for fresh new uploads that have no incomplete parts:

```go
func (store S3Store) headIncompletePartForUpload(ctx context.Context, uploadId string) (int64, error) {

...

  if err != nil {
    if isAwsError[*types.NoSuchKey](err) || isAwsError[*types.NotFound](err) || isAwsErrorCode(err, "AccessDenied") {
      err = nil
    }
    return 0, err
  }
```

However, yesterday on testing a fresh upload (no parts uploaded yet, completely fresh), I received 

`status=500 body="ERR_INTERNAL_SERVER_ERROR: operation error S3: HeadObject, https response error StatusCode: 403, RequestID: xxxxxxx, HostID: uploadID+multipartID, api error Forbidden: Forbidden\n"`

In my debugger, the error looked like this:

![image](https://github.com/tus/tusd/assets/16463756/e3701779-2e69-4f6f-baf7-ec3d334fdb65)

Do we have to exclude `Forbidden` for fresh uploads without incomplete parts too or am I missing something?

Upon reading the AWS error codes documentation, `Forbidden` is not an error code. Only `AccessDenied` is. But as you can see from my debugger screenshot, I am indeed getting the Forbidden code from what presumably must be from AWS's servers and not within any of the client code.

Lastly, many people should encounter this error as incomplete parts are not always going to be there and when you do a HEAD on it, its going to come back empty.

Thank you.

edit: this seems to affect `getIncompletePartForUpload` too.

Acconut:
Do you have allowed all of these actions?

https://github.com/tus/tusd/blob/5e9be5a18d3672b12d8fe7fd80bc26baae582472/pkg/s3store/s3store.go#L9-L13

I don't know the exact details but depending on the permissions AWS S3 either returns a 404 Not Found or 403 Forbidden error for non-existing objects. It could also be that we are missing a permission in the above documentation.

Acconut:
> In fact after extensive testing, I have no idea when and when a `.part` file will appear. Sometimes an actual `.part` file appears in the S3 console browser.

The `.part` files are sometimes created when an upload is paused in the middle of the transfer. S3 has a minimum part size for its multipart uploads and if the amount of remaining data is smaller than this minimum part size, we store it in a `.part` file, which will be consumed when the upload is resumed.


[...]

PR #1019, opened 2023-10-24
Add additional s3 error to ignore when file doesn't exist.

jimydavis:
Referring to this issue: https://github.com/tus/tusd/issues/1010


Acconut (approved):
Thank you!

PR #1026, opened 2023-11-04
feat: Allow users to set json log output

NoUseFreak:
Add `-format=json` options to allow logging in json format.

```
{"time":"2023-11-04T11:17:26.702372+01:00","level":"INFO","msg":"Using '/workdir/tusd/data' as directory storage."}
{"time":"2023-11-04T11:17:26.702557+01:00","level":"INFO","msg":"Using 0.00MB as maximum size."}
{"time":"2023-11-04T11:17:26.702594+01:00","level":"INFO","msg":"Supported tus extensions: creation,creation-with-upload,termination,concatenation,creation-defer-length"}
{"time":"2023-11-04T11:17:26.7026+01:00","level":"INFO","msg":"Using 0.0.0.0:8080 as address to listen."}
{"time":"2023-11-04T11:17:26.702603+01:00","level":"INFO","msg":"Using /files/ as the base path."}
{"time":"2023-11-04T11:17:26.702631+01:00","level":"INFO","msg":"Using /metrics as the metrics path."}
{"time":"2023-11-04T11:17:26.702947+01:00","level":"INFO","msg":"You can now upload files to: http://[::]:8080/files/"}
```

### DIFF
--- a/cmd/tusd/cli/flags.go
+++ b/cmd/tusd/cli/flags.go
@@ -65,6 +65,7 @@ var Flags struct {
 	PprofMutexProfileRate            int
 	BehindProxy                      bool
 	VerboseOutput                    bool
+	LogFormat                        string
 	S3TransferAcceleration           bool
 	TLSCertFile                      string
 	TLSKeyFile                       string
@@ -178,6 +179,7 @@ func ParseFlags() {
 		f.BoolVar(&Flags.ShowGreeting, "show-greeting", true, "Show the greeting message")
 		f.BoolVar(&Flags.ShowVersion, "version", false, "Print tusd version information")
 		f.BoolVar(&Flags.VerboseOutput, "verbose", true, "Enable verbose logging output")
+		f.StringVar(&Flags.LogFormat, "log-format", "text", "Logging format (txt or json)")
 	})
 
 	fs.AddGroup("Timeout options", func(f *flag.FlagSet) {
--- a/cmd/tusd/cli/log.go
+++ b/cmd/tusd/cli/log.go
@@ -16,22 +16,38 @@ func SetupStructuredLogger() {
 		level = slog.LevelDebug
 	}
 
-	slogger := slog.New(
-		slog.NewTextHandler(logWriter{stdout}, &slog.HandlerOptions{
+	replaceAttrFunc := func(groups []string, a slog.Attr) slog.Attr {
+		// Remove time attribute, because that is handled by the logger
+		if a.Key == slog.TimeKey {
+			return slog.Attr{}
+		}
+		// Rename `msg` to `event`
+		if a.Key == slog.MessageKey {
+			a.Key = "event"
+		}
+		return a
+	}
+
+	var handler slog.Handler
+	switch Flags.LogFormat {
+	case "json":
+		handler = slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{
 			Level: level,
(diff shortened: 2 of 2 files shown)


cmd/tusd/cli/flags.go
```diff
@@ -178,6 +179,7 @@ func ParseFlags() {
 		f.BoolVar(&Flags.ShowGreeting, "show-greeting", true, "Show the greeting message")
 		f.BoolVar(&Flags.ShowVersion, "version", false, "Print tusd version information")
 		f.BoolVar(&Flags.VerboseOutput, "verbose", true, "Enable verbose logging output")
+		f.StringVar(&Flags.OutputFormat, "format", "txt", "Logging format (txt or json)")
```
Acconut, line 182: Can we call the flag `log-format` with a default of `text`?

cmd/tusd/cli/log.go
```diff
@@ -16,22 +16,37 @@ func SetupStructuredLogger() {
...
+		if a.Key == slog.MessageKey {
+			a.Key = "event"
+		}
+		return a
+	}
+
+	var handler slog.Handler
+	switch Flags.OutputFormat {
+	case "json":
+		handler = slog.NewJSONHandler(logWriter{log.New(os.Stdout, "", 0)}, &slog.HandlerOptions{
```
Acconut, line 34:
```suggestion
		handler = slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{
```

I don't think we need `logWriter` here. That is just used for the text format to get the desired behavior we want. Or does that not work then?

cmd/tusd/cli/log.go
```diff
@@ -16,22 +16,37 @@ func SetupStructuredLogger() {
...
+		handler = slog.NewTextHandler(logWriter{log.New(os.Stdout, "", log.LstdFlags|log.Lmicroseconds)}, &slog.HandlerOptions{
+			Level:       level,
+			ReplaceAttr: replaceAttrFunc,
+		})
+	}
+
+	slog.SetDefault(slog.New(handler))
+
+	if Flags.OutputFormat == "json" {
+		stdout = log.Default()
```
Acconut, line 47: Could you explain these lines?

Acconut (changes requested):
Thank you for the PR!

cmd/tusd/cli/log.go
NoUseFreak: the argument needs to be an io.Writer instance.

cmd/tusd/cli/log.go
NoUseFreak: Comment added, these are used in the cli package.

NoUseFreak: If I would not do this, non of the output of the cli package would be in json. The goal was to ensure all output is json.

cmd/tusd/cli/log.go
Acconut: `os.Stdout` is of type `*os.File` which implements `io.Writer`, doesn't it?

NoUseFreak: Correct, fixed

Acconut:
Thank you for the PR! I has been merged in https://github.com/tus/tusd/commit/9b6ada97f0662100be607f48c25effd84150a718 🎉 I apologize that your authorship of this commit got lost when I added a commit locally and squashed them together.

PR #1042, opened 2023-12-05
s3store fix incomplete part keys

butonic:
When using a different prefix for metadata files uploads break because the code sometimes uses `metadataKeyWithPrefix` when it should use `keyWithPrefix`.

Unless I understood something completely wrong and failed to understand what the debug session was trying to tell me ;-)


Acconut:
Thank you for this PR, but I am not too sure about this change. The documentation clearly states that the metadata prefix will be used for the `.part` object: https://github.com/tus/tusd/blob/c856f723619a0a388c619d68357b9510c8632640/pkg/s3store/s3store.go#L107-L113

IIRC, the idea is that the object prefix defines the prefix for the completed file uploads. All additional objects (e.g. .info and .part) will be put under the metadata prefix.

Acconut:
Closing this issue due to inactivity. Feel free to leave a comment if you want to continue the discussion :)

Issue #432, opened 2020-10-08
Byte-range support for iOS compatibility


beevelop:
There were already a few discussions about supporting Byte-range (content-range / partial-range) a few years ago. For now, `tus` and `tusd` don't implement support for it, which is understandable in the respective context:
- https://github.com/tus/tus-resumable-upload-protocol/issues/2: "Range" vs. "Content-Range"
- https://github.com/tus/tus-resumable-upload-protocol/issues/116: Upload-Offset vs Range?
- https://github.com/tus/tusd/issues/296: Upload to tusd behind HTTPS reverse proxy disconnects with EOF

The decision to not support Byte-range leads to another nasty side-effect on our "favourite" mobile operating system _iOS_: **Media files (audio / video) stored on `tus` can not be played on iOS!**

> HTTP servers hosting media files for iOS must support byte-range requests, which iOS uses to perform random access in media playback. (Byte-range support is also known as content-range or partial-range support.) Most, but not all, HTTP 1.1 servers already support byte-range requests.

This renders `tus` unusable for e.g. storing video / audio notes through a mobile app and enabling the user to play the respective files **without** downloading them in advance.

Acconut:
I can understand that this is an inconvenience for you but tusd is not intended to serve content. Just as the tus protocol itself, tusd was built for handling file uploads and not file downloads. This decision helps us keep tusd maintainable and focused on one specific thing (which are file uploads in our case). For handling file downloads there are already other tools which take care of this job a lot better than we could. If you save files on disk, we recommend using nginx or some other proxy. If you use the S3Store or GCSStore, you can download the content from the bucket directly. In all of those cases, the byte ranges should be supported.

RubenGarcia:
You can also support byte-ranges in your GET endpoint, which tus does not document and you are thus free to implement your required functionality. IOS players need not even be aware of the POST / PATCH / etc endpoints tus creates.

Acconut:
Yes, you are correct, @RubenGarcia. The tus protocol does not prevent the support of byte range requests. However, supporting this feature is not in line with the goals that we have/had for tusd. We want tusd to focus on uploads and not on downloads, which it is not optimized for.

RubenGarcia:
That's what I'm saying. 
I meant beevelop can support byte-ranges in non-tus mandated endpoints.
I did not mean for the tus protocol to be modified.
[...]

PR #1048, opened 2023-12-14
handler: support range requests for GET /upload

markspolakovs:
Where the store supports it, use [`http.ServeContent`](https://pkg.go.dev/net/http#ServeContent) to serve the contents of the upload, thereby adding support for range requests:

> The main benefit of ServeContent over io.Copy is that it handles Range requests properly, sets the MIME type, and handles If-Match, If-Unmodified-Since, If-None-Match, If-Modified-Since, and If-Range requests.

This helps optimise downloading large files directly from tusd.

### DIFF
--- a/pkg/handler/unrouted_handler.go
+++ b/pkg/handler/unrouted_handler.go
@@ -646,6 +646,7 @@ func (handler *UnroutedHandler) HeadFile(w http.ResponseWriter, r *http.Request)
 		} else {
 			resp.Header["Upload-Length"] = strconv.FormatInt(info.Size, 10)
 			resp.Header["Content-Length"] = strconv.FormatInt(info.Size, 10)
+			resp.Header["Accept-Ranges"] = "bytes"
 		}
 
 		resp.StatusCode = http.StatusOK
@@ -999,6 +1000,13 @@ func (handler *UnroutedHandler) GetFile(w http.ResponseWriter, r *http.Request) 
 	src, err := upload.GetReader(c)
 	if err != nil {
 		handler.sendError(c, err)
+		return
+	}
+
+	if seeker, ok := src.(io.ReadSeeker); ok {
+		handler.sendResp(c, resp)
+		http.ServeContent(w, r, info.ID, time.Time{}, seeker)
+		_ = src.Close()
 		return
 	}
 
--- a/pkg/handler/get_test.go
+++ b/pkg/handler/get_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/golang/mock/gomock"
+
 	. "github.com/tus/tusd/v2/pkg/handler"
 )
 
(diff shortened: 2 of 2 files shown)


Acconut:
Thank you for this PR! Do you know which stores return an `io.ReadSeeker`? It's probably only the file store.

markspolakovs:
> Do you know which stores return an io.ReadSeeker? It's probably only the file store.

From a cursory look that appears to be the case - though I imagine that in theory we could wrap e.g. s3store with a ReadSeeker that uses range requests. For now though this would only work for the file store, the Range header would be ignored for all the others.

pkg/handler/unrouted_handler.go
```diff
@@ -646,6 +646,7 @@ func (handler *UnroutedHandler) HeadFile(w http.ResponseWriter, r *http.Request)
 		} else {
 			resp.Header["Upload-Length"] = strconv.FormatInt(info.Size, 10)
 			resp.Header["Content-Length"] = strconv.FormatInt(info.Size, 10)
+			resp.Header["Accept-Ranges"] = "bytes"
```
markspolakovs, line 649: We may need to check if the underlying store returns a ReadSeeker before advertising this

Acconut:
> though I imagine that in theory we could wrap e.g. s3store with a ReadSeeker that uses range requests.

Yes, that would be possible, but likely not very efficient. From briefly looking through the code behind `http.ServeContent`, it's apparent that the function performs multiple seek operations to determine the file size, content type, and reading the requested range. Each of these seek operations would result in one range request to S3. So a single range request to tusd can trigger multiple requests to S3.

A better approach would be to pass the range request directly to the datastore, so it can fetch the requested range from the underlying storage provider. In tusd, we enabled storages to implement such optional features through additional interfaces. For example, a storage can implement the `TerminaterDataStore` to provide cancellable uploads:

https://github.com/tus/tusd/blob/e2784199a268bd03cdc188f376386fadc6735582/pkg/handler/datastore.go#L121-L132

We could define a similar interface for fetching ranges. For the s3store, the implementation would then send a range request to S3. For the filestore, it could just seek to the desired position and read from there. The handler can then also detect whether the storage supports range requests at all and set the `Accept-Ranges` header accordingly.

This approach would require more effort to get it working for your use case, but also allows this feature to be provided by other storages in the future. Maybe the interface could even be extended to also allow handling the `If-*` headers.

What do you think?

markspolakovs:
I like that approach - it's a lot more effort, but it'd mean that we can extend the support to the other stores (I didn't realise that that's how ServeContent operates...).

Out of interest, I did some searching and found #432 where you were against implementing `Range` support in tusd - have you changed your mind on this? For context, my use case for adding this was to optimise another internal service downloading files from tusd - in the past when it failed it'd need to restart from the beginning. (I've worked around this by having this service read directly from minio, but it does mean I lose the flexibility of moving between backing stores and keeping tusd as a "generic" API).

In any case, we should probably move this conversation to an issue to sketch out API design if you're in favour of this in principle - closing this PR for now.

Acconut:
> Out of interest, I did some searching and found #432 where you were against implementing `Range` support in tusd - have you changed your mind on this?

Yes, I did change my opinion here. We have recently been involved in the HTTP working group with the goal of making a IETF standard for resumable uploads. In the end, we hope that resumable uploads are one component in the HTTP toolbox that interacts well with the existing HTTP mechanism. Since then I am more interested in making sure that tusd works better with concepts, such as range requests.

> In any case, we should probably move this conversation to an issue to sketch out API design if you're in favour of this in principle - closing this PR for now.

Great idea, feel free to open an issue if you are still interested in working on this :)

PR #1050, opened 2023-12-18
docs: Unify style of CLI arguments

martnst:
Update CLI option examples from 2 leading dashes to just 1.

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -175,10 +175,10 @@ With file hooks enabled, tusd will execute scripts or other executable files in 
 
 #### Hook Directory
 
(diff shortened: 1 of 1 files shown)


Acconut:
Thank you for the PR, but what is your motivation behind this? Go's flag parser supports options being supplied with two and just one dash. So the examples are correct AFAIK.

martnst:
Ah okay. I don't know much about go yet. I had noticed that `tusd -h` lists everything with a single dash. Thus, I thought the docs might be outdated. 

So with that, now I think for sake of consistency it would be nice to align it.

Acconut:
Thanks for this! I just saw that we also use single dashes in the majority of our documentation, so I agree that we should keep it consistent :)

Issue #1031, opened 2023-11-07
High memory usage when uploading to Azure Storage


jspath-ankored:
**Describe the bug**
I see very high memory usage when uploading large files to Azure Storage.  For example, a 221MB file upload results in using over 221MB of memory.  I see this behavior when uploading to actual Azure Storage as well as uploading to Azurite locally.

It seems like something in the tusd Azure Storage code may be retaining all of the file info in memory, instead of streaming it to Azure Storage or an Azure Storage emulator, like Azurite.

**To Reproduce**
1. Setup tusd to upload to Azure Storage
2. Ensure you have a way to monitor memory on the tusd server
3. Upload a large file >= 221MB
4. Look at the memory usage after

**Expected behavior**
I would expect the memory usage to stay relatively low.

**Setup details**
Please provide following details, if applicable to your situation:
- Operating System: Running tusd docker image `tusproject/tusd:v2.0`
- Used tusd version: 2.0
- Used tusd data storage: Azure Storage
- Used tusd configuration:
  - `--azure-storage uploads --azure-endpoint https://AZURE-ACCOUNT.blob.core.windows.net --hooks-http MYWEBHOOKURL --hooks-http-forward-headers Authorization --hooks-enabled-events pre-create,pre-finish --cors-expose-headers X-Upload-Properties-Set,X-Upload-File-Path --behind-proxy --expose-pprof`
    - I use `--behind-proxy` on Azure only, not locally. But I see the problem both locally and on Azure.
- Used tus client library: Uppy.js

jspath-ankored:
here's a heap dump of me running it locally:

[heap.txt](https://github.com/tus/tusd/files/13301188/heap.txt)

Acconut:
From https://community.transloadit.com/t/tusd-failing-on-large-file-upload/16883/11

Acconut:
From the heap dump, it is apparent that there is a 128MB and a 64MB allocated at

```
#	0x1007dd07b	bytes.growSlice+0x8b									/usr/local/go/src/bytes/buffer.go:249
#	0x1007dca9b	bytes.(*Buffer).grow+0x12b								/usr/local/go/src/bytes/buffer.go:151
#	0x1007dcea3	bytes.(*Buffer).ReadFrom+0x43								/usr/local/go/src/bytes/buffer.go:209
#	0x100912e57	bufio.(*Reader).WriteTo+0xb7								/usr/local/go/src/bufio/bufio.go:533
[...]

PR #1070, opened 2024-01-23
azurestore: Buffer upload data on disk instead of in-memory

Acconut:
Closes https://github.com/tus/tusd/issues/1031

Instructions for testing this out locally:
1. `npm install -g azurite`
2. `brew install azure-cli` or similar depending on OS
3. Start Azurite: `azurite --location ./azurite/ --debug ./azurite/debug.log`
4. Create container: `az storage container create --name mycontainer --connection-string "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://127.0.0.1:10000/devstoreaccount1;"`
5. Start tusd: `AZURE_STORAGE_ACCOUNT=devstoreaccount1 AZURE_STORAGE_KEY=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw== go run cmd/tusd/main.go -azure-endpoint http://127.0.0.1:10000/devstoreaccount1 -azure-storage mycontainer`

### DIFF
--- a/pkg/azurestore/azurestore.go
+++ b/pkg/azurestore/azurestore.go
@@ -1,13 +1,14 @@
 package azurestore
 
 import (
-	"bufio"
 	"bytes"
 	"context"
-	"encoding/binary"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
+	"io/fs"
+	"os"
 	"strings"
 
 	"github.com/tus/tusd/v2/internal/uid"
@@ -18,6 +19,11 @@ type AzureStore struct {
 	Service      AzService
 	ObjectPrefix string
 	Container    string
+
+	// TemporaryDirectory is the path where AzureStore will create temporary files
+	// on disk during the upload. An empty string ("", the default value) will
+	// cause AzureStore to use the operating system's default temporary directory.
+	TemporaryDirectory string
 }
 
 type AzUpload struct {
@@ -25,6 +31,8 @@ type AzUpload struct {
 	InfoBlob    AzBlob
 	BlockBlob   AzBlob
 	InfoHandler *handler.FileInfo
+
+	tempDir string
 }
 
 func New(service AzService) *AzureStore {
@@ -73,6 +81,7 @@ func (store AzureStore) NewUpload(ctx context.Context, info handler.FileInfo) (h
 		InfoHandler: &info,
 		InfoBlob:    infoBlob,
 		BlockBlob:   blockBlob,
+		tempDir:     store.TemporaryDirectory,
 	}
 
 	err = azUpload.writeInfo(ctx)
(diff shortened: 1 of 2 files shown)


quality-leftovers:
I'm wondering if writing to disk always makes sense since some azure hosting services (like azure container apps) only allow mounting files from azure storage account so you'd be writing to a storage account before uploading to a storage account. Without mounting a storage account everything probably just ends up back in memory.

Looking at the `s3store_part_producer` it seems to solve this already since it supports a `_memory` option for the parts producer. There also seems to be nothing `s3` specific to the producer. Wouldn't it be possible to use the parts generator here?

Acconut:
> I'm wondering if writing to disk always makes sense since some azure hosting services might only allow mounting files from azure storage account (not sure, need to check - just got it for container apps env wrong).

It would seem odd to me that no temporary directory is provided for storing intermediate results. I would expect some temporary storage to always be available.



> Looking at the `s3store_part_producer` it seems to solve this already since it supports a `_memory` option for the parts producer. There also seems to be nothing `s3` specific to the producer. Wouldn't it be possible to re-use the parts generator here?

I am not sure about this. For S3, we split a PATCH request into multiple parts of a fixed size (by default 50MB), which are uploaded and then discarded as they come in. There is a limit on how many parts can be buffer concurrently (by default 20), so there is an upper limit on the memory consumption.

For the azurestore, there is not such chunking, but instead the entire PATCH request is read into a temporary storage before being uploaded. If we want to reuse the part generator from S3, the azurestore must be adjusted so that it also uploads in smaller parts or otherwise we will have the same unlimited memory usage as without this PR.

Does that make sense?

quality-leftovers:
True, it is unusual to have no local storage, but there are VMs Dasv5 and Easv5 without any temp disk. But it is possible to attach "remote" storage disks so it sould be ok.

Regarding the parts generator also true. I was just wondering why the S3 store has such sophisticated logic while the azurestore is a bit dumb :P.

Acconut:
> Regarding the parts generator also true. I was just wondering why the S3 store has such sophisticated logic while the azurestore is a bit dumb :P.

Well, the S3 API has a few restrictions (minimal size of upload parts and maximal number of total parts) that make this kind of logic necessary. I would prefer it if we didn't have to do much of this :)

Would you be interested in testing this PR? Maybe you can run it locally and see if it works well for you? I don't use Azure normally, so a second pair of eyes would help catch issues.

quality-leftovers:
I did some testing running the container both locally and as Azure Container App and was unable to reproduce any high memory or out-of-memory problems. Existing upplication / test suite also worked "as normal".

Acconut:
Ok, thank you very much! Then let's get this released!

Issue #669, opened 2022-02-24
Authentication


meidlinga:
**Is your feature request related to a problem? Please describe.**
When using tusd for multiple users, different users should not be able to access other users uploads in any way.
In my current situation, this applies to all possible operations, tusd supports on uploads.
Eventually in other situations, get could be whitelisted, while I would keep it simple initially.
I am aware, that the currently suggested approach is: "Deal with this in a proxy" But I think providing a reusable example would be a good idea.

**Describe the solution you'd like**
The basic idea would be:
* Intercept postCreate, where we have an upload id and store it together with a user id, or an authentication token. 
* Other operations check the referenced upload id against the authentication token and reject any mismatches.

My currently favoured way would be implementing it in go using the approach described in "Use tusd as package", because it would mean less moving parts. But I am not sure if this is possible, or advisable. 

**Describe alternatives you've considered**
As an alternative I considered combining a proxy and a hook.
The hook would be postCreate, since we have an upload id to register with an Authentication token (hash).
The proxy would be HAProxy with a custom Lua extension, blocking any request, where the uploadid does not match the authentication token.

The benefit here would be, that a proxy probably has better isolation possibilities than implementing it directly in tusd.

**Can you provide help with implementing this feature?**
Yes, since this could be a problem for other developers to, I would like to create and contribute a reusable solution.

**Additional context**
Details, like storing the token securely (in a serverside cache, jwt, cookie, etc), dealing with token refresh must be handled, but are currently not described further.

Acconut:
Yes, that is a very good idea. We have discussed and agreed on implementing additional hooks to facilitate authentication and authorization. Feel free to have a look at the discussions in #431 and #483. However, no work has been done so far as we first want to merge https://github.com/tus/tusd/pull/516 which is a major overhaul of the entire hook system. Making any changes before this is complete is not sensible.

meidlinga:
Linking for future ref: #95

G2G2G2G:
Is this for the uploading too? There's a hook but uppy (for example) doesn't even handle status codes properly nor give you a method of reading the status code

```Go
               PreUploadCreateCallback: func(hook tusd.HookEvent) error {
			return checkauth(hook.Upload)
		},

func checkauth(u tusd.FileInfo) error {

//fancy checks up here

//metadata database query to check on some stuff sent by client via the u.MetaData["keys"]

    switch err {
	case nil:
		return nil
	case pgx.ErrNoRows:
[...]

PR #1074, opened 2024-01-26
hooks: add pre-resume blocking hook

sberthier:
`pre-resume` blocking hook, inspired by `pre-create` and `pre-finish` hooks. Disable by default.

Called on patch request to continue an existing upload. Upload can be rejected with `RejectUpload` and if so `HTTPResponse` is used in error http response. Otherwise response is ignored.

Useful for authentication/authorization of PATCH requests, as evoked in https://github.com/tus/tusd/issues/669#issuecomment-1051924886

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -17,6 +17,7 @@ The table below provides an overview of all available hooks.
 |----------------|-----------|------------------------------------------------------------------------|---------------------------------------------------------------------------------|---------------------|
 | pre-create     | Yes       | before a new upload is created.                                        | validation of meta data, user authentication, specification of custom upload ID | Yes                 |
 | post-create    | No        | after a new upload is created.                                         | registering the upload with the main application, logging of upload begin       | Yes                 |
+| pre-resume     | Yes       | before an existing upload is continued.                                | validation of user authentication                                 | No                 |
 | post-receive   | No        | regularly while data is being transmitted.                             | logging upload progress, stopping running uploads                               | No                  |
 | pre-finish     | Yes       | after all upload data has been received but before a response is sent. | sending custom data when an upload is finished                                  | Yes                 |
 | post-finish    | No        | after all upload data has been received and after a response is sent.  | post-processing of upload, logging of upload end                                | Yes                 |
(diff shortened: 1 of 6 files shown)


Acconut:
Thank you for this PR and congratulations on implementing this on your own! I see the use case for such a hook, but a more general solution that is not exclusive to PATCH requests would be preferable as also mentioned in #669: An authentication hook that is executed before any request to tusd is processed would allow authenticating the users, testing their access rights to the corresponding upload, checking for user resource limits etc. I am not sure what your exact use case is but such an authentication hook would probably also serve your needs.

There has been no work started on this yet, but this PR could serve as a great template for adding the new authentication hook. Let me know what you think about this approach.

sberthier:
An global `auth` hook would server my needs too. 

I see the interest of having that hook, however it will make call hook twice all most at the same time for some request, depending on hooks enabled (`auth` and `pre-create` for instance). Adding a `pre-get` hook would have allowed all routes to be protected just by a cost of an call each.

I could try implement that hook (named `auth` or `pre-all` ?), certainly in pkg/handler/unrouted_handler.go `Middleware`, should not be that different than this `pre-resume`.

Acconut:
It would be helpful if the authentication hook is invoked after the upload information has been fetched. Then this info can be passed to the authentication hook, which can use its upload ID and meta data to decide in combination with the client request to decide whether the access to upload is granted or not. In this case, the authentication hook can not be triggered in the middleware logic, but should be done inside HeadFile, PatchFile, etc.

When creating an upload, I don't think it's necessary to invoke the authentication hook since that task can be done by the pre-create hook. We can avoid the duplicate hook invocation there.

Does that make sense?

sberthier:
Yes it does.

Any preference for the hook name? Could be `pre-access`, invoked before any access done to the upload, Head, Get, Patch, Delete.

Acconut:
pre-access would be a good fit, yes. Thank you for working on this!

In addition to the upload information and request details, we should also include the requested access mode into the hook. For PATCH and DELETE  requests this would be `write` and for HEAD and GET it would be `read`.

One thing I am not sure yet about is how to handle concatenation. When a client requests to concatenate upload A and B together, a pre-access hook should be fired with info about A and B to check whether the client has access to both. So the upload info field in the hook likely has to be an array, which would be different than the structure of other hooks, where we always only have one corresponding upload.

PR #1077, opened 2024-01-30
hooks: add pre-access blocking hook

sberthier:
- To replace `pre-resume` PR (#1074)
- Add `Access.Mode` and `Access.Files` to `HookEvent` for `pre-access` and `pre-create` (when Upload-Concat) hook.
  -  `Files` is an array because of [Concatenation](https://tus.io/protocols/resumable-upload#concatenation) extension that access many files to make a new one
  -  `Mode` is _read_ (Get/Head) or _write_ (Patch/Delete)
- Add `pre-access` before each http request Get/Head/Patch/Delete.
  - This new hook is disabled by default
  - Access can be rejected by setting `RejectAccess` in HookResponse. Http response status code will be 403, if not override in HttpResponse.
  - Useful for authentication/authorization of upload access, based on http request headers for instance

close #1074


pkg/handler/unrouted_handler.go
```diff
@@ -338,7 +340,8 @@ func (handler *UnroutedHandler) PostFile(w http.ResponseWriter, r *http.Request)
 	}
 
 	if handler.config.PreUploadCreateCallback != nil {
-		resp2, changes, err := handler.config.PreUploadCreateCallback(newHookEvent(c, info))
+		access := newAccessInfo(AccessModeRead, partialFileInfos)
+		resp2, changes, err := handler.config.PreUploadCreateCallback(newHookEvent(c, &info, &access))
```
Acconut, line 344: For concatenation, I think we should first emit a pre-access hook to check the permissions for read access on the partial uploads. If that succeeds we can continue with the upload creation and the corresponding pre-create hook. Then users don't have to duplicate access-related logic in pre-access and pre-create.

A good location for that hook is likely a few lines above after the call to `handler.sizeOfUploads`.

pkg/handler/hooks.go
```diff
@@ -26,23 +26,58 @@ type HookEvent struct {
...
+		upload = *info
+	}
+	var access AccessInfo
+	if accessInfo != nil {
+		access = *accessInfo
+	}
 	return HookEvent{
 		Context: c,
-		Upload:  info,
+		Upload:  upload,
```
Acconut, line 67: In hindsight, it would be great if we could set `Upload: nil`, but that would a require a change to the type, which would be breaking.

Acconut (changes requested):
Wow, thank you very much for this huge PR! Congratulations on adding a new hook on your own. I can a look through it focusing on the implementation (and less on the tests and documentation) for now, and left a few comments. But overall this is looking great!

sberthier:
I made the changes you suggested, except the last one to set hook.upload to nil !

Acconut:
Thank you for the updates, I will look into this in two weeks and then we should be able to get this merged.

quality-leftovers:
Hi everyone. Just wanted to ask, if this this is still being worked on?

Acconut:
There hasn't been any work on this since, but we are still interested in adding this to tusd.

videni:
@Acconut   it seems  the PR is ready to merge, any reason why is postponed please?

Acconut:
In general, this PR looks good but I need to give it another review and probably polish a few edges. I'm not sure when I'll get to that as time is tight, but I'll try my best. Apologies for the delay!

Issue #1075, opened 2024-01-28
Uniform Termination Process for Store Implementations: Ensuring Consistent Removal of Binary and .info Files


DarioCalovic:
**Describe the bug**
Issueing the termination process the filestore removes both the binary as well as  the .info file. The gcc store only removes the binary file.

**To Reproduce**
The .info is being ignored explicit: https://github.com/tus/tusd/blob/main/pkg/gcsstore/gcsservice.go#L355

```
func (service *GCSService) FilterObjects(ctx context.Context, params GCSFilterParams) ([]string, error) {

# ...

  if strings.HasSuffix(objAttrs.Name, "info") {
    continue
  }

# ...


```

**Expected behavior**
The expected behavior is that all store implementations, including the GCS store, should uniformly handle the terminating process by removing both the binary files and their corresponding .info files. The termination process should be consistent across different store implementations to ensure a standardized and predictable behavior.

Acconut:
Good catch! `FilterObjects` has been designed to find objects in a bucket which should be composed to form the final object. The `.info` should not be included here and is thus explicitly removed from the list. For terminating the upload, this behavior is not wanted, as you correctly found.

A solution would be to add an object `IncludeInfoObject` to `GCSFilterParams`, which can be set by the termination code to instruct `FilterObjects` to not exclude `.info` objects. Would you be interested in opening a PR for this?

PR #1082, opened 2024-02-07
Add IncludeInfoObject flag for terminating the .info in gcs as well

DarioCalovic:
*Description*
Issueing the termination process the filestore removes both the binary as well as the .info file. The gcc store only removes the binary file.

*How to solve*
A user wrote: 

> A solution would be to add an object IncludeInfoObject to GCSFilterParams, which can be set by the termination code to instruct FilterObjects to not exclude .info objects.

That's what i have done, if i understood the assessment correctly. Everywhere where the `GCSFilterParams` was declared, I included the `IncludeInfoObject` even if the default value would be false anyway. Only at the termination process, the `IncludeInfoObject` is set to `true`.

_Side note: My very first pull request, just trying to help._

Related Issue #1075

### DIFF
--- a/pkg/gcsstore/gcsservice.go
+++ b/pkg/gcsstore/gcsservice.go
@@ -43,6 +43,9 @@ type GCSFilterParams struct {
 
 	// Prefix specifies the prefix of which you want to filter object names with.
 	Prefix string
+
+	// IncludeInfoObject indicates whether to include or ignore the info object.
+	IncludeInfoObject bool
 }
 
 // GCSReader implements cloud.google.com/go/storage.Reader.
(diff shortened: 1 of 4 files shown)


Acconut:
Thank you very much this PR and congrats on your first contribution!

I added a test for `FilterObjects` with `IncludeInfoObject: true`, but it seems to be behaving differently than expected. The returned list only includes the `.info` object, but not the other objects. Maybe you could have a second look at the implementation changes for `FilterObjects`.

DarioCalovic:
@Acconut You are right. Could you look at the latest commit? I am not sure, if i changed the behaviour too much. First of all, i noticed, that this function always returns an empty `""` on the first index. Is this intentionally? Second, i tried to simplify the validation process. Let me know what you think.

Acconut:
> First of all, i noticed, that this function always returns an empty `""` on the first index. Is this intentionally?

I noticed the same behavior during my tests. Unfortunately, I don't know if this is needed or not. The gcsstore was added in #106 by @tab1293 and I haven't worked much with it. Maybe he could shed some light on this :)

> I am not sure, if i changed the behaviour too much.

I'll have another look

DarioCalovic:
Let me do some tests, and i will come back at you.

PR #1097, opened 2024-03-18
handler: Support RUFH draft -03

Acconut:
This PR adds support for the [new draft version](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-resumable-upload-03). It does not include major changes beside allowing a new response code for HEAD requests (which we don't use) and an increment in the interoperability version.

This PR preserves the support for earlier draft versions.


Nealsoni00:
Is there a plan to support the proposed upload progress notifications via informational responses?

Acconut:
We should support that, yes. Can you please open a new issue for this?

PR #1101, opened 2024-03-31
examples: Correct hook type in plugin example

ehsansouri23:
Correct comment in hook example


Acconut (approved):
Thank you very much for bringing this up! I believe the comment had the right intention, but the condition should check for post-finish. I quickly fixed that.

PR #1118, opened 2024-04-22
Handle slash trimmed path in example server

blmhemu:
The tusd handler works on slash trimmed paths i.e `/files`. Making the example server do the same.

**Usecase:**
Just wasted a couple of hours debugging this behaviour using the example server.


Acconut (approved):
Thank you very much! The comment above the call to http.Handle also just adds to the confusion, making this a great improvement.

Just as a reference for others: This mirrors the logic from the tusd CLI: https://github.com/tus/tusd/blob/15f05c021c9770dc0cd2818aa4ef44b9e361f5b2/cmd/tusd/cli/serve.go#L94-L100

PR #1120, opened 2024-04-23
Fix graceful terminate in filestore.

blmhemu:
In the current logic, we first delete the bin file and then the info file. We should instead do the reverse and ignore any errors when deleting the bin file. This ensures that any disgraceful shutdown would properly recover.

*Note:* This is only fixed in the filestore for now - might need to fix for other stores as well.

### DIFF
--- a/pkg/filestore/filestore.go
+++ b/pkg/filestore/filestore.go
@@ -24,10 +24,8 @@ import (
 	"github.com/tus/tusd/v2/pkg/handler"
 )
 
-var (
-	defaultFilePerm      = os.FileMode(0664)
-	defaultDirectoryPerm = os.FileMode(0754)
-)
+var defaultFilePerm = os.FileMode(0664)
+var defaultDirectoryPerm = os.FileMode(0754)
 
 // See the handler.DataStore interface for documentation about the different
 // methods.
--- a/pkg/filestore/filestore.go
+++ b/pkg/filestore/filestore.go
@@ -15,6 +15,7 @@ package filestore
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -175,10 +176,16 @@ func (upload *fileUpload) Terminate(ctx context.Context) error {
 	// Account for server crashes or force kills
 	// Say if killed just after removing the bin file
 	err := os.Remove(upload.binPath)
-	if !os.IsNotExist(err) {
+	if !errors.Is(err, os.ErrNotExist) {
 		return err
 	}
-	return os.Remove(upload.infoPath)
+
+	err = os.Remove(upload.infoPath)
+	if !errors.Is(err, os.ErrNotExist) {
+		return err
+	}
+
+	return nil
 }
 
 func (upload *fileUpload) ConcatUploads(ctx context.Context, uploads []handler.Upload) (err error) {
--- a/pkg/filestore/filestore.go
+++ b/pkg/filestore/filestore.go
@@ -24,8 +24,10 @@ import (
 	"github.com/tus/tusd/v2/pkg/handler"
 )
 
-var defaultFilePerm = os.FileMode(0664)
-var defaultDirectoryPerm = os.FileMode(0754)
+var (
+	defaultFilePerm      = os.FileMode(0664)
+	defaultDirectoryPerm = os.FileMode(0754)
+)
 
 // See the handler.DataStore interface for documentation about the different
 // methods.
@@ -172,13 +174,10 @@ func (upload *fileUpload) GetReader(ctx context.Context) (io.ReadCloser, error) 
 }
 
 func (upload *fileUpload) Terminate(ctx context.Context) error {
-	if err := os.Remove(upload.infoPath); err != nil {
-		return err
-	}
-	if err := os.Remove(upload.binPath); err != nil {
-		return err
-	}
-	return nil
+	// Ignore errors when removing bin to account for server crashes or force kills
+	// This can happen say if the server crashes just after removing the bin file
+	os.Remove(upload.binPath)
+	return os.Remove(upload.infoPath)
 }
 
 func (upload *fileUpload) ConcatUploads(ctx context.Context, uploads []handler.Upload) (err error) {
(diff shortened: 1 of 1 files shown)


Acconut:
Thank you for the PR. I am fine with reversing the order but ignoring all errors appears dangerous to me. A better approach would be to ignore errors when the file was already deleted (using `errors.Is(err, fs.ErrNotExist)`), but properly returning any different error. Or did you experience a different error?

blmhemu:
Thanks for the review - updated to account for other errors.

pkg/filestore/filestore.go
```diff
@@ -172,13 +174,13 @@ func (upload *fileUpload) GetReader(ctx context.Context) (io.ReadCloser, error)
...
 
 func (upload *fileUpload) Terminate(ctx context.Context) error {
-	if err := os.Remove(upload.infoPath); err != nil {
-		return err
-	}
-	if err := os.Remove(upload.binPath); err != nil {
+	// Account for server crashes or force kills
+	// Say if killed just after removing the bin file
+	err := os.Remove(upload.binPath)
+	if !os.IsNotExist(err) {
```
Acconut, line 180: Please use the newer `errors.Is(err, fs.ErrNotExist)` method here.

pkg/filestore/filestore.go
```diff
@@ -24,8 +24,10 @@ import (
 	"github.com/tus/tusd/v2/pkg/handler"
 )
 
-var defaultFilePerm = os.FileMode(0664)
-var defaultDirectoryPerm = os.FileMode(0754)
+var (
```
Acconut, line 27: Can you please remove these unrelated changes?

pkg/filestore/filestore.go
```diff
@@ -172,13 +174,13 @@ func (upload *fileUpload) GetReader(ctx context.Context) (io.ReadCloser, error)
...
-	}
-	if err := os.Remove(upload.binPath); err != nil {
+	// Account for server crashes or force kills
+	// Say if killed just after removing the bin file
+	err := os.Remove(upload.binPath)
+	if !os.IsNotExist(err) {
 		return err
 	}
-	return nil
+	return os.Remove(upload.infoPath)
```
Acconut, line 183: We should also ignore errors when the file was already deleted here and return `nil` in these cases.

Acconut (changes requested):
Thanks for the update! Just a few nits.

pkg/filestore/filestore.go
blmhemu: I think it may not be needed as that would be an illegal call - because both the files would have been already deleted.

pkg/filestore/filestore.go
blmhemu: Ahh - autoformatter stuff.

pkg/filestore/filestore.go
blmhemu: I think it might not work because:
* os.remove wraps the original error `return &PathError{Op: "remove", Path: name, Err: e}`
* os.IsNotExist(err) - retrieves this underlying and matches against `fs.ErrNotExist`.

Acconut: `os.PathError` implements the `Unwrap` method, which is used by `errors.Is` to check if the errors match. The [documentation](https://pkg.go.dev/errors) for `errors` also explicitly mentions `fs.ErrNotExist`, so I would assume it works and that has also been my experience in the past. Or am I missing something?

pkg/filestore/filestore.go
Acconut: The underlying files might be removed by a cron job for cleaning up and in those cases `Terminate` should not return an error because the files are already gone. This is likely an edge case, but since we are already touching this logic it makes sense to do it properly.

pkg/filestore/filestore.go
blmhemu: Used `os.ErrNotExist` (which is same as `fs.ErrNotExist`)

Acconut:
I wonder if this change will actually allow the upload to be cleaned up if only one of the two files remain. Yes, the Terminate method would do that, but a DELETE request would fail because the filestore needs both files present in order to load the upload (and then terminate it). At least that's what I remember right now.

Did you test the termination of an upload where one file has been deleted?

blmhemu:
> I wonder if this change will actually allow the upload to be cleaned up if only one of the two files remain. Yes, the Terminate method would do that, but a DELETE request would fail because the filestore needs both files present in order to load the upload (and then terminate it). At least that's what I remember right now.
> 
> Did you test the termination of an upload where one file has been deleted?

Ahh, you are right, the handler would not allow it to pass and returns err not found at `getUpload` method. Do you have any ideas over here ?

Acconut:
> Also, seeing the `NewUpload` code, it seems to first create the bin file and then the info file. So a server crash here could also lead to un-freed resources. It seems like we may need to relax our checks in some handlers.

I am not sure if adjusting the handler logic is the correct way here. If the upload state on disk got corrupted (e.g. one of the two files is missing), the client will not be able to resume the upload and start a new one. Trying to consider and recover from every possible type of corruption in tusd would be challenging if not impossible. The better way to deal with corrupted states is to start a new upload and let the previous, corrupt upload be cleaned up by a cron job (e.g. a job that removes all files which are untouched for the last 24hrs). This approach would be more thorough as it can also clean up uploads that tusd considers corrupt. Does that make sense?

Nevertheless, I would still merge this PR. Ignoring `fs.ErrNotExist` when deleting files is a good improvement.

blmhemu:
> > Also, seeing the `NewUpload` code, it seems to first create the bin file and then the info file. So a server crash here could also lead to un-freed resources. It seems like we may need to relax our checks in some handlers.
> 
> I am not sure if adjusting the handler logic is the correct way here. If the upload state on disk got corrupted (e.g. one of the two files is missing), the client will not be able to resume the upload and start a new one. Trying to consider and recover from every possible type of corruption in tusd would be challenging if not impossible. The better way to deal with corrupted states is to start a new upload and let the previous, corrupt upload be cleaned up by a cron job (e.g. a job that removes all files which are untouched for the last 24hrs). This approach would be more thorough as it can also clean up uploads that tusd considers corrupt. Does that make sense?

Yep ! The handler can only do so much. The cronjob approach looks good (the job can look at info file and match it with the os.stat and remove them after a threshold).

> Nevertheless, I would still merge this PR. Ignoring `fs.ErrNotExist` when deleting files is a good improvement.

Sure. NP.

Acconut (approved):

Acconut:
Thank you for working on this!

PR #1123, opened 2024-04-24
s3store: Allow `+` in object IDs

OwynWang:
To be compatible with object id that contains `+` .

### DIFF
--- a/pkg/s3store/s3store.go
+++ b/pkg/s3store/s3store.go
@@ -1148,7 +1148,7 @@ func (store S3Store) deleteIncompletePartForUpload(ctx context.Context, uploadId
 }
 
 func splitIds(id string) (objectId, multipartId string) {
-	index := strings.Index(id, "+")
+	index := strings.LastIndex(id, "+")
 	if index == -1 {
 		return
 	}
--- a/pkg/s3store/s3store.go
+++ b/pkg/s3store/s3store.go
@@ -1148,6 +1148,8 @@ func (store S3Store) deleteIncompletePartForUpload(ctx context.Context, uploadId
 }
 
 func splitIds(id string) (objectId, multipartId string) {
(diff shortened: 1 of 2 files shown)


Acconut:
Thank you for this PR. I wonder if there are S3-compatible which include + in their multipart IDs. AWS does not do this as far as I know, but other implementations might differ. DigitalOcean Spaces includes ~ if I recall correctly.

OwynWang:
So we could not ensure whether object ID and multipart ID contains `+` or other special characters. Maybe we could encode these IDs use base64.

OwynWang:
Amend committed.

Acconut:
> Maybe we could encode these IDs use base64.

This has been discussed before, but also comes with downsides. It would be a breaking change as the URL format changes and some people depend on the format to extract the object ID. Switching to a base 64 encoding would breaking such integrations.

In addition, we want to move away from including the multipart ID entirely in the upload URL, as this prevents users from fully customizing the upload URL. It will also make the upload URL from the s3store similar to how all other storages work.

For that reason, I don't think we should invest more time into looking into base64 encoding IDs. Since I don't remember any S3 provider using `+` in multipart IDs, we can merge this for now. If people run into issues, we will revert it.

Acconut (changes requested):
I just saw that you changed this PR to use base64 encoding. Please revert it to the original version where the last segment of the upload ID was used as the multipart ID. Then we can merge this.

OwynWang:
> I just saw that you changed this PR to use base64 encoding. Please revert it to the original version where the last segment of the upload ID was used as the multipart ID. Then we can merge this.
OK, reverted.

Acconut (approved):

Acconut (approved):
Thank you! I added a test case to ensure that this works and keeps working.

PR #1129, opened 2024-05-01
A redis based lock for distribued tusd instances

whytheplatypus:
This introduces a redis backed locker. This is probably most useful for tusd instances that are distributed, ephemeral, or have to scale up and down.

It uses the redis pub/sub feature to request a lock be released.

The tests use what looks like a pretty [nice little redis implementation](https://github.com/alicebob/miniredis) meant for testing.


Acconut:
Thank you very much for this PR! I will have a more thorough look next week. There are a few things I would like to add to this PR are: documentation for the package itself, documentation for our homepage (https://tus.github.io/tusd), integration with the CLI (so that it can also be used with the tusd binary), and maybe more in-depth tests to ensure that the release notifications also work properly. But I can take care of those :)

pcfreak30:
Just thought I would point this out as you implemented your own locking design.

https://github.com/go-redsync/redsync is available and I likely will be switching to https://github.com/redis/rueidis soon and https://github.com/redis/rueidis/blob/main/rueidislock/lock.go is a route.

@whytheplatypus

Acconut:
@pcfreak30 I am not sure what you mean here. This PR is using the redsync package already instead of implementing its own locking approach. Do you want to imply that using the lock from rueidis is preferable over redsync?

pcfreak30:
> @pcfreak30 I am not sure what you mean here. This PR is using the redsync package already instead of implementing its own locking approach. Do you want to imply that using the lock from rueidis is preferable over redsync?

woops i did *not* see that. I saw some of the pub/sub code and was thinking the locking system was re-implemented vs using a established library.

Other that that I would look at `rueidis` because I myself am probably going to have to move all my redis stuff to it, even if i stay using `redsync` as a library, because im getting a lot of connection pool timeouts and from reading, i think `rueidis` will handle things better and performs better overall based on their readme vs `go-redis`.

whytheplatypus:
> Awesome, would be great if this gets merged.
> 
> @whytheplatypus, I'm wondering if you've had a chance to run this on production and how it looks so far?

we have! so far it's looking solid. We've been able to scale horizontally cleanly and seen an increase in stability, though I don't have before and after numbers for you.

We did have to make one change in the lock exchange behavior, using short lived a pub sub channel per lock rather than using only one long lived one. This was because that one channel was getting full, while the short lived channels don't and also seem to perform better. It probably could have been done by adjusting some redis config but the change also simplified the code and so far has been working well at scale. I'll try and get that change in this PR in the next couple days.

whytheplatypus:
> > @pcfreak30 I am not sure what you mean here. This PR is using the redsync package already instead of implementing its own locking approach. Do you want to imply that using the lock from rueidis is preferable over redsync?
> 
> woops i did _not_ see that. I saw some of the pub/sub code and was thinking the locking system was re-implemented vs using a established library.
> 
> Other that that I would look at `rueidis` because I myself am probably going to have to move all my redis stuff to it, even if i stay using `redsync` as a library, because im getting a lot of connection pool timeouts and from reading, i think `rueidis` will handle things better and performs better overall based on their readme vs `go-redis`.

maybe we can look at getting an interface in front of the needed interactions with redis, so that swapping libraries is an option.

pcfreak30:
@whytheplatypus im about to switch to this in a fork of `tusd` (I am also sticking to the default v9 redis client for now), and one thing i noticed is the assumption to create a new redis instance. I will need to be able to pass an existing in as well. While ill modify that in my fork, would appreciate you adding it.

whytheplatypus:
> @whytheplatypus im about to switch to this in a fork of `tusd` (I am also sticking to the default v9 redis client for now), and one thing i noticed is the assumption to create a new redis instance. I will need to be able to pass an existing in as well. While ill modify that in my fork, would appreciate you adding it.

I added something that should let you do this. That said we ran into some trouble where being able to recreate a redis client and connection from a uri allowed us to recover. We didn't look into it enough to find a specific cause so it may or may not apply here.

pcfreak30:
> > @whytheplatypus im about to switch to this in a fork of `tusd` (I am also sticking to the default v9 redis client for now), and one thing i noticed is the assumption to create a new redis instance. I will need to be able to pass an existing in as well. While ill modify that in my fork, would appreciate you adding it.
> 
> I added something that should let you do this. That said we ran into some trouble where being able to recreate a redis client and connection from a uri allowed us to recover. We didn't look into it enough to find a specific cause so it may or may not apply here.

FYI, you need to re-base/merge the branch. Would liker to see this get merged in 

@Acconut .

pkg/redislocker/new.go
```diff
@@ -0,0 +1,55 @@
+package redislocker
```
Acconut, line 1: Some documentation for Godoc would be helpful to have. Both for the package itself and key methods and types.

pkg/redislocker/redislocker.go
```diff
@@ -0,0 +1,148 @@
...
+	"errors"
+	"time"
+
+	"golang.org/x/exp/slog"
+
+	"github.com/tus/tusd/v2/pkg/handler"
+)
+
+var (
+	LockExchangeChannel = "tusd_lock_release_request_%s"
```
Acconut, line 14: It would be helpful the users could configure a prefix for these keys, so they scope them and avoid collision with other services.

pkg/redislocker/redislocker.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (l *redisLock) Lock(ctx context.Context, releaseRequested func()) error {
+	l.logger.Debug("locking upload", "id", l.id)
+	if err := l.requestLock(ctx); err != nil {
+		return err
+	}
+	go l.exchange.Listen(l.ctx, l.id, releaseRequested)
+	go func() {
+		if err := l.keepAlive(l.ctx); err != nil {
```
Acconut, line 72: Any error that isn't related to the context ending should be logged here as well. `l.keepAlive` performs some logging, but this could be moved here.

pkg/redislocker/redislocker.go
```diff
@@ -0,0 +1,148 @@
...
+
+func (l *redisLock) keepAlive(ctx context.Context) error {
+	//insures that an extend will be canceled if it's unlocked in the middle of an attempt
+	for {
+		select {
+		case <-time.After(time.Until(l.mutex.Until()) / 2):
+			l.logger.Debug("extend lock attempt started", "time", time.Now())
+			_, err := l.mutex.ExtendContext(ctx)
+			if err != nil {
+				l.logger.Error("failed to extend lock", "time", time.Now(), "error", err)
```
Acconut, line 118: Ideally, failures to extend the lock/lease would be communicated to the main request handler, which can then shut down the request handling. Currently, the locker interface isn't well adapted for this yet, but I think calling `releaseRequested` would have the same effect as the handler is then starting to shut down.

pkg/redislocker/redislocker.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/tus/tusd/v2/pkg/handler"
+)
+
+var (
+	LockExchangeChannel = "tusd_lock_release_request_%s"
+	LockReleaseChannel  = "tusd_lock_released_%s"
+	LockExpiry          = 8 * time.Second
+)
+
+type LockExchange interface {
```
Acconut, line 19: `LockExchange` seems to be only used in `BidirectionalLockExchange`, so these interface could be merged.

pcfreak30:
> I don't have the capacity right now to work on these changes on my own, so any help is greatly appreciated.

How should I do that as I assume I dont have access to edit the branch this is on.

Acconut:
Depends a bit on whether @whytheplatypus is interested in continuing this PR or not. If not, it's best you open a new PR and we move the discussion there. If you want to get started ASAP, this is probably the best way for now.

pcfreak30:
> @Acconut I may have some time this week to take a look at this again.
> 
> for context this was done for a cdc project and the currently in use code can be found here https://github.com/CDCgov/data-exchange-upload/tree/main/upload-server/pkg/redislocker

Please LMK if you end up don't. My work relies on this and I will have to take on this PR if it comes to it so it can be upstreamed and I don't keep the burden of a fork.

pcfreak30:
> > Please LMK if you end up don't. My work relies on this and I will have to take on this PR if it comes to it so it can be upstreamed and I don't keep the burden of a fork.
> 
> @pcfreak30 would you like permissions to my fork? I'm happy to do that if it would be easier.

I have several tasks to take on and I would prefer if you would complete this PR so I don't have to reverse engineer your thoughts :P.

But if you don't, then I will have to do it eventually...

whytheplatypus:
@Acconut I think I got most of your comments. Docs are incoming...

whytheplatypus:
@Acconut ok docs and tests are looking better. (sorry for the churn I couldn't reproduce the test failures locally)

As I made these changes I'm wondering if eventually the `Lock` interface could return a context so that something using that lock could exit if the lock was lost unexpectedly, in this case keep alive failing for some unknown reason.

whytheplatypus:
@Acconut would it make more sense for me to move this into it's own repo and folks can use it as a package (as it would be used anyway)? Happy to do whatever works best for you and the tus team.

Issue #1124, opened 2024-04-29
Am I doing the gcp bucket set up right? keeps failing with EOF error


emilaleksanteri:
**Question**
Am I doing the gcp bucket set up right? keeps failing with EOF error, if I change the storage to be local file storage, then everything works perfect. My client is pretty much identical from the Uppy.js example for react
**Setup details**
Please provide following details, if applicable to your situation:
- Operating System: macOS
- Used tusd version: 2.4.0
- Used tusd data storage: Google Cloud Storage
- Used tusd configuration: Using the go library to have it on my server
- Used tus client library:  tus-js-client, Uppy.js

Here is my current handler creation:
```go
func FileUploadHandler() *tusd.UnroutedHandler {
	bucket := "sample-bucket"
	absPath, _ := filepath.Abs("./")
	gcpCredentials := filepath.Join(absPath, "config", "credentials.json")

	gcsService, err := gcsstore.NewGCSService(gcpCredentials)
	if err != nil {
		log.Fatalf("Unable to create gcsService: %s\n", err.Error())
	}

	locker := memorylocker.New()

	store := gcsstore.New(bucket, gcsService)
	store.ObjectPrefix = "store-products-csv"

	composer := tusd.NewStoreComposer()
	store.UseIn(composer)
	locker.UseIn(composer)

	handler, err := tusd.NewUnroutedHandler(tusd.Config{
		BasePath:              "/store-products-csv/",
		StoreComposer:         composer,
		NotifyCompleteUploads: true,
		NotifyCreatedUploads:  true,
		Cors: &tusd.CorsConfig{
			Disable: true,
		},
	})

	if err != nil {
		log.Fatalf("could not create tusd handler: %s\n", err.Error())
	}

	go func() {
		for {
			event := <-handler.CompleteUploads
			fmt.Printf("\nSuccessfully uploaded a file with an id: %v", event.Upload.ID)
			createdUpload := <-handler.CreatedUploads
			fmt.Printf("\nCreated an upload with an id: %v", createdUpload.Upload.ID)
		}
	}()

	return handler

}
```

routes set up with chi:
```go
        r.Post("/store-products-csv/", fileRoutes.PostFile)
	r.Head("/store-products-csv/{id}", fileRoutes.HeadFile)
	r.Patch("/store-products-csv/{id}", fileRoutes.PatchFile)
	r.Get("/store-products-csv/{id}", fileRoutes.GetFile)
```

Acconut:
[...]

PR #1131, opened 2024-05-07
gcsstore: Fix error creating new uploads

emilaleksanteri:
Change r.Read(buf) to io.ReadFull(buf) as discussed in issue #1124 to fix EOF error on .info read from GCS

### DIFF
--- a/pkg/gcsstore/gcsservice.go
+++ b/pkg/gcsstore/gcsservice.go
@@ -47,6 +47,7 @@ type GCSFilterParams struct {
 
 // GCSReader implements cloud.google.com/go/storage.Reader.
 // It is used to read Google Cloud storage objects.
+// TODO: Remain, Size, ContentType seem to only be used in tests. Maybe we can replace this interface with an io.ReadCloser?
 type GCSReader interface {
 	Close() error
 	ContentType() string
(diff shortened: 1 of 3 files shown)


Acconut:
The tests are failing now because they are running out of memory: https://github.com/tus/tusd/actions/runs/8983182072/job/24727119301?pr=1131

The test has probably be adjusted, so that io.ReadAll can terminate.

emilaleksanteri:
> The tests are failing now because they are running out of memory: https://github.com/tus/tusd/actions/runs/8983182072/job/24727119301?pr=1131
> 
> The test has probably be adjusted, so that io.ReadAll can terminate.

On it, can I run the tests for the gcsstore with just go test or is there some other stuff I should know for local dev on the repo?

emilaleksanteri:
@Acconut switched io.ReadAll to io.ReadFull from a buffer size of the ReadObject reader, this seemed to also solve the EOF issue and solved all the issues with the tests.

Acconut:
Thanks for your update, but I still prefer using `io.ReadAll` instead of relying on `Size()` to give us a correct file size. I added a commit, switching to `io.ReadAll`, but also adjusting the tests, which had a custom implementation of GCSReader, which never returned an EOF, causing `io.ReadAll` to run out of memory.

Acconut (approved):
Thank you very much for debugging this!

PR #1155, opened 2024-07-12
filestore: add dirs(0775) and files(0644) chmod(2) perms options

avreg:
Hi.

This pool request allows flexible use of DAC access attributes to access intermediate directories, if used.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -6,4 +6,8 @@ set -o pipefail
 
 . /usr/local/share/load-env.sh
 
+if [ -n "$UMASK" ]; then
+   umask "$UMASK"
+fi
+
 exec tusd "$@"
--- a/pkg/filestore/filestore.go
+++ b/pkg/filestore/filestore.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
 
@@ -25,22 +26,22 @@ import (
 	"github.com/tus/tusd/v2/pkg/handler"
 )
 
-var defaultFilePerm = os.FileMode(0664)
-var defaultDirectoryPerm = os.FileMode(0754)
-
 // See the handler.DataStore interface for documentation about the different
 // methods.
 type FileStore struct {
 	// Relative or absolute path to store files in. FileStore does not check
 	// whether the path exists, use os.MkdirAll in this case on your own.
 	Path string
+
+	DirPerm  fs.FileMode
+	FilePerm fs.FileMode
 }
 
 // New creates a new file based storage backend. The directory specified will
 // be used as the only storage entry. This method does not check
 // whether the path exists, use os.MkdirAll to ensure.
-func New(path string) FileStore {
-	return FileStore{path}
+func New(path string, dirPerms uint32, filePerms uint32) FileStore {
+	return FileStore{path, os.FileMode(dirPerms), os.FileMode(filePerms)}
 }
 
 // UseIn sets this store as the core data store in the passed composer and adds
@@ -73,7 +74,7 @@ func (store FileStore) NewUpload(ctx context.Context, info handler.FileInfo) (ha
 	}
 
 	// Create binary file with no content
-	if err := createFile(binPath, nil); err != nil {
+	if err := createFile(binPath, store.DirPerm, store.FilePerm, nil); err != nil {
 		return nil, err
 	}
 
@@ -81,6 +82,8 @@ func (store FileStore) NewUpload(ctx context.Context, info handler.FileInfo) (ha
 		info:     info,
 		infoPath: infoPath,
 		binPath:  binPath,
+		dirPerm:  store.DirPerm,
+		filePerm: store.FilePerm,
 	}
 
 	// writeInfo creates the file by itself if necessary
@@ -133,6 +136,8 @@ func (store FileStore) GetUpload(ctx context.Context, id string) (handler.Upload
 		info:     info,
 		binPath:  binPath,
 		infoPath: infoPath,
+		dirPerm:  store.DirPerm,
+		filePerm: store.FilePerm,
 	}, nil
 }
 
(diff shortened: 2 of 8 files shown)


docker/entrypoint.sh
```diff
@@ -6,4 +6,8 @@ set -o pipefail
 
 . /usr/local/share/load-env.sh
 
+if [ -n "$UMASK" ]; then
```
Acconut, line 9: This change seems unrelated to the PR itself. Can you please remove it if that's the case?

pkg/filestore/filestore.go
```diff
@@ -18,29 +18,30 @@ import (
...
+	DirPerm  fs.FileMode
+	FilePerm fs.FileMode
 }
 
 // New creates a new file based storage backend. The directory specified will
 // be used as the only storage entry. This method does not check
 // whether the path exists, use os.MkdirAll to ensure.
-func New(path string) FileStore {
-	return FileStore{path}
+func New(path string, dirPerms uint32, filePerms uint32) FileStore {
```
Acconut, line 43: We shouldn't change the signature of exported functions as this would be a breaking change. Instead, we can keep `New` unchanged and users can set the `DirPerm` and `FilePerm` fields after they called `New`.

Acconut (commented):
Thank you for this PR! I had a brief look at this, but I was wondering what other file permissions you want to use? Should we change the default file permission?

docker/entrypoint.sh
avreg: On many UNIX systems, the default umask is 022. 
This significantly limits the usefulness of the new options in the context of "group" and "other" permissions
If we don't use the UMASK environment variable, we have to call umask(2) directly in the go application (hard-coded zero or add new option like -fs-umask=umask).

Acconut: I don't know how umask works. Can you provide some hints or links so I can understand what this is about?

avreg: Here is an example using Go code
https://stackoverflow.com/a/61645606/23561452

I needed this patch to get group write access to the intermediate directories of the downloaded file.

```
// pre-create hook:
return {
            ChangeFileInfo: {
                ID: "subdir1/subdir2/.../basename",
            },
        };
```

pkg/filestore/filestore.go
avreg: I corrected the code, please take a look.

docker/entrypoint.sh
Acconut: Sorry, I meant that I don't know what `umask` is in general.

avreg: Ok, without this PR, for example, it is not possible to get write/delete access a uploaded file for o(ther) users who are not a tusd user and not in the tusd group, It is also not possible to get write/delete access to созданным subdirectories for members of the tusd group.

Do you know how traditional Unix permissions for files and directories work? man chmod(2)?
Without realizing it, I'm afraid I can't explain it to you succinctly.

Thanks for your work.

avreg: A simpler solution might have been to use umask(2) with this patch:

```diff
diff --git a/pkg/filestore/filestore.go b/pkg/filestore/filestore.go
index 6f9ab52..c2f5697 100644
--- a/pkg/filestore/filestore.go
+++ b/pkg/filestore/filestore.go

-var defaultFilePerm = os.FileMode(0664)
-var defaultDirectoryPerm = os.FileMode(0754)
+var defaultFilePerm = os.FileMode(0666)
+var defaultDirectoryPerm = os.FileMode(0777)
```

A matter of taste.

Acconut: Ok, so I read up a bit on umask and this is what I understand so far: The umask is a property of the process and it restricts the permissions that are set on newly created files and directories. Even if we tell Go to create a file with permission 0664, the file will actually end up with 0644 if our umask is 022. Is that correct?

If so, this does not seem to be a problem specific to Docker container, but something that affects tusd processes in general.

I like the simpler solution that you are proposing. Instead of defining permissions in the code that won't end up on the created files, we can delegate the control of this to the end user. Either by providing a flag to set the umask on the process or by delegating the work of setting the correct umask to the end user. What do you think? How is this handled in other Docker images?

avreg: Yes, now you exactly understand the mechanics of how umask(2) works.

```diff
diff --git a/docker/entrypoint.sh b/docker/entrypoint.sh
index 88a70ef..f1934bb 100644
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -6,4 +6,18 @@ set -o pipefail
 
 . /usr/local/share/load-env.sh
 
+if printenv UMASK >/dev/null; then
+   umask "$UMASK"
+fi
```

With this and the other above patches and without explicitly setting env UMASK for the docker container, the tusd daemon will create new directories/files with 0755/0644 permissions (since usually linux distributions set umask 022 for users by default).

If a user needs something special, like "wx" access for a group, they will have to pass env UMASK=002 to the container. And then the permissions of newly created directories/files will become 0775/0664.

This changes is lightweight and should satisfy most users. Unless the user wants different permissions for files and directories, e.g. 0755 (g=rx) for directories but 0660 (g=rwx) for files ;)

Acconut: Thanks for the patience. I'm slowly getting it. Do you know how other Docker images handle this? Do they expose a UMASK environment variable as well? Or are there other patterns to control file permissions from containers?

avreg: The docker ecosystem has "Isolate containers with a user namespace", which would be interesting to use with a TUSD container, but it won't solve the UMASK problem.

Acconut: This doesn't really ask my question, so let me try to rephrase it: Do other Docker images also use the `UMASK` environment variable for configuring the umask of the process running in the Docker container? Does the Docker community have other defacto ways of configuring the umask?

avreg: Я не использовал никакие другие проекты, где мне нужно было бы менять UMASK. Поэтому мой ответ на ваш вопрос - я не знаю и я не знаю каких либо де-факто решений. 

Однако, если поискать в google, можно найти много постов, где люди ищут подобные решения в других проектах.

Вообще, касательно umask существует не так много решений:
1. pam_umask(8) (systemwide only),
2. RC, like bashrc (systemwide and user),
3. umask (CLI).

PR #1159, opened 2024-07-24
slog: replace golang.org/x/exp/slog with log/slog

fosmjo:


Acconut:
Thank you for this PR! I am supportive of this change, but we have to defer merging it until the next major release since this is a breaking change. At the time of writing, there is no concrete plan for the next major release.

PR #1170, opened 2024-08-20
http2 cleartext support

kristophM:
This PR is in reference to [this](https://github.com/tus/tus-js-client/issues/700#issuecomment-2208270992) thread in the tus-js-client repository issues. @Acconut invited us to create a lightweight PR to solve this issue, on the tusd side. 

PROBLEM

In GCP Cloud Run, HTTP/2 traffic is accepted but TLS is terminated at the load balancer and traffic is forwarded to the tusd container as cleartext (non-encrypted) http. This is problematic because tusd expects TLS for HTTP/2 traffic. 

SOLUTION

[Golang's net/http2 library ](https://pkg.go.dev/golang.org/x/net/http2/h2c#section-documentation) allows for a drop in support of HTTP/2 in cleartext (h2c), with simultaneous capability for HTTP/1.1 cleartext. Meaning, if traffic is identified to be HTTP/1.1, the h2c handler will bypass HTTP/2 and serve based on HTTP/1.1. 

This PR uses the `newHandler()` method to wrap the existing `mux` handler, such that no additional configuration is needed  to handle either HTTP version. HTTP/2 encrypted traffic can also be handled as usual by making calls directly to tusd via https. If behind a load balancer such as Cloud Run, all that would be required in terms of configuration is the `-behind-proxy` flag. 

TESTING
We built an image based on this PR and tested it for file uploads on Cloud Run with tus-js-client and it performed successfully.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ tusd_*_*
 __pycache__/
 examples/hooks/plugin/hook_handler
 .idea/
+.vscode/
--- a/cmd/tusd/cli/flags.go
+++ b/cmd/tusd/cli/flags.go
@@ -15,6 +15,7 @@ var Flags struct {
 	HttpHost                         string
 	HttpPort                         string
 	HttpSock                         string
+	EnableH2C                        bool
 	MaxSize                          int64
 	UploadDir                        string
 	Basepath                         string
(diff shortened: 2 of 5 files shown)


Acconut:
Thank you for this PR! The use of h2c on its own looks good, but I would prefer if it sits behind a feature flag and is not enabled by default. The caveat about memory consumption from https://github.com/tus/tus-js-client/issues/700#issuecomment-2208270992 is concerning enough to let user opt into this behavior. I can imagine that in the future we have multiple flags to enable different HTTP versions, e.g. `--enable-h2c` and `--enable-h3`.

kristophM:
Thanks @Acconut good point - pushed a new commit. Let me know if it addresses your concern.

Acconut:
Thanks for the updates! I will try them out on my own in about two weeks

Acconut:
Merged in https://github.com/tus/tusd/commit/dfaf9f76332b477fa4be87748315bab61f1b15fd, thank you very much for this!

Acconut:
This is released in https://github.com/tus/tusd/releases/tag/v2.5.0.

Issue #136, opened 2017-06-28
Wrong URL when new upload is created


alexandrevicenzi:
The [absFileURL](https://github.com/tus/tusd/blob/master/unrouted_handler.go#L652) method returns a wrong URL if my server list in a random port. It only works if my server is running on ports 80 for HTTP and 443 for HTTPS.

![screenshot from 2017-06-28 14-46-00](https://user-images.githubusercontent.com/2637672/27652137-07e29afc-5c11-11e7-950f-0640eca4277d.png)

Acconut:
Are you using a proxy in front of tusd? If so, did you set the `-behind-proxy` flag? Also what exactly do you mean with "random port"? Are you referring to listen on port 0?

alexandrevicenzi:
I'm using tusd as a lib, and I'm running on port 7781 behind nginx on the same port (dev only). `RespectForwardedHeaders` is set to false.

What I fould is that `absFileURL` doens't get the port in any case.

Acconut:
`RespectForwardedHeaders` should be enabled to ensure that the correct URL is generated. Did you try to enable it yet? Also, did you solve the problem or did you close the issue by accident?

alexandrevicenzi:
The problem was on nginx, I just enabled proxy_redirect to change the Location url.

Acconut:
Glad to hear that.

kvz:
Would it make sense to add this contribution to https://github.com/tus/tusd/blob/master/.infra/files/nginx.conf ?

jheaff1:
This still seems to be an issue, with `-behind-proxy` the Location returned by the Creation request strips the port, assuming that it is on 80/443. Is there any way around this?

jheaff1:
Sorry if I was unclear, I mean if the proxy is NOT running on 80/443, eg running on 8443. The returned Location header does not include “:8443” in this case if the “-behind-proxy” flag is set

Acconut:
Ok, now I understand. Did you follow the instructions at https://github.com/tus/tusd/blob/main/docs/faq.md#can-i-run-tusd-behind-a-reverse-proxy for configuring your proxy?

jheaff1:
Yep I followed those instructions. I also used the following config as a template:

https://github.com/tus/tusd/blob/main/examples/nginx.conf

Acconut:
> Yep I followed those instructions

[...]

PR #1190, opened 2024-09-19
Add note about port of X-Forwarded-Host to nginx.conf example

Chaoses-Ib:
tusd will use `X-Forwarded-Host` as the host if `-behind-proxy` is enabled. However, the example Nginx config doesn't put the port number into `X-Forwarded-Host`, but only the address. This makes the URL returned by tusd to be on port 80/443 instead of what Nginx is actually listening on.

Changing `proxy_set_header X-Forwarded-Host $host` to `proxy_set_header X-Forwarded-Host $host:$server_port` can fix this problem. But this leads to the occurrence of `:80`/`:443` in the returned URL even the client doesn't put `:80`/`:443` in the `Host` header, which can be a little weird. So I just add a comment about this.

Related issue: #136

### DIFF
--- a/examples/nginx.conf
+++ b/examples/nginx.conf
@@ -57,6 +57,7 @@ server {
         proxy_http_version       1.1;
 
         # Add X-Forwarded-* headers
+        # Replace $host with $host:$server_post if server is not listening on 80/443
         proxy_set_header X-Forwarded-Host $host;
         proxy_set_header X-Forwarded-Proto $scheme;
 


Acconut:
Thanks for the PR! We can also consider adding a logic in tusd's URL generation to drop the default ports for HTTP(S) if they are used.

TaridaGeorge:
I've stumbled upon the same problem. I've used the fix mentioned above but I think the absFileURL from unrouted_handler.go.
https://github.com/tus/tusd/blob/fb5b6209a7e1c807c9de8f081fa39ba8386b91d0/pkg/handler/unrouted_handler.go#L403-L404

Acconut:
Thank you for raising this issue. I updated our Nginx example and the URL generation code to exclude default ports.

Issue #1192, opened 2024-09-22
Data race using slog


wongak:
**Describe the bug**
`tusd.Config{}` Logger is set. When performing an upload a data race is reported:
```
==================
WARNING: DATA RACE
Write at 0x00c0006ae3b0 by goroutine 349:
  github.com/tus/tusd/v2/pkg/handler.(*UnroutedHandler).PostFile()
      .../vendor/github.com/tus/tusd/v2/pkg/handler/unrouted_handler.go:398 +0xb60
...
Previous read at 0x00c0006ae3b0 by goroutine 398:
  github.com/tus/tusd/v2/pkg/handler.(*httpContext).Value()
      <autogenerated>:1 +0x44
```

**To Reproduce**
1. Configure a logger
2. Run with `-race`

Currently still investigating which other circumstances lead to this.

But
```
	c.log = c.log.With("id", id)
```
writing to a context without creating a copy is probably not correct.

**Expected behavior**
No data race should happen

**Setup details**
Please provide following details, if applicable to your situation:
- Operating System: macOS
- Used tusd version: github.com/tus/tusd/v2 v2.4.0
- Used tusd data storage: S3 with local minio

wongak:
Unfortunately this uncovers quite a few more data race issues, which go beyond the scope of this issue:

```
==================
WARNING: DATA RACE
Write at 0x00c0000c40a0 by goroutine 205:
  github.com/tus/tusd/v2/pkg/handler.(*bodyReader).closeWithError()
      .../tusd/pkg/handler/body_reader.go:123 +0x3c
  github.com/tus/tusd/v2/pkg/handler.UnroutedHandler.newContext.func1()
      .../tusd/pkg/handler/context.go:66 +0x13c

Previous read at 0x00c0000c40a0 by goroutine 202:
  github.com/tus/tusd/v2/pkg/handler.(*bodyReader).Read()
      .../tusd/pkg/handler/body_reader.go:44 +0x40
  io.ReadAll()
      .../go/1.23.1/libexec/src/io/io.go:712 +0x84
```

and

```
==================
WARNING: DATA RACE
Write at 0x00c0002180e0 by goroutine 130:
  github.com/tus/tusd/v2/pkg/s3store.(*s3Upload).uploadParts.func2()
      .../tusd/pkg/s3store/s3store.go:515 +0x4dc
  github.com/tus/tusd/v2/pkg/s3store.(*s3Upload).uploadParts.gowrap2()
[...]

PR #1193, opened 2024-09-22
Data race fix for context slog.Logger

wongak:
See https://github.com/tus/tusd/issues/1192

This adds a mock slog Logger to create a data race issue. Run go test -race to trigger.

I am currently thinking about different approaches for a fix. But it's a bit complicated.

The way `httpContext` is implemented does not allow any mutations. It would need to be cloned, and any requests would need to be copied as well with `http.Request.WithContext` to contain the httpContext.


Acconut:
Thank you for looking into this! As mentioned in https://github.com/tus/tusd/issues/1192#issuecomment-2393270585, it's probably easier to set the `id` when creating the context. We might have to move some logic around for this, but that's doable.

Acconut:
@wongak I enabled the race detector for CI in this PR and reverted your fix (for now) in an attempt to reproduce this race warning. However, neither locally nor on CI is the `WithSlog` test failing. Is it reporting a race warning locally for you?

wongak:
I can't reproduce it either. Which is a bit confusing. I got the errors before applying your changes. But with your fixes, it seems to be fine.

Acconut:
The data race report from your [original comment](https://github.com/tus/tusd/issues/1192#issue-2541044142) mentions the use of `(*httpContext).Value()`, although it doesn't provide a stack trace:

```
Previous read at 0x00c0006ae3b0 by goroutine 398:
  github.com/tus/tusd/v2/pkg/handler.(*httpContext).Value()
      <autogenerated>:1 +0x44
```

I wonder if this warning appears when the context is accessed by the data store, for example.

But if you cannot reproduce this anymore, we can also close this issue since the other data races have been addressed.

Acconut:
I am closing this PR since we are not able to reproduce this data race anymore. Feel free to reopen it if the problem appears again!

PR #1205, opened 2024-10-11
azurestore: Migrate to current Azure SDK

xelat09:
Hi 

Here an effort to migrate to the new azure sdk, manual tests with azurite did work. I could also do some testing with the real thing, once merged.
Maybe you could have a review, why some tests are not green yet and if this is mergable at all :). 

Regards


xelat09:
I just retried the tests in bash and it worked as expected

pkg/azurestore/azureservice.go
```diff
@@ -67,87 +72,92 @@ type AzBlob interface {
...
 	case "container":
-		containerAccessType = azblob.PublicAccessContainer
+		containerAccessType = container.PublicAccessTypeContainer
 	case "blob":
-		containerAccessType = azblob.PublicAccessBlob
-	case "":
+		containerAccessType = container.PublicAccessTypeBlob
 	default:
-		containerAccessType = azblob.PublicAccessNone
+		containerAccessType = ""
```
Acconut, line 114: The previous default was `azblob.PublicAccessNone`. Was this change intentional?

pkg/azurestore/azureservice.go
```diff
@@ -67,87 +72,92 @@ type AzBlob interface {
 }
 
 type BlockBlob struct {
-	Blob       *azblob.BlockBlobURL
-	AccessTier azblob.AccessTierType
-	Indexes    []int
+	Blob           *blockblob.Client
```
Acconut, line 75: I'm a bit surprised by the field being named `Blob` but the type indicating it's a client object. Can you shed some light on this?

pkg/azurestore/azureservice.go
```diff
@@ -67,87 +72,92 @@ type AzBlob interface {
...
+	cred, err := azblob.NewSharedKeyCredential(config.AccountName, config.AccountKey)
 	if err != nil {
 		return nil, err
 	}
 
-	// Might be limited by the storage account
-	// "" or default inherits the access type from the Storage Account
-	var containerAccessType azblob.PublicAccessType
+	serviceURL := fmt.Sprintf("%s/%s", config.Endpoint, config.ContainerName)
+	retryOpts := policy.RetryOptions{
```
Acconut, line 93: Are those retry options new?

pkg/azurestore/azureservice.go
```diff
@@ -67,87 +72,92 @@ type AzBlob interface {
...
 	case "cool":
-		blobAccessTierType = azblob.AccessTierCool
+		blobAccessTier = to.Ptr(blob.AccessTierCool)
 	case "hot":
-		blobAccessTierType = azblob.AccessTierHot
+		blobAccessTier = to.Ptr(blob.AccessTierHot)
 	case "":
 	default:
-		blobAccessTierType = azblob.DefaultAccessTier
+		blobAccessTier = nil
```
Acconut, line 135: Similar to above, the previous default was `azblob.DefaultAccessTier`. Can we keep that?

pkg/azurestore/azureservice.go
```diff
@@ -67,87 +72,92 @@ type AzBlob interface {
...
+		Blob:           blobClient,
+		Indexes:        []int{},
+		BlobAccessTier: service.BlobAccessTier,
+	}, nil
 }
 
 // Delete the blockBlob from Azure Blob Storage
 func (blockBlob *BlockBlob) Delete(ctx context.Context) error {
-	_, err := blockBlob.Blob.Delete(ctx, azblob.DeleteSnapshotsOptionInclude, azblob.BlobAccessConditions{})
+	_, err := blockBlob.Blob.Delete(ctx, nil)
```
Acconut, line 160: Should we keep `DeleteSnapshotsOptionInclude` or whatever its newer form is?

pkg/azurestore/azureservice.go
```diff
@@ -161,31 +171,22 @@ func (blockBlob *BlockBlob) Upload(ctx context.Context, body io.ReadSeeker) erro
...
-	downloadResponse, err := blockBlob.Blob.Download(ctx, 0, azblob.CountToEnd, azblob.BlobAccessConditions{}, false, azblob.ClientProvidedKeyOptions{})
-
-	// If the file does not exist, it will not return an error, but a 404 status and body
-	if downloadResponse != nil && downloadResponse.StatusCode() == 404 {
-		return nil, handler.ErrNotFound
-	}
+	resp, err := blockBlob.Blob.DownloadStream(ctx, nil)
 	if err != nil {
-		// This might occur when the blob is being uploaded, but a block list has not been committed yet
-		if isAzureError(err, "BlobNotFound") {
```
Acconut, line 182: Should we check for this error as well?

pkg/azurestore/azureservice.go
```diff
@@ -295,14 +290,17 @@ func blockIDIntToBase64(blockID int) string {
 	return blockIDBinaryToBase64(binaryBlockID)
 }
 
-func blockIDBase64ToInt(blockID string) int {
+func blockIDBase64ToInt(blockID *string) int {
 	blockIDBase64ToBinary(blockID)
 	return int(binary.LittleEndian.Uint32(blockIDBase64ToBinary(blockID)))
 }
 
-func isAzureError(err error, code string) bool {
```
Acconut, line 303: Could we keep this function and inspect errors from Azure by looking into their service code instead of just checking their error message? Using the service code appears more robust in the long term to me.

Acconut (commented):
Thank you very much for this! Overall, the PR is looking great and I'm positive we can merge this soon. I'm relieved that no fundamental changes are necessary to the azurestore itself and that most changes are just simple name changes.

pkg/azurestore/azureservice.go
xelat09: there is no such type anymore, it is just the default behaviour if createOptions is nil. I did refactor according to this

pkg/azurestore/azureservice.go
xelat09: yeah, that's ugly.. I tried to hard to not change too much :)
refactored this too

pkg/azurestore/azureservice.go
xelat09: similar to above, the default behaviour if options is nil, refactored

pkg/azurestore/azureservice.go
xelat09: yes, reintroduced that

pkg/azurestore/azureservice.go
xelat09: yes, sure.. i did that

xelat09:
thank you for the review, I'm still learning, so don't hesitate to point out anything that needs reworking

pkg/azurestore/azureservice.go
xelat09: yeah fixed that too

PR #1207, opened 2024-10-13
cli: Allow Application Default Credentials discovery for GCP

xytis:
When deploying tusd under GCP, it is better to use integrated Application Default Credentials discovery mechanism, that exists under most google SDK's.

This change simply allows _not_ providing any string as credential file, and then client library defaults to trying ADC as per https://cloud.google.com/docs/authentication/application-default-credentials 

If this proposal is acceptable, I could update the readme as well.

### DIFF
--- a/cmd/tusd/cli/composer.go
+++ b/cmd/tusd/cli/composer.go
@@ -76,8 +76,8 @@ func CreateComposer() {
 				"Please remove underscore from the value", Flags.GCSObjectPrefix)
 		}
 
-		// Legacy: account file used to be provided by GCS_SERVICE_ACCOUNT_FILE environment variable.
-		// Now it is more common to default into ADC discovery mechanism.
+		// Application Default Credentials discovery mechanism is attempted to fetch credentials,
+		// but an account file can be provided through the GCS_SERVICE_ACCOUNT_FILE environment variable.
 		gcsSAF := os.Getenv("GCS_SERVICE_ACCOUNT_FILE")
 
 		service, err := gcsstore.NewGCSService(gcsSAF)
--- a/cmd/tusd/cli/composer.go
+++ b/cmd/tusd/cli/composer.go
@@ -76,12 +76,9 @@ func CreateComposer() {
 				"Please remove underscore from the value", Flags.GCSObjectPrefix)
 		}
 
-		// Derivce credentials from service account file path passed in
-		// GCS_SERVICE_ACCOUNT_FILE environment variable.
+		// Legacy: account file used to be provided by GCS_SERVICE_ACCOUNT_FILE environment variable.
+		// Now it is more common to default into ADC discovery mechanism.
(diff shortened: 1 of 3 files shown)


Acconut:
The idea looks good to me, thank you for bringing this up.

What error does the SDK and thus tusd now return if no GCS_SERVICE_ACCOUNT_FILE was given and ADC failed as well? We must ensure that this still gives a descriptive error message.

> I could update the readme as well.

That would be great!

xytis:
Default error is okey-ish:
> 2024/10/22 07:10:01.750054 Attempting to use default GCS credentials because no GCS_SERVICE_ACCOUNT_FILE env var was provided
2024/10/22 07:10:01.802682 Unable to create Google Cloud Storage service: dialing: google: could not find default credentials. See https://cloud.google.com/docs/authentication/external/set-up-adc for more information

Acconut:
Yes, that seems acceptable, especially if our documentation covers authentication.

xytis:
Ok, updated the docs, plus I removed the warning message (since it now is explained in the docs).

docs/_storage-backends/google-cloud-storage.md
```diff
@@ -6,11 +6,20 @@ nav_order: 5
...
-To enable this backend, you must supply the path to the corresponding account file using environment variables and specify the bucket name using `-gcs-bucket`, for example:
+To enable this backend, you must specify the bucket name using `-gcs-bucket`, for example:
+
+```bash
+$ tusd -gcs-bucket=my-test-bucket.com
+[tusd] Using 'gcs://my-test-bucket.com' as GCS bucket for storage.
+...
+```
+
+By default, Application Default Credentials [discovery mechanism](https://cloud.google.com/docs/authentication/external/set-up-adc) will be attempted.
```
Acconut, line 21:
```suggestion
By default, [Application Default Credentials discovery mechanism](https://cloud.google.com/docs/authentication/external/set-up-adc) will be attempted.

```

cmd/tusd/cli/composer.go
```diff
@@ -76,12 +76,9 @@ func CreateComposer() {
 				"Please remove underscore from the value", Flags.GCSObjectPrefix)
 		}
 
-		// Derivce credentials from service account file path passed in
-		// GCS_SERVICE_ACCOUNT_FILE environment variable.
+		// Legacy: account file used to be provided by GCS_SERVICE_ACCOUNT_FILE environment variable.
+		// Now it is more common to default into ADC discovery mechanism.
```
Acconut, line 80:
```suggestion
		// Application Default Credentials discovery mechanism is attempted to fetch credentials,
		// but an account file can be provided through the GCS_SERVICE_ACCOUNT_FILE environment variable.
```

Acconut (approved):
Thank you for the contribution!

PR #1208, opened 2024-10-20
handler, s3store: Serve GET requests directly from S3

### LINKED ISSUE #1064: handler: support HTTP range requests in GET /:upload
Some discussion around API design in the quoted (since-closed) PR.

>> Out of interest, I did some searching and found #432 where you were against implementing `Range` support in tusd - have you changed your mind on this?

>Yes, I did change my opinion here. We have recently been involved in the HTTP working group with the goal of making a IETF standard for resumable uploads. In the end, we hope that resumable uploads are one component in the HTTP toolbox that interacts well with the existing HTTP mechanism. Since then I am more interested in making sure that tusd works better with concepts, such as range requests.

>> In any case, we should probably move this conversation to an issue to sketch out API design if you're in favour of this in principle - closing this PR for now.

>Great idea, feel free to open an issue if you are still interested in working on this :)

_Originally posted by @Acconut in https://github.com/tus/tusd/issues/1048#issuecomment-1881180105_

pcfreak30:
- Add ContentServerDataStore interface
- Update handlers to use ContentServerDataStore when available
- Implement range request handling for s3upload
- Add tests for new ContentServerDataStore functionality
- Update Go version to 1.22.1

Closes #1064

### DIFF
--- a/pkg/handler/composer.go
+++ b/pkg/handler/composer.go
@@ -14,6 +14,8 @@ type StoreComposer struct {
 	Concater           ConcaterDataStore
 	UsesLengthDeferrer bool
 	LengthDeferrer     LengthDeferrerDataStore
+	ContentServer      ContentServerDataStore
+	UsesContentServer  bool
 }
 
 // NewStoreComposer creates a new and empty store composer.
@@ -84,4 +86,8 @@ func (store *StoreComposer) UseConcater(ext ConcaterDataStore) {
 func (store *StoreComposer) UseLengthDeferrer(ext LengthDeferrerDataStore) {
 	store.UsesLengthDeferrer = ext != nil
 	store.LengthDeferrer = ext
+}
+func (store *StoreComposer) UseContentServer(ext ContentServerDataStore) {
+	store.UsesContentServer = ext != nil
+	store.ContentServer = ext
 }
--- a/pkg/handler/datastore.go
+++ b/pkg/handler/datastore.go
@@ -3,6 +3,7 @@ package handler
 import (
 	"context"
 	"io"
+	"net/http"
 )
 
 type MetaData map[string]string
@@ -119,6 +120,16 @@ type DataStore interface {
 	// GetUpload fetches the upload with a given ID. If no such upload can be found,
 	// ErrNotFound must be returned.
 	GetUpload(ctx context.Context, id string) (upload Upload, err error)
+}
+
+// ServableUpload defines the method for serving content directly
+type ServableUpload interface {
+	ServeContent(ctx context.Context, w http.ResponseWriter, r *http.Request) error
+}
+
+// ContentServerDataStore is the interface for data stores that can serve content directly
+type ContentServerDataStore interface {
+	AsServableUpload(upload Upload) ServableUpload
 }
 
 type TerminatableUpload interface {
--- a/pkg/handler/unrouted_handler.go
+++ b/pkg/handler/unrouted_handler.go
@@ -1047,6 +1047,17 @@ func (handler *UnroutedHandler) GetFile(w http.ResponseWriter, r *http.Request) 
 		return
 	}
 
+	// If the data store implements ContentServerDataStore, use the ServableUpload interface
+	if handler.composer.UsesContentServer {
+		servableUpload := handler.composer.ContentServer.AsServableUpload(upload)
+		err = servableUpload.ServeContent(c, w, r)
+		if err != nil {
+			handler.sendError(c, err)
+		}
+		return
+	}
+
+	// Fall back to the existing GetReader implementation if ContentServerDataStore is not implemented
 	contentType, contentDisposition := filterContentType(info)
 	resp := HTTPResponse{
 		StatusCode: http.StatusOK,
--- a/pkg/s3store/s3store.go
+++ b/pkg/s3store/s3store.go
@@ -79,6 +79,7 @@ import (
 	"net/http"
 	"os"
 	"regexp"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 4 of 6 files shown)


pcfreak30:
Please note I tried to avoid changing the go version, but `go mod tidy` seems to automatically do it.

This is a draft to get feedback.

@Acconut

pkg/s3store/s3store.go
```diff
@@ -100,6 +101,91 @@ import (
 // considered valid into a header value according to RFC2616.
 var nonPrintableRegexp = regexp.MustCompile(`[^\x09\x20-\x7E]`)
 
+var _ handler.ServerDataStore = &S3Store{}
+
+func (store S3Store) AsServableUpload(upload handler.Upload) handler.ServableUpload {
+	return &S3ServableUpload{
```
Acconut, line 107: If we implement `ServeContent` directly on `s3Upload`, we do not have to introduce another struct `S3ServableUpload`. This is already what we do for `AsTerminatableUpload`:

https://github.com/tus/tusd/blob/9779a84cae29d083b3496ba732fe7c76755bd4ed/pkg/s3store/s3store.go#L367-L369

pkg/s3store/s3store.go
```diff
@@ -100,6 +101,91 @@ import (
...
+		upload: upload.(*s3Upload),
+	}
+}
+
+type S3ServableUpload struct {
+	store  *S3Store
+	upload *s3Upload
+}
+
+func (su *S3ServableUpload) ServeContent(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
```
Acconut, line 118: Before merging, we should move these functions down in the file as they are not the most important ones. Maybe place the logic next to the other optional interfaces that S3Store implements.

pkg/s3store/s3store.go
```diff
@@ -1249,3 +1335,95 @@ func (store S3Store) releaseUploadSemaphore() {
 	store.uploadSemaphore.Release()
 	store.uploadSemaphoreDemandMetric.Dec()
 }
+
+func (store S3Store) copyToResponseWriter(w http.ResponseWriter, r io.Reader) (int64, error) {
```
Acconut, line 1339: Can you explain what the benefit of this is over `io.Copy`?

pkg/s3store/s3store.go
```diff
@@ -100,6 +101,91 @@ import (
...
+	w.Header().Set("Content-Length", strconv.FormatInt(info.Size, 10))
+	w.Header().Set("Content-Type", info.MetaData["filetype"])
+	w.Header().Set("ETag", *result.ETag)
+
+	// Stream the content
+	_, err = su.store.copyToResponseWriter(w, result.Body)
+	return err
+}
+
+func (su *S3ServableUpload) handleRangeRequest(ctx context.Context, w http.ResponseWriter, r *http.Request, info handler.FileInfo, input *s3.GetObjectInput, rangeHeader string) error {
```
Acconut, line 157: I'm wonder whether dedicated handling for range requests is necessary. If the request includes the Range header we can add it to `GetObject` without parsing the ranges and let S3 deal with it. Afterwards we just copy the Content-Range, Content-Length and ETag header from S3's response to tusd's response and copy the body.

The advantage is that we don't have to parse the header (and possibly introduce subtle bugs) and don't have to construct the Content-Range and Content-Length headers on our own.

We would also not need a dedicated `handleRangeRequest` function as this could all live in `ServeContent`.

What do you think?

pkg/handler/composer.go
```diff
@@ -14,6 +14,8 @@ type StoreComposer struct {
 	Concater           ConcaterDataStore
 	UsesLengthDeferrer bool
 	LengthDeferrer     LengthDeferrerDataStore
+	Server             ServerDataStore
```
Acconut, line 17: As mentioned in https://github.com/tus/tusd/issues/1064#issuecomment-2401568772, I am not sure on my own whether `Server` is a good name as it does have a different meaning in the HTTP-land. How about `ContentServer` as an alternative?

We would then have `StoreComposer.ContentServer`, `StoreComposer.UsesContentServer`, `StoreComposer.UseContentServer`, and `tusd.ContentServerDataStore`.

Acconut (commented):
Thank you very much for working on this! I left a few comments but the direction where this is heading is looking good.

pkg/s3store/s3store.go
pcfreak30: I can change this, though I did see value in it.

pkg/s3store/s3store.go
pcfreak30: :+1:

pkg/s3store/s3store.go
pcfreak30: TBH, I rely on AI for development and a lot of this I was able to complete due to its effort. The data copy and range logic is the most complicated parts, and as I did not have a good reason to remove them based on what I saw, I just ensured it was all valid.

If needed, I can replace with io.Copy and move on.

pkg/s3store/s3store.go
pcfreak30: I believe given that multiple ranges are possible, it made since to check it ourselves. and due to the complexity of the range checking, its why I had a unit test for the range logic made.

I can punt it all to be forwarded to s3, but I also think it is valid to verify at-least the ranges earlier in  the s3 impl.

pkg/handler/composer.go
pcfreak30: works for me, I don't really care :P.

pcfreak30:
@Acconut changes have been made. kudos.

pkg/handler/datastore.go
```diff
@@ -121,6 +122,16 @@ type DataStore interface {
 	GetUpload(ctx context.Context, id string) (upload Upload, err error)
 }
 
+// ServableUpload defines the method for serving content directly
+type ServableUpload interface {
+	ServeContent(ctx context.Context, w http.ResponseWriter, r *http.Request) error
```
Acconut, line 127: This function needs proper documentation. That's best done when we have agreed on what the handler does and does not.

pkg/handler/unrouted_handler.go
```diff
@@ -1013,6 +1013,17 @@ func (handler *UnroutedHandler) GetFile(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
+	// If the data store implements ContentServerDataStore, use the ServableUpload interface
+	if handler.composer.UsesContentServer {
+		servableUpload := handler.composer.ContentServer.AsServableUpload(upload)
+		err = servableUpload.ServeContent(c, w, r)
```
Acconut, line 1019: I wonder if the handler should set the Content-Length, Content-Type and Content-Disposition before passing it to `ServeContent`. The implementation can then decide to overwrite or keep them. The idea is that we already have logic for extracting the filename and filetype from metadata and this way we can keep using them.

pkg/s3store/s3store.go
Acconut: The goal of the new ServeContent interface is to reduce the complexity that we have to handle inside of tusd and adding logic to each store to validate the Range header is not having the opposite effect. In addition, deferring the validation to S3 (or whatever provider other stores use) is preferable since it knows best which values it will accept and we avoid discrepancies between tusd's validation and S3's.

pkg/s3store/s3store.go
```diff
@@ -376,6 +377,81 @@ func (store S3Store) AsConcatableUpload(upload handler.Upload) handler.Concatabl
...
+	result, err := su.store.Service.GetObject(ctx, input)
+	if err != nil {
+		return err
+	}
+	defer result.Body.Close()
+
+	// Set headers
+	w.Header().Set("Content-Length", strconv.FormatInt(info.Size, 10))
+	w.Header().Set("Content-Type", info.MetaData["filetype"])
+	w.Header().Set("ETag", *result.ETag)
```
Acconut, line 416: How about forwarding the Content-Disposition and Content-Encoding headers as well if they are present? I think that would be useful as well.

Acconut (changes requested):
Thank you for the updates! This is moving in the right direction and I feel like we're getting close :)

pkg/s3store/s3store.go
pcfreak30: done

pkg/handler/unrouted_handler.go
pcfreak30: you decide?

pkg/handler/datastore.go
pcfreak30: lmk what/when you want docs on this.

pcfreak30:
@Acconut changes have been made.

pkg/s3store/s3store.go
pcfreak30: done

pkg/handler/unrouted_handler.go
Acconut: Just implemented this by myself.

pcfreak30: Sorry, huh? where?

Acconut:
Thank you for the updates! I added some documentation and better handling for range requests and conditional requests, in particular regarding error responses. I couldn't push these changes to this PR because you didn't allow maintainers to modify your PR. So for now, I put them into a separate branch: https://github.com/tus/tusd/compare/main...s3-content-server

Feel free to review them. If you allow me to modify this PR, I can push them here. If not, we'll find another way to collaborate.

pcfreak30:
@Acconut sorry GitHub doesn't seem to be allowing me to give you access :/.

Acconut:
Ok, then I'll continue the work in my own PR: https://github.com/tus/tusd/pull/1228
Thank you for your help so far!

Acconut:
I just merged this as part of https://github.com/tus/tusd/pull/1228. Thank you!

PR #1209, opened 2024-10-21
handler: unwrap errors using `errors.As`

mcdoker18:
A datastore may wrap errors with additional information. Before the fix, the error was converted to the InternalServerError type by the handler.

### DIFF
--- a/pkg/handler/patch_test.go
+++ b/pkg/handler/patch_test.go
@@ -3,6 +3,7 @@ package handler_test
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/handler/patch_test.go
```diff
@@ -140,7 +141,10 @@ func TestPatch(t *testing.T) {
 	})
 
 	SubTest(t, "UploadNotFoundFail", func(t *testing.T, store *MockFullDataStore, composer *StoreComposer) {
-		store.EXPECT().GetUpload(gomock.Any(), "no").Return(nil, ErrNotFound)
+		// we wrap the error in order to ensure proper error handling
```
Acconut, line 144:
```suggestion
		// We wrap the error test whether the handler correctly unwraps it again
		// to get an handler.Error.
```

Acconut (approved):
Thank you!

PR #1214, opened 2024-11-11
filestore: Fix `.info` file not deleted if the binary file was successfully deleted

bajelidze:
This PR fixes a bug where if the bin file was deleted without errors in the filestore, the `.info` file would not be deleted.

The main issue is that the `!errors.Is(err, os.ErrNotExist)` check is being made without checking that `err` is actually not `nil`. The `nil` value will pass this `if` case and the subsequent cleanup of `.info` file will not happen in that case because the function returns at that point.

I checked the fix in my setup and now it actually removes the `.info` file as well.


Acconut (approved):
Thank you for this PR!

PR #1223, opened 2024-12-10
chore: update S3 storage to support Content-Disposition header

rinor:
- Add `Content-Disposition` header support in S3 uploads
- Rename `filterContentType` to `FilterContentType` (export)
- Update test cases for new header support
- Correct typo in function documentation
- Adjust metadata handling in S3Store uploads


rinor:
> What do you think? Does this make sense?

Indeed it does, closing this then.

Acconut:
Feel free to let me know if you are interested in helping with these improvements :)

rinor:
Do we still want to populate those `Content*` values from metadata (no need to update clients), or we want to have them as separate items, ie:
```go
// FileInfo contains information about a single upload resource.
type FileInfo struct {
    ....
    MetaData MetaData

    // Content-Disposition header
    ContentDisposition *string

    // Content-Encoding header
    ContentEncoding *string

    // Content-Language header
    ContentLanguage *string

    // Content-Type header
    ContentType *string
    .....
}
```

PR #1234, opened 2025-01-05
implement ContentServerDataStore for azure blob storage too

xelat09:
Hi & Happy new year

Implement ContentServerDataStore for azure blob storage too, tests with azurite+azure blob are ok.
Incomplete range requests (Range: bytes=100- ) are not supported, hope that's ok

Regards


pkg/azurestore/azureservice.go
```diff
@@ -187,6 +191,31 @@ func (blockBlob *BlockBlob) Download(ctx context.Context) (io.ReadCloser, error)
...
+	}
+
+	statusCode := http.StatusOK
+	if resp.ContentRange != nil {
+		// Use 206 Partial Content for range requests
+		statusCode = http.StatusPartialContent
+	} else if resp.ContentLength != nil && *resp.ContentLength == 0 {
+		statusCode = http.StatusNoContent
+	}
+	w.WriteHeader(statusCode)
```
Acconut, line 212: Can we forward response headers like Content-Type from Azure's response? Similar to how its done in the s3store: https://github.com/tus/tusd/blob/a50bc426ddd52a484dc476419c5bd77397ea05e6/pkg/s3store/serve_content.go#L95-L128

pkg/azurestore/azureservice.go
```diff
@@ -309,3 +343,37 @@ func checkForNotFoundError(err error) error {
...
+	if len(rangeParts) != 2 {
+		return nil, fmt.Errorf("invalid Range header format")
+	}
+
+	offset, err := strconv.ParseInt(rangeParts[0], 10, 64)
+	if err != nil {
+		return nil, fmt.Errorf("invalid offset in Range header")
+	}
+
+	count, err := strconv.ParseInt(rangeParts[1], 10, 64)
```
Acconut, line 370: You mentioned that "Incomplete range requests (Range: bytes=100- ) are not supported". From the SDK's documentation, it seems to be possible to let the count be zero and get the entire resource starting at the offset:

https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob@v1.5.0/internal/exported#HTTPRange

> An HTTPRange which has an offset and zero value count indicates from the offset to the resource's end. 

Could you implement support for this?

Acconut (changes requested):
Thank you very much for this PR, it's greatly appreciated! I had a brief first look and left a few comments. Let me know what you think!

xelat09:
yeah, as suggested, azure supports all those headers too. please have a look. tests with azurite do work.. 

do you see anything else missing?

pkg/azurestore/azureservice.go
```diff
@@ -67,6 +68,8 @@ type AzBlob interface {
 	Upload(ctx context.Context, body io.ReadSeeker) error
 	// Download returns a readcloser to download the contents of the blob
 	Download(ctx context.Context) (io.ReadCloser, error)
+    // Serves the contents of the blob directly handling special HTTP headers like Range, if set
+    ServeContent(ctx context.Context, w http.ResponseWriter, r *http.Request) error
```
Acconut, line 72: Can you please run `go fmt` to address the formatting inconsistencies?

pkg/azurestore/azureservice.go
```diff
@@ -253,6 +281,11 @@ func (infoBlob *InfoBlob) Download(ctx context.Context) (io.ReadCloser, error) {
 	return resp.Body, nil
 }
 
+// ServeContent is not needed for infoBlob
+func (infoBlob *InfoBlob) ServeContent(ctx context.Context, w http.ResponseWriter, r *http.Request) error {
+    return nil
```
Acconut, line 286:
```suggestion
    return errors.New("azurestore: ServeContent is not implemented for InfoBlob")
```

Please return an error here so we notice it early if this method is accidentally called.

pkg/azurestore/azureservice.go
```diff
@@ -187,6 +190,31 @@ func (blockBlob *BlockBlob) Download(ctx context.Context) (io.ReadCloser, error)
...
+    }
+
+    statusCode := http.StatusOK
+    if resp.ContentRange != nil {
+        // Use 206 Partial Content for range requests
+        statusCode = http.StatusPartialContent
+    } else if resp.ContentLength != nil && *resp.ContentLength == 0 {
+        statusCode = http.StatusNoContent
+    }
+    w.WriteHeader(statusCode)
```
Acconut, line 211: Can we forward response headers like Content-Type from Azure's response? Similar to how its done in the s3store: https://github.com/tus/tusd/blob/a50bc426ddd52a484dc476419c5bd77397ea05e6/pkg/s3store/serve_content.go#L95-L128

Acconut (changes requested):
Thank you very much for the quick updates! The handler is the request headers looks much better yet. Could you also have a look at the failing tests at https://github.com/tus/tusd/actions/runs/12736636713/job/35496766413?pr=1234? Talking about tests, it would be great to have some tests for ServeContent in azurestore as well. For inspiration you can use s3store's tests: https://github.com/tus/tusd/blob/main/pkg/s3store/serve_content_test.go

pkg/azurestore/azureservice.go
```diff
@@ -309,3 +376,47 @@ func checkForNotFoundError(err error) error {
...
 	return err
 }
+
+// parse the Range, If-Match, If-None-Match, If-Unmodified-Since, If-Modified-Since headers if present
+func ParseDownloadOptions(r *http.Request) (*azblob.DownloadStreamOptions, error) {
+	input := azblob.DownloadStreamOptions{AccessConditions: &azblob.AccessConditions{}}
+
+	if val := r.Header.Get("Range"); val != "" {
+		// zero value count indicates from the offset to the resource's end, suffix-length is not required
+		input.Range = azblob.HTTPRange{Offset: 0, Count: 0}
```
Acconut, line 386: Does Azure support fetching multiple ranges in a single request? I.e. using `Range: <unit>=<range-start>-<range-end>, …, <range-startN>-<range-endN>` (from https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Range). S3 doesn't so the s3store doesn't support it. We can do the same here, but I was just wondering.

pkg/azurestore/azureservice.go
```diff
@@ -309,3 +376,47 @@ func checkForNotFoundError(err error) error {
...
+		bytesEnd := 0
+		if _, err := fmt.Sscanf(val, "bytes=%d-%d", &input.Range.Offset, &bytesEnd); err != nil {
+			if _, err := fmt.Sscanf(val, "bytes=%d-", &input.Range.Offset); err != nil {
```
Acconut, line 389: This doesn't seem to be handling requests where only the last bytes are requested using `Range: <unit>=-<suffix-length>`. Is that supported by Azure?

Acconut (changes requested):
Hello there, I just had another look over this PR. Apologies for missing it earlier. The PR seems mostly finished. I think the handling of Range header could be improved a bit and then there is still this comment: https://github.com/tus/tusd/pull/1234/files#r1905384839

Let me know if you are still interested in completing this PR. If not, I can consider pulling it across the finish line.

pkg/azurestore/azureservice.go
xelat09: done that see azureservice.go:230+

pkg/azurestore/azureservice.go
xelat09: that isn't supported either. how should this be handled? throw an error/map to a different httprange?

It would be nice, if you could pull it across the finish line, I'm willing to do the testing :)

Acconut: Ok, I see that the Azure SDK doesn't support this natively: https://github.com/Azure/azure-sdk-for-go/blob/f29a1d52c4e6894a536b0da18ac4399692e02c4c/sdk/storage/azblob/internal/exported/exported.go#L23

However, I think the corresponding `AzUpload` might have the total size of the object already in its info object. So we could calculate the corresponding offset based on the size as `offset = size - suffix-length` (and maybe -1 depending on how the bytes are counted, not sure).

I don't want to drag this on too long, but if there is an easy way to support fetching trailing bytes, I think it would be great to have it :)

pkg/azurestore/azureservice.go
Acconut: > Regarding multiple range in a single request, that is not supported. see also https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob@v1.6.1/internal/exported#HTTPRange

Ok, I see. Can you please add a comment mentioning that Azure doesn't support multiple ranges and thus azurestore falls back to fetching the full resource, just like s3store does?

Acconut:
Superseeded by https://github.com/tus/tusd/pull/1381.

Issue #1236, opened 2025-01-13
Pass request context to the logger


mmhand123:
**Is your feature request related to a problem? Please describe.**
We're trying to integrate handling errors as they come up with sentry and am trying to capture exceptions any time errors come up in TUS. What I've landed on so far, which might not be the best way, is to implement a custom handler for the logger I pass to TUS so that whenever an error log comes in, it captures an exception to Sentry. This works but is missing some request specific context that gets setup when doing routing. 

It would be great if logging had access to the request context here.

**Describe the solution you'd like**
Ideally the logger would have access to the request context via `c.log.ErrorCtx` (or `InfoCtx` etc). This way my custom handler can pull out the sentry information from the request context.

**Describe alternatives you've considered**
Right now it's just capturing generic exceptions, which somewhat works, but is missing important information. Not sure of other ways to achieve what we want but open to any suggestions!

**Can you provide help with implementing this feature?**
Yes, I can provide a PR if it seems to be a good solution!

PR #1237, opened 2025-01-14
handler: Pass request context to logger when logging

mmhand123:
This allows the logger passed to the tusd handler to access the request specific context and pull out things like sentry information or trace information. Implements #1236

### DIFF
--- a/pkg/handler/unrouted_handler.go
+++ b/pkg/handler/unrouted_handler.go
@@ -177,10 +177,10 @@ func (handler *UnroutedHandler) Middleware(h http.Handler) http.Handler {
 		// We also update the write deadline, but makes sure that it is larger than the read deadline, so we
 		// can still write a response in the case of a read timeout.
 		if err := c.resC.SetReadDeadline(time.Now().Add(handler.config.NetworkTimeout)); err != nil {
-			c.log.Warn("NetworkControlError", "error", err)
+			c.log.WarnCtx(c, "NetworkControlError", "error", err)
 		}
 		if err := c.resC.SetWriteDeadline(time.Now().Add(2 * handler.config.NetworkTimeout)); err != nil {
-			c.log.Warn("NetworkControlError", "error", err)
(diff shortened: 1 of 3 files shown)


pkg/handler/unrouted_handler.go
```diff
@@ -177,10 +177,10 @@ func (handler *UnroutedHandler) Middleware(h http.Handler) http.Handler {
 		// We also update the write deadline, but makes sure that it is larger than the read deadline, so we
 		// can still write a response in the case of a read timeout.
 		if err := c.resC.SetReadDeadline(time.Now().Add(handler.config.NetworkTimeout)); err != nil {
-			c.log.Warn("NetworkControlError", "error", err)
+			c.log.WarnCtx(c, "NetworkControlError", "error", err)
```
Acconut, line 180: Can you please use methods like `WarnContext` instead of `WarnCtx`? The latter ones are deprecated (see https://pkg.go.dev/golang.org/x/exp/slog#Logger.WarnCtx) and not part of the `log/slog` package in the standard library. In the next major release, we want to migrate from `golang.org/x/exp/slog` to `log/slog`, so this would be helpful.

Acconut (changes requested):
Thank you very much for this quick PR. One minor note, then we can merge this.

pkg/handler/unrouted_handler.go
mmhand123: Yeah! Sorry about that - realize you said it in the issue, didn't realize that the `exp/slog` package had an update to provide it (earlier version only had `Ctx` variants). Updated in [3e95b11](https://github.com/tus/tusd/pull/1237/commits/3e95b1179960437a061ae0dc49551bd05183728c)!

Acconut: Ah, no worries! I didn't know what the currently used version of `slog` didn't have the newly named methods yet!

Acconut (approved):
Thank you very much for your work!

Issue #1238, opened 2025-01-15
When using filestore as the storage backend, why are chunk files not closed during the merging phase?


nervtech:
Hi,

I was reviewing the code and noticed that in the file merge process, the file opened with os.Open is not explicitly closed.

https://github.com/tus/tusd/blob/d69efe04584c858703f7b99921714cdd4291c65c/pkg/filestore/filestore.go#L242-L253

From my understanding, not closing the file could lead to resource leaks, especially with file descriptors.

Is there a specific reason why src.Close() is omitted here?

Acconut:
Thank you for pointing this out! You are right that this function is missing calls to `src.Close`. This seems to be a mistake in my eyes and not serving any purpose. Would you be interested in opening a PR to fix this?

PR #1239, opened 2025-01-16
filestore: Close files correctly when concatenating uploads

nervtech:
Provided one solution to [issue #1238](https://github.com/tus/tusd/issues/1238)


Acconut (approved):
Thank you very much for this PR! I pushed two additional changes and now this is ready to be merged.

Issue #1194, opened 2024-09-25
opus files are downloaded as octet-stream


masewo:
**Describe the bug**
Files with content type `audio/ogg; codecs=opus` are downloaded as `application/octet-stream`.
The content type does not pass the regex check in the `unrounted_handler.go`.

**To Reproduce**
1. Upload an opus file
2. Download this file
3. See content type

**Expected behavior**
The content type is `audio/ogg; codecs=opus`.

**Setup details**
Please provide following details, if applicable to your situation:
- Operating System: Windows
- Used tusd version: 2.5.0
- Used tusd data storage: disk store
- Used tusd configuration: Go
- Used tus client library: Uppy.js

Acconut:
That's correct. Additional parameters in the media type are currently not allowed. Would you be interested in a PR for extending the regular expression to allow the syntax specified in https://httpwg.org/specs/rfc9110.html#media.type?

Alternatively, we could also consider moving away from the regular expression entirely and instead use Go's builtin media type parser: https://pkg.go.dev/mime#ParseMediaType. The benefit is that we do not have to maintain a complex regular expression that might be incompatible with the actual specs.

If the parsing fails, we would use `application/octet-stream`. Otherwise, we set `Content-Disposition` depending on whether the media type is allowed to be displayed inline or not.

What do you think?

Acconut:
> I did not have good experiences recently with go's built-in mime package since it relies on information of the OS where it is installed on (causing issue for me on Alpine systems).

That's surprising to me as I would expect media parsing to not load OS resources since it's just parsing a string. However, I haven't tested this on my own yet. Nevertheless, I would like to explore that route more as I also mentioned in a comment in your PR.

Acconut:
Fixed in https://github.com/tus/tusd/pull/1253.

PR #1253, opened 2025-03-07
handler: Allow media type parameters in Content-Type headers

masewo:
This pull request includes several changes to the `pkg/handler` package, focusing on improving the handling of MIME types and adding new tests. The most important changes include fixing a regular expression, updating comments, modifying the MIME type whitelist, and adding new unit tests.

Improvements to MIME type handling:

* [`pkg/handler/unrouted_handler.go`](diffhunk://#diff-470f9eeff1025160ca44d324fc316e0ed2ebeb9da748e43c9dd30dfafe01d132L37-R37): Fixed a regular expression in `reMimeType` to correctly match MIME types with a plus sign.
* [`pkg/handler/unrouted_handler.go`](diffhunk://#diff-470f9eeff1025160ca44d324fc316e0ed2ebeb9da748e43c9dd30dfafe01d132L1115-R1117): Updated comments to correct typos and improve clarity.
* [`pkg/handler/unrouted_handler.go`](diffhunk://#diff-470f9eeff1025160ca44d324fc316e0ed2ebeb9da748e43c9dd30dfafe01d132L1133-R1161): Added `video/mp4` to the `mimeInlineBrowserWhitelist` and reordered entries for better readability.

Testing enhancements:

* [`pkg/handler/unrouted_handler_test.go`](diffhunk://#diff-72046bebba6a53220e9ec3f8311cdefac4b024d1d44f765dbe9675b6d3c8db77L1-L8): Changed the package name from `handler_test` to `handler` to allow testing of unexported functions.
* [`pkg/handler/unrouted_handler_test.go`](diffhunk://#diff-72046bebba6a53220e9ec3f8311cdefac4b024d1d44f765dbe9675b6d3c8db77R34-R91): Added a new test function `TestFilterContentType` to ensure the `filterContentType` function behaves correctly with various metadata inputs.

### DIFF
--- a/pkg/handler/unrouted_handler.go
+++ b/pkg/handler/unrouted_handler.go
@@ -34,7 +34,6 @@ const (
 var (
 	reForwardedHost  = regexp.MustCompile(`host="?([^;"]+)`)
 	reForwardedProto = regexp.MustCompile(`proto=(https?)`)
-	reMimeType       = regexp.MustCompile(`^[a-z]+\/[a-z0-9\-\+\.]+$`)
 	// We only allow certain URL-safe characters in upload IDs. URL-safe in this means
 	// that their are allowed in a URI's path component according to RFC 3986.
 	// See https://datatracker.ietf.org/doc/html/rfc3986#section-3.3
@@ -1112,9 +1111,9 @@ func (handler *UnroutedHandler) GetFile(w http.ResponseWriter, r *http.Request) 
 // mimeInlineBrowserWhitelist is a map containing MIME types which should be
 // allowed to be rendered by browser inline, instead of being forced to be
 // downloaded. For example, HTML or SVG files are not allowed, since they may
-// contain malicious JavaScript. In a similiar fashion PDF is not on this list
+// contain malicious JavaScript. In a similar fashion, PDF is not on this list
 // as their parsers commonly contain vulnerabilities which can be exploited.
-// The values of this map does not convey any meaning and are therefore just
+// The values of this map do not convey any meaning and are therefore just
 // empty structs.
 var mimeInlineBrowserWhitelist = map[string]struct{}{
 	"text/plain": {},
@@ -1125,14 +1124,17 @@ var mimeInlineBrowserWhitelist = map[string]struct{}{
 	"image/bmp":  {},
 	"image/webp": {},
 
-	"audio/wave":      {},
-	"audio/wav":       {},
-	"audio/x-wav":     {},
-	"audio/x-pn-wav":  {},
-	"audio/webm":      {},
-	"video/webm":      {},
-	"audio/ogg":       {},
-	"video/ogg":       {},
+	"audio/wave":     {},
+	"audio/wav":      {},
+	"audio/x-wav":    {},
+	"audio/x-pn-wav": {},
+	"audio/webm":     {},
+	"audio/ogg":      {},
+
+	"video/mp4":  {},
+	"video/webm": {},
+	"video/ogg":  {},
+
 	"application/ogg": {},
 }
 
@@ -1140,23 +1142,22 @@ var mimeInlineBrowserWhitelist = map[string]struct{}{
 // Content-Disposition headers for a given upload. These values should be used
 // in responses for GET requests to ensure that only non-malicious file types
 // are shown directly in the browser. It will extract the file name and type
-// from the "fileame" and "filetype".
+// from the "filename" and "filetype".
 // See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition
 func filterContentType(info FileInfo) (contentType string, contentDisposition string) {
 	filetype := info.MetaData["filetype"]
 
-	if reMimeType.MatchString(filetype) {
-		// If the filetype from metadata is well formed, we forward use this
-		// for the Content-Type header. However, only whitelisted mime types
-		// will be allowed to be shown inline in the browser
+	if ft, _, err := mime.ParseMediaType(filetype); err == nil {
+		// If the filetype from metadata is well-formed, we forward use this for the Content-Type header.
+		// However, only allowlisted mime types	will be allowed to be shown inline in the browser
 		contentType = filetype
-		if _, isWhitelisted := mimeInlineBrowserWhitelist[filetype]; isWhitelisted {
+		if _, isWhitelisted := mimeInlineBrowserWhitelist[ft]; isWhitelisted {
 			contentDisposition = "inline"
 		} else {
 			contentDisposition = "attachment"
 		}
 	} else {
-		// If the filetype from the metadata is not well formed, we use a
+		// If the filetype from the metadata is not well-formed, we use a
 		// default type and force the browser to download the content.
 		contentType = "application/octet-stream"
 		contentDisposition = "attachment"
--- a/pkg/handler/get_test.go
+++ b/pkg/handler/get_test.go
@@ -110,7 +110,7 @@ func TestGet(t *testing.T) {
 			upload.EXPECT().GetInfo(gomock.Any()).Return(FileInfo{
 				Offset: 0,
 				MetaData: map[string]string{
-					"filetype": "non-a-valid-mime-type",
+					"filetype": "non a valid mime type",
 				},
 			}, nil),
 		)
--- a/pkg/handler/unrouted_handler_test.go
+++ b/pkg/handler/unrouted_handler_test.go
@@ -1,11 +1,9 @@
-package handler_test
+package handler
 
 import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-
-	. "github.com/tus/tusd/v2/pkg/handler"
 )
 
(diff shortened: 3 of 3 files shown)


pkg/handler/unrouted_handler.go
```diff
@@ -34,7 +34,7 @@ const (
 var (
 	reForwardedHost  = regexp.MustCompile(`host="?([^;"]+)`)
 	reForwardedProto = regexp.MustCompile(`proto=(https?)`)
-	reMimeType       = regexp.MustCompile(`^[a-z]+\/[a-z0-9\-\+\.]+$`)
+	reMimeType       = regexp.MustCompile(`^(?:application|audio|example|font|haptics|image|message|model|multipart|text|video|x-(?:[0-9A-Za-z!#$%&'*+.^_` + "`" + `|~-]+))\/([0-9A-Za-z!#$%&'*+.^_` + "`" + `|~-]+)((?:[ \t]*;[ \t]*[0-9A-Za-z!#$%&'*+.^_` + "`" + `|~-]+=(?:[0-9A-Za-z!#$%&'*+.^_` + "`" + `|~-]+|"(?:[^"\\]|\.)*"))*)$`)
```
Acconut, line 37: Thank you very much for this PR! I'm a bit worried about this regular expression as I'm not able to understand it due to its complexity. I would be in favor of exploring solutions using Go's builtin media type parser (see https://github.com/tus/tusd/issues/1194#issuecomment-2375989032) to reduce the ease of maintenance for us.

masewo: Go's builtin media type parser (https://pkg.go.dev/mime#ParseMediaType) only **parses** the media type, which means it separates _type_ and _subtype_ from the optional parameters (as defined in RFC 6838).
For example `image_png; foo=bar` becomes `image_png` with parameters `foo -> bar` but it does not check if `image_png` is a valid media types known by IANA (which it's not).

masewo: We could:
- Split the filetype at the `/` and check if both parts are not empty (really simple check).
- Find a simpler regex.
- Use Go's builtin `TypeByExtension` (https://pkg.go.dev/mime#TypeByExtension), with which we could check if the filename's extension matches with our filetype.
  - It has a very small buildin table.
  - It relies on information available in the OS (on Alpine for example you would have to install the package `mailcap`).
  - What do we do if we have no filename?
- Use external Go dependencies like filetype (https://github.com/h2non/filetype) or mimetype (https://github.com/gabriel-vasile/mimetype)

masewo: I updated the PR using Go's buildin `ExtensionsByType`, also adding `mailcap` to the `Dockerfile`.

Acconut: I think you are misunderstand what this code is supposed to do and what it shouldn't do. Tusd itself does not care whether a media type has been registered with the IANA or not. Checking that is not tusd's responsibility. If you want to restrict file uploads to specific media types depending your application, the logic has to be implemented in the `pre-create` hook. Tusd will accept whatever media type `pre-create` allows.

The only way in which tusd inspects the media type is to determine whether it can instruct browsers to render the file content or force a download. `image/png` can be safely rendered, for example, while `text/html` cannot as it leads the door to XSS vulnerabilities. Hence, tusd only allows inline rendering via Content-Disposition for a list of selected media types. If a file with an unknown media type is encountered, tusd will still forward the media type to the client, but just instruct browsers to not render it directly.

As far as I understand, your original problem is that `video/ogg` files with parameters in the media type as served with a Content-Disposition header field that instructs browser to download the video instead of showing it directly. That's because the current regular expression for parsing media types does not support parameters. That's why I suggest replacing this limiting regular expression with Go's built-in parser which handles parameters correctly.

> Go's builtin media type parser (https://pkg.go.dev/mime#ParseMediaType) only **parses** the media type, which means it separates _type_ and _subtype_ from the optional parameters (as defined in RFC 6838).

Yes, you are correct. But tusd only needs to parse the media type and check it against an internal list to determine the Content-Disposition header. It doesn't have to determine whether the media type is registered with the IANA (or any other registry).

I hope this helps.

masewo: @Acconut I think I already understood it, since there is a good amount of comments explaining what the code is supposed to do, but I thought you would like to change the behaviour because of the link to https://github.com/tus/tus-node-server/pull/655

I simplified the PR again. It uses the ParseMediaType now to get rid of the extensions. Hope this suits now more your goals.

pkg/handler/unrouted_handler.go
```diff
@@ -1125,38 +1125,40 @@ var mimeInlineBrowserWhitelist = map[string]struct{}{
...
+// from the "filename" and "filetype".
 // See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition
 func filterContentType(info FileInfo) (contentType string, contentDisposition string) {
 	filetype := info.MetaData["filetype"]
 
-	if reMimeType.MatchString(filetype) {
-		// If the filetype from metadata is well formed, we forward use this
-		// for the Content-Type header. However, only whitelisted mime types
-		// will be allowed to be shown inline in the browser
+	if ft, _, err := mime.ParseMediaType(filetype); err == nil && reMimeType.MatchString(ft) {
```
Acconut, line 1151:
```suggestion
	if ft, _, err := mime.ParseMediaType(filetype); err == nil {
```

The regular expression is not needed anymore if the parsing happens using `ParseMediaType`.

pkg/handler/unrouted_handler.go
```diff
@@ -1125,38 +1125,40 @@ var mimeInlineBrowserWhitelist = map[string]struct{}{
...
+		// If the filetype from metadata is well-formed, we forward use this for the Content-Type header.
+		// However, only allowlisted mime types	will be allowed to be shown inline in the browser
 		contentType = filetype
 		if _, isWhitelisted := mimeInlineBrowserWhitelist[filetype]; isWhitelisted {
```
Acconut, line 1155:
```suggestion
		if _, isWhitelisted := mimeInlineBrowserWhitelist[ft]; isWhitelisted {
```

This should now check against the extracted media type extracted and not the entire value. Otherwise media types with parameters will have an incorrect Content-Disposition header.

Acconut (changes requested):
Great to hear that we are now on the same page!

pkg/handler/unrouted_handler.go
masewo: Right.

pkg/handler/unrouted_handler.go
masewo: Then weird filetypes like `image_png` would pass the check and not be transformed to `application/octet-stream`, what the regex prohibited before.

masewo: Done. Was an unrealistic case anyway. Hope we do not have too many unit tests now.

Acconut: You're right, such media types will be included in responses now. They can be rejected through pre-create hooks though, so we should be fine.

Acconut:
Can you please have a look at the failing tests?

masewo:
`non-a-valid-mime-type` is now a valid mime type. I changed it to `non a valid mime type`.

masewo:
@Acconut Seems like a flaky e2e test. The `TestUnexpectedNetworkReset` sometimes fails on the Windows platform because the PATCH and HEAD requests happen to fast after each other every 10th run or so, causing the file locker not unlock fast enough.

Acconut (approved):
Thank you for your contribution!

Acconut:
> @Acconut Seems like a flaky e2e test. The `TestUnexpectedNetworkReset` sometimes fails on Windows, because the PATCH and HEAD requests happen too fast after each other, causing the file locker not unlock the lock file fast enough. Happens for me in every 10th run or so.

Thank you for looking into the failing e2e although it's not caused by your PR here. I have also noticed that it randomly fails on Windows, while working fine on Linux. It seems to be caused by minor differences in Go's or the OS' networking stack on these platforms.

Adding a short delay between the PATCH and HEAD request is fine. The test is about ensuring that tusd saves data even when the TCP connection is reset. This goal will not be compromised by a short delay.

That being said, it seems like tusd on Windows is returning internal server errors where it actually should return ErrFileLocked:

>  level=ERROR event=InternalServerError method=HEAD path=b31ad07cb800343739b5b9ad38a0e720 requestId="" id=b31ad07cb800343739b5b9ad38a0e720 message="open D:\\a\\tusd\\tusd\\internal\\e2e\\data\\b31ad07cb800343739b5b9ad38a0e720.lock: Access is denied."

and 

> event=InternalServerError method=HEAD path=b2fd0bd224ede04083b048e03372db84 requestId="" id=b2fd0bd224ede04083b048e03372db84 message="open D:\\a\\tusd\\tusd\\internal\\e2e\\data\\b2fd0bd224ede04083b048e03372db84.lock: The process cannot access the file because it is being used by another process."

Feel free to open another PR for the delay, if you are interested

Issue #1043, opened 2023-12-11
Add env vars to configuration the tusd service


halabooda:
**Is your feature request related to a problem? Please describe.**
I using golang-plugins for external behaviours and using internal abstractions of tusd (composer, store) then I would like configuring main tusd and my plugins via common var envs, but tusd use only flags.

**Describe the solution you'd like**
I propose to use env tags. For example:
```
S3Bucket   string `envconfig:"S3_BUCKET"  required:"false"`
S3Endpoint string `envconfig:"S3_ENDPOINT"  required:"false"`
Basepath   string `envconfig:"BASE_PATH"  required:"false"`
HttpHost   string `envconfig:"HTTP_HOST"  required:"false"`
HttpPort   string `envconfig:"HTTP_PORT"  required:"false"`
```

**Can you provide help with implementing this feature?**
Yes. I can add envconfig tags (or an alternative lib) and rename the config abstraction from "flags" to config for use flags and envs together

Acconut:
I am open to adding this as it seems useful in certain situations. We should probably just discuss which library to use for this. A quick search also revealed https://github.com/jnovack/flag, which is advertised as a drop-in replacement for Go's `flag` package. Since we currently use `flag`, this would mean minimal changes to our code. Do you have other recommendations?

Maksclub:
@Acconut thanks, https://github.com/jnovack/flag seems good compromise!

Acconut:
Would you be interested in trying it out? :)

Acconut:
From https://github.com/tus/tusd/pull/1255#issuecomment-2775234331:

> I'm even thinking about future extensions to tusd where users could configure multiple data storages an the pre-create hook selects one of them. Tusd would allow this configuration by repeating storage-related flags, such as:
> 
> ```
> tusd --storage-name s3-a --s3-bucket bucketA --storage-name s3-b --s3-bucket bucketB --storage-name disk --dir /srv/storage/
> ```
> 
> Such a concept is harder to map to environment variables and thus I am unsure if that's even a path we should head into.

I think for now, it's best to close this until we have concrete proposals and plans on how we want to allow tusd to be configured.

PR #1255, opened 2025-03-22
Use github.com/jnovack/flag for envvar configs

nikitos:
https://github.com/tus/tusd/issues/1043

### DIFF
--- a/cmd/tusd/cli/flags.go
+++ b/cmd/tusd/cli/flags.go
@@ -1,11 +1,11 @@
 package cli
 
 import (
-	"flag"
 	"path/filepath"
 	"strings"
 	"time"
 
+	"github.com/jnovack/flag"
 	"github.com/tus/tusd/v2/internal/grouped_flags"
 	"github.com/tus/tusd/v2/pkg/hooks"
 	"golang.org/x/exp/slices"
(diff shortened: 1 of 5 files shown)


Acconut:
Thank you for this PR! I haven't looked at the package in question in detail, but I am a bit worried since its last release was four years ago. I realize that some software can be complete and doesn't have to change, but I wondering if this might bite us in the future if we want to use newer features in Go's `flag` which are not ported to this package.

nikitos:
you suggested it yourself) i linked related issue) i can try something else if you have alternates

Acconut:
Yes, you are right in that I originally pointed to that package. However, to be frank, I didn't thoroughly review the package at the time and I don't have any alternative.

I'm even thinking about future extensions to tusd where users could configure multiple data storages an the `pre-create` hook selects one of them. Tusd would allow this configuration by repeating storage-related flags, such as:

```
tusd --storage-name s3-a --s3-bucket bucketA --storage-name s3-b --s3-bucket bucketB --storage-name disk --dir /srv/storage/
```

Such a concept is harder to map to environment variables and thus I am unsure if that's even a path we should head into.

nikitos:
ok, i understand, so decline or request changes on this PR when you be sure about this feature

Issue #1016, opened 2023-10-04
Store the Mimetype on storage backends supporting it


meidlinga:
**Question**
I would like to include the uploaded files mime type in the upload meta data.
Then I would like to store this information along the uploaded file on Azure storage.
On a quick glance over the code, I could not find this feature.
Would this currently be possible in any way?

**Setup details**
Please provide following details, if applicable to your situation:
- Operating System: Linux
- Used tusd version: --
- Used tusd data storage: Azure
- Used tusd configuration: --
- Used tus client library: --

Acconut:
A similar question also appeared for S3 in https://github.com/tus/tusd/issues/978. I think it would be worth adding this to storages, which provide some mechanism for saving the file name and file type in a dedicated manner. Does Azure have such a functionality?

meidlinga:
Yes, Azure supports it by adding a specific header in the upload request. It is `x-ms-blob-content-type`, documented here: https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob?tabs=azure-ad#request-headers-all-blob-types

Acconut:
That's good to know. It is also very similar to the `Content-Type` header that is used for AWS S3 to set the file type for an object: https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateMultipartUpload.html#API_CreateMultipartUpload_RequestSyntax

It should not be too hard to set those two values when creating an upload. tusd currently inspects the `filetype` meta data entry and sets `Content-Disposition` accordingly when a user downloads a file via tusd. We can also read this value and pass it to Azure, S3 etc.

Acconut:
It would be great if tusd could set Content-Type, Content-Encoding, and Content-Disposition for storages that support it.

quality-leftovers:
Looking at the code (as of today) it seems s3store supports setting the blob `Content-Type` using either
-  `Content-Type` header when using TUS V2 draft protocol
- `filetype` metadata field using TUS V1 protocol

The s3store also attaches metadata to the blob while azurestore just ignores it.

So I did play around a bit with the azurestore. See: https://github.com/quality-leftovers/tusd/commit/4909c02a5096fc47d353b4499dd65f4cd86c7661

If there is interest in getting this into main I'll try to find some time to work on an actual PR.

Acconut:
> Looking at the code (as of today) it seems s3store supports setting the blob `Content-Type`

[...]

PR #1268, opened 2025-05-05
azurestore: Support blob metadata and Content-Type

quality-leftovers:
azurestore: Support blob metadata and Content-Type

### DIFF
--- a/pkg/azurestore/azurestore.go
+++ b/pkg/azurestore/azurestore.go
@@ -9,11 +9,23 @@ import (
 	"io"
 	"io/fs"
 	"os"
+	"regexp"
 	"strings"
 
 	"github.com/tus/tusd/v2/internal/uid"
 	"github.com/tus/tusd/v2/pkg/handler"
 )
+
+// Handler that assigns blob metadata based on file info
+type AssignBlobMetadataFunc func(handler.FileInfo) (handler.MetaData, error)
+
+// This regular expression matches every character which is not
+// considered valid into a header value according to RFC2616.
+var invalidMetadataValueCharsRegexp = regexp.MustCompile(`[^\x09\x20-\x7E]`)
+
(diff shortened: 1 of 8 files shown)


quality-leftovers:
Solves #1016 for azure

pkg/azurestore/azurestore.go
```diff
@@ -258,3 +315,42 @@ func (store *AzureStore) keyWithPrefix(key string) string {
...
+	} else {
+		return assignBlobMetadata
+	}
+}
+
+// The default metadata handler is very permissive to ensure we do not break backwards compatibility.
+// The only realistic failure scenario should be exceeding the metadata size limit.
+func assignBlobMetadata(fileinfo handler.FileInfo) (handler.MetaData, error) {
+	result := make(map[string]string)
+	for key, value := range fileinfo.MetaData {
```
Acconut, line 335: As explained in https://github.com/tus/tusd/issues/1016#issuecomment-2841519268, I would like storage to not blindly pass all upload metadata values to the storage as these values are not sanitized ot validated. Instead, pre-create hooks should be used to control which metadata values should be attached to the object on Azure, S3 etc. The object metadata would then be read from `FileInfo.Storage["MetaData"]`. The need for sanitization and a setting to control whether metadata is forwarded would then vanish as these tasks are performed by the pre-create hook to suite the application-specific needs.

Currently, the s3store does not exhibit this behavior, but I would like to change that in the next major release. If you want to, we can defer the work here until I have adjusted the s3store and we can replicate it here.

Does my thinking make sense to you?

quality-leftovers: I think this would be the best course of actions. 

Let's close this PR and I'll re-open a new one once I can use the s3store implementation as reference for the azurestore.

quality-leftovers:
Defer until s3store implements new metadata handling.

PR #1272, opened 2025-05-21
azurestore: Fix the GetOffset to work with enabled Azure blob versioning

MariuszBartnik:
azurestore: Fix the GetOffset to work with enabled Azure blob versioning


Acconut (approved):
I'm not an Azure user on my own and can't test this, but it appears sensible that only uncommitted blocks are considered for the upload. Let's give this a try and revert if users run into problems. Thank you!

PR #1280, opened 2025-06-09
filestore: Expose keys used in storage map

wylswz:
Added & exposed some consts, so that custom hook implementations can use them

### DIFF
--- a/pkg/filestore/filestore.go
+++ b/pkg/filestore/filestore.go
@@ -30,7 +30,9 @@ var defaultFilePerm = os.FileMode(0664)
 var defaultDirectoryPerm = os.FileMode(0754)
 
 const (
-	StorageKeyPath     = "Path"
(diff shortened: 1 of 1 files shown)


pkg/filestore/filestore.go
```diff
@@ -29,6 +29,11 @@ import (
 var defaultFilePerm = os.FileMode(0664)
 var defaultDirectoryPerm = os.FileMode(0754)
 
+const (
+	StorageKeyPath     = "Path"
+	StorageKeyInfoPath = "InfoPath"
+)
```
Acconut, line 35: Can you please add appropriate comments for documentation?

Acconut (changes requested):
Sounds like a sensible thing. Some documentation would be helpful, though.

Acconut (approved):
Thank you

Issue #343, opened 2020-01-08
GCS Rate limit exceeded.


PokeGuys:
**Describe the bug**
Updating `.info` file too frequently (1 request per second) triggered [GCS object rate limit](https://cloud.google.com/storage/docs/key-terms#immutability).

**To Reproduce**
1. Start upload file faster than 1 request per second.
2. See error
```
googleapi: Error 429: The rate of change requests to the object storage/8f3ffb04745ec523b4689ce083ee868d.info exceeds the rate limit. Please reduce the rate of create, update, and delete requests.
```

**Expected behavior**
Adding an exponential backoff / taking advantage of using object metadata to avoid rate limit error on updating `.info` file.

Acconut:
Thanks for reporting. If I understand correctly, this excerpt for the GCS documentation explains this behavior:

> There is no limit to how frequently you can create or update different objects in a bucket. However, a single particular object can only be updated or overwritten up to once per second. For example, if you have an object bar in bucket foo, then you should only upload a new copy of foo/bar about once per second. Updating the same object more frequently than once per second may result in 429 Too Many Requests errors. You should retry failed requests using truncated exponential backoff.

We would be happy to accept a PR for retrys and/or alternative solutions but we don't have the resources to work on this on our own.

Alternatively, you might want to reduce the number of requests you send for one upload. Sending more than 1 req/s for a single upload sounds wasteful to me. Is there a reason why you are doing this?

PokeGuys:
The problem is `chunkSize` cannot be controlled by server-side configuration. If the `chunkSize` is too small or the user's network speed is too fast, it will trigger the GCS rate limit by keep changing the `Offset` in `.info` file.

Acconut:
Do you control the clients and can increase their `chunkSize`? In most cases it is not even necessary to set the chunkSize parameter anyways.

And, as I said, we are open for PRs!

rbastic:
Hi!

@Acconut - how do you feel about adding generic debounce and throttle implementations to the code that can be applied to any store? I can try to take a crack at a PR soon.

If we make these configurable, they could be applied to any potential cloud provider. I'm already imagining Spaces support for DigitalOcean and may consider this scenario as a use case. From the documentation, I'll have to support these limits:

[...]

PR #1300, opened 2025-08-04
GCS storage: Do not write info on getInfo

Antti:
I didn't find the reason why is it writing back to info file when reading info (S3 storage is not doing it)

Fixes https://github.com/tus/tusd/issues/343


Acconut (approved):
Looks good, thank you!

PR #1343, opened 2026-02-07
build(deps): use std or local helper to simplify deps

ferhatelmas:
* use more testify for denser code which also drops x/exp constraints reference (only for e2e since other changes touch)

* use std for slog/slices instead of x/exp

* use a simple tcp proxy to drop toxiproxy/zerolog which are only used for e2e tests otherwise


ferhatelmas:
sure

> for dropping x/exp/slices

https://github.com/tus/tusd/pull/1353

PR #1366, opened 2026-04-16
Add Idempotency-Key header support for upload creation requests

### LINKED ISSUE #1364: Concatenation operation is not idempotent, causing duplicate uploads on retried requests
**Describe the bug**                                                                                                                                                                         
When a client sends a `POST` request with `Upload-Concat: final;url1 url2 ...`, the server creates a new upload, concatenates the partial uploads, and returns `201 Created`. If something in the request chain kills the connection before the response reaches the client due to a long file concatenation time (e.g. a WAF timeout rule, proxy timeout, or network interruption), the client never receives the response and tusd never emits the finished events/hooks. On retry, the server creates a **new** final upload with a new random ID and performs the concatenation again, wasting storage and triggering duplicate processing via hooks.

**To Reproduce**                                          
  1. Upload two partial uploads (`Upload-Concat: partial`)
  2. Send a `POST` with `Upload-Concat: final;/files/a /files/b`
  3. Kill the connection before the `201 Created` response is delivered to the client (e.g. via a WAF timeout rule or proxy timeout)
  4. Retry the same `POST` with `Upload-Concat: final;/files/a /files/b`
  5. Observe that a second, duplicate final upload is created and concatenation runs again

**Expected behavior**                                     
The concat operation should be idempotent. When the same set of partial uploads is presented for concatenation a second time, the server should detect that a final upload already exists for those partials and:                                      
  - If concatenation already completed (offset == size): return the existing final upload
  - If the upload was created but concatenation didn't start (offset == 0): retry the concatenation on the existing upload                                                                     
  - Not create a duplicate final upload

**Setup details**                  
  - Used tusd version: v2.9.2
  - Used tusd data storage: Disk storage using a PVC within K8s
  - Used tus client library: node tus client

m7kvqbe1:
## Summary

- Add support for the [Idempotency-Key](https://www.ietf.org/archive/id/draft-ietf-httpapi-idempotency-key-header-07.html) header on upload creation (`POST`) requests, allowing clients to safely retry without creating duplicate uploads
- Introduce an `IdempotencyKeyStore` interface with two implementations: file-based (for disk storage backends, persists across restarts) and in-memory (for cloud storage backends)
- Handle interrupted concatenation state: retry if offset is 0, return existing if complete, error if corrupted (0 < offset < size)
- Wire up the appropriate store automatically in the CLI based on the active storage backend

## Motivation

Closes #1364

Upload creation requests (particularly concatenation) are not currently idempotent. When a concat request is interrupted (e.g. by a WAF timeout), the client never receives the response and retries, creating a duplicate final upload.

This implements the approach suggested by @Acconut in #1364: a general-purpose `Idempotency-Key` header that works for **all** upload creation requests, not just concatenation. For the filestore backend, key-to-upload-ID mappings are persisted to disk alongside upload data (as `.idempotency-key` files), so they survive server restarts.

See also #1365 for a previous concat-only approach.

## Design

**New interface** (`IdempotencyKeyStore`) added to `StoreComposer`, following the same pattern as `Locker`:

```go
type IdempotencyKeyStore interface {
    FindUploadID(ctx context.Context, key string) (string, error)
    StoreUploadID(ctx context.Context, key string, uploadID string) error
}
```

**Two implementations shipped:**

| Backend | Store | Durability |
|---------|-------|-----------|
| filestore | `fileidempotencystore` (disk) | Survives restarts |
| S3/GCS/Azure | `memoryidempotencystore` (memory) | Within process lifetime |

**Handler logic** (in `PostFile`):
1. Read `Idempotency-Key` header
2. If store configured and key present: look up existing upload
3. If found and complete: return existing (201 + Location)
4. If found and incomplete concat (offset==0): retry concatenation
5. If found and corrupted concat (0 < offset < size): return 500
6. If found and non-concat: return existing upload info for PATCH resume
7. If not found: create normally, store key->ID mapping

**Opt-out:** `--disable-idempotency` CLI flag.

## Test plan

- [x] `TestIdempotency/ConcatRetryComplete` -- retried concat with already-complete upload returns existing
- [x] `TestIdempotency/ConcatRetryIncomplete` -- retried concat with offset==0 retries the concatenation
- [x] `TestIdempotency/ConcatRetryCorrupted` -- retried concat with partial data returns error
- [x] `TestIdempotency/RegularUploadRetry` -- retried regular upload returns existing with offset
- [x] `TestIdempotency/NoStoreConfigured` -- header ignored when no store configured
- [x] `TestIdempotency/NoHeaderSent` -- normal flow when no header
- [x] `TestIdempotency/DeletedUploadFallsThrough` -- stale mapping to deleted upload creates new upload
- [x] `TestFileIdempotencyStore` -- file-based store unit tests
- [x] All existing handler tests continue to pass
- [x] Full `go test ./pkg/...` passes

### DIFF
--- a/cmd/tusd/cli/composer.go
+++ b/cmd/tusd/cli/composer.go
@@ -197,6 +197,8 @@ func CreateComposer() {
 
 		if !Flags.DisableIdempotency {
 			idempotencyStore := fileidempotencystore.New(dir)
+			idempotencyStore.DirModePerm = os.FileMode(Flags.DirPerms) & os.ModePerm
+			idempotencyStore.FileModePerm = os.FileMode(Flags.FilePerms) & os.ModePerm
 			idempotencyStore.UseIn(Composer)
 		}
 	}
--- a/cmd/tusd/cli/composer.go
+++ b/cmd/tusd/cli/composer.go
@@ -13,8 +13,10 @@ import (
 	"github.com/tus/tusd/v2/pkg/azurestore"
 	"github.com/tus/tusd/v2/pkg/filelocker"
 	"github.com/tus/tusd/v2/pkg/filestore"
+	"github.com/tus/tusd/v2/pkg/fileidempotencystore"
 	"github.com/tus/tusd/v2/pkg/gcsstore"
 	"github.com/tus/tusd/v2/pkg/handler"
+	"github.com/tus/tusd/v2/pkg/memoryidempotencystore"
 	"github.com/tus/tusd/v2/pkg/memorylocker"
 	"github.com/tus/tusd/v2/pkg/s3store"
 
@@ -83,6 +85,11 @@ func CreateComposer() {
 
 		// Attach the metrics from S3 store to the global Prometheus registry
 		store.RegisterMetrics(prometheus.DefaultRegisterer)
+
+		if !Flags.DisableIdempotency {
+			idempotencyStore := memoryidempotencystore.New()
+			idempotencyStore.UseIn(Composer)
+		}
 	} else if Flags.GCSBucket != "" {
 		if Flags.GCSObjectPrefix != "" && strings.Contains(Flags.GCSObjectPrefix, "_") {
 			stderr.Fatalf("gcs-object-prefix value (%s) can't contain underscore. "+
@@ -113,6 +120,11 @@ func CreateComposer() {
 
 		locker := memorylocker.New()
 		locker.UseIn(Composer)
+
+		if !Flags.DisableIdempotency {
+			idempotencyStore := memoryidempotencystore.New()
+			idempotencyStore.UseIn(Composer)
+		}
 	} else if Flags.AzStorage != "" {
 
 		accountName := os.Getenv("AZURE_STORAGE_ACCOUNT")
@@ -156,6 +168,11 @@ func CreateComposer() {
 
 		locker := memorylocker.New()
 		locker.UseIn(Composer)
+
+		if !Flags.DisableIdempotency {
+			idempotencyStore := memoryidempotencystore.New()
+			idempotencyStore.UseIn(Composer)
+		}
 	} else {
 		dir, err := filepath.Abs(Flags.UploadDir)
 		if err != nil {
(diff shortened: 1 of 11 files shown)


Murderlon:
Review by Devin: https://app.devin.ai/review/tus/tusd/pull/1366

m7kvqbe1:
> Review by Devin: https://app.devin.ai/review/tus/tusd/pull/1366

Thanks @Murderlon.

I've added some fixups to this PR based on Devin's review:

- https://github.com/tus/tusd/pull/1366/commits/684a8f828a6e98ce01e162943f73936a9ee7f67b
- https://github.com/tus/tusd/pull/1366/commits/6b918fab990ad81f5cc15d49e15c8350fb75e4e7
- https://github.com/tus/tusd/pull/1366/commits/917c98fdf8bef90570506c8ef1a271309ae874ae

Murderlon:
cc @Acconut

Acconut (commented):
Thank you for opening this PR. I'm wondering if this is actually capable of solving the timeout issues when concatenating large uploads. Tusd would just retry the concatenating operation when it detects an existing upload resource with the same idempotency key.

But if we assume that the concatenation operation times out because the underlying process is too slow, retrying the operation would likely not help. And tusd aborts the concatenation operation whenever the request times out on the client or proxy side. Unless the process is just sometimes too slow, and there is a chance that it will be faster at the new attempt, retrying offers little improvement. Tusd would actually have to resume the concatenation operation, but I'm not sure this is possible for all storage implementations. It might be doable for filestore and gcsstore, but not for s3store as far as I can tell.

Idempotency keys are a helpful tool to prevent duplicate upload resources from being created, but it might not be that helpful for solving concatenation failures on its own as I initially assumed. If we take a look at the original issue of duplicate uploads because of failed concatenations, the duplicates are just a symptom, but not the actual fault. That is the concatenation failing in the first place due to the request timing out. To prevent that  the concatenation could be run in a background process, which has more time to complete than the short-lived request context. So even if the request times out after X seconds, the concatenation can continue. Deduplication via Idempotency-Key would still play an important role here as it ensures that client retrieves the original upload resources when it retries the upload concatenation request.

What do you think about this idea? The hurdle to getting this in tusd is higher since it currently doesn't have a system for running tasks in the background. But we might want to have this for other features as well (e.g. cleaning up old uploads).

m7kvqbe1:
https://github.com/tus/tusd/issues/1364#issuecomment-4587155260

PR #1379, opened 2026-06-12
azurestore: Fix InvalidHeaderValue error due to empty sentinel block

DotaPie:
Need this fix for my deployment of TUSD in production. This should fix #1358.

### DIFF
--- a/pkg/azurestore/azureservice.go
+++ b/pkg/azurestore/azureservice.go
@@ -15,12 +15,14 @@
 package azurestore
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"encoding/binary"
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


DotaPie:
Re-pushed commit (added GPG verification)

Acconut:
Thank you for submitting a fix! Can you describe what you've done and how it addresses the problem?

DotaPie:
Problem: empty sentinel block staged in NewUpload (to fix #1286) is length = 0, and azure rejects zero-length StageBlock with InvalidHeaderValue (#1358), so NewUpload fails

Downgrading to v2.9.0 was first workaround for our problem. This PR (forked from latest main) stages the sentinel as a 1-byte block under a reserved ID instead, keeping #1349's fix intact.

We built and deployed it ([ghcr.io/dotapie/tusd:fix-1358](https://github.com/DotaPie/tusd/pkgs/container/tusd)) to our AKS cluster  and redirecting uploads to azure blob storage now works just fine.

Edit 1: It was passing E2E test, because it is using azurite, which accepts zero-length StageBlock

Acconut (approved):
Thank you for fixing this issue!

PR #1380, opened 2026-06-23
Update Go version and dependencies

hamishforbes:
There's a bunch of 'critical' CVEs in both Alpine, Go stdlib and primarily golang.org/x/crypto.
While these CVEs probably aren't actually relevant to tusd they do get flagged by scanners so would be nice to have them fixed! :)


Acconut (approved):
Thank you!

PR #1381, opened 2026-06-24
azurestore: Support servecontent

### LINKED ISSUE #1382: azurestore: implement ServeContent
As previously discussed I've taken up the feature PR of @xelat09 to implement `ServeContent` for azure.

I've created a draft PR https://github.com/tus/tusd/pull/1381 for initial feedback. I've noticed that the behavior for retrieving empty files is different between stores
- s3: Returns `204 StatusNoContent`
- file: Returns `200 OK`
- gcs: Not possible

As far as I can tell azure blob API and S3 API actually both yield `200 OK` and `Content-Length: 0` for empty files. But I'm unsure what the desired return for tusd would be and whether changing it for azurestore to 200 OK would be fine.

**Context**:

> create a branch based on the head of this PR (xelat09:main)

Great, this sounds like the best approach. We squash PRs when merging, so the commit history is not too relevant.

_Originally posted by @Acconut in https://github.com/tus/tusd/issues/1234#issuecomment-3975029328_

quality-leftovers:
Supersedes https://github.com/tus/tusd/pull/1234

> Implement ContentServerDataStore for azure blob storage too, tests with azurite+azure blob are ok.
> Incomplete range requests (Range: bytes=100- ) are not supported, hope that's ok

**Behavioral changes**
- Return 200 OK instead of 204 NoContent for empty file. This is currently inconsistent between stores. See e2e test `TestDownload`. I'm unsure on whether this is ok.


pkg/azurestore/azureservice.go
```diff
@@ -364,3 +470,49 @@ func checkForNotFoundError(err error) error {
...
+		// zero value count indicates from the offset to the resource's end, suffix-length is not required
+		input.Range = azblob.HTTPRange{Offset: 0, Count: 0}
+		bytesEnd := 0
+		if _, err := fmt.Sscanf(val, "bytes=%d-%d", &input.Range.Offset, &bytesEnd); err != nil {
+			if _, err := fmt.Sscanf(val, "bytes=%d-", &input.Range.Offset); err != nil {
+				return nil, err
+			}
+		}
+		if bytesEnd != 0 {
+			input.Range.Count = int64(bytesEnd) - input.Range.Offset + 1
```
Acconut, line 490: Bug: `Range: bytes=0-0` sets `bytesEnd` to 0, so this condition skips setting `Count` and Azure treats the range as open-ended (entire blob from offset 0) instead of a single byte.

Please use an explicit “end was parsed” flag (or `*int64`) rather than treating `0` as absent. Worth a unit test for `bytes=0-0` → `Offset: 0, Count: 1`.

pkg/azurestore/azureservice.go
```diff
@@ -230,6 +234,103 @@ func (blockBlob *BlockBlob) Download(ctx context.Context) (io.ReadCloser, error)
...
+	if err != nil {
+		return err
+	}
+	result, err := blockBlob.BlobClient.DownloadStream(ctx, downloadOptions)
+	if err != nil {
+		var azureError *azcore.ResponseError
+		if errors.As(err, &azureError) {
+			if http.StatusNotFound == azureError.StatusCode {
+				// Either upload is incomplete, or the upload content blob has been deleted and it cannot be served.
+				return handler.ErrNotFound
```
Acconut, line 249: When the handler reaches `ServeContent`, the upload already exists (`GetUpload` succeeded). Mapping Azure 404 to `handler.ErrNotFound` makes incomplete uploads look like they don’t exist.

`s3store` returns `ERR_INCOMPLETE_UPLOAD` (400) in this situation — please do the same here for consistent client behavior.

pkg/azurestore/azureservice.go
```diff
@@ -230,6 +234,103 @@ func (blockBlob *BlockBlob) Download(ctx context.Context) (io.ReadCloser, error)
...
+			}
+
+			// Pass-through 412 and 416 with relevant headers
+			if http.StatusRequestedRangeNotSatisfiable == azureError.StatusCode ||
+				http.StatusPreconditionFailed == azureError.StatusCode {
+				if azureError.RawResponse != nil {
+					for _, header := range []string{"Content-Range", "X-Ms-Error-Code", "X-Ms-Request-Id", "Date"} {
```
Acconut, line 256: Please don’t forward `X-Ms-Error-Code` / `X-Ms-Request-Id` to clients — no need to expose the storage backend that way. Keep `Content-Range` for 416 (and `Date` is fine if you want it).

pkg/azurestore/azureservice.go
```diff
@@ -230,6 +234,103 @@ func (blockBlob *BlockBlob) Download(ctx context.Context) (io.ReadCloser, error)
...
+	// StatusCode, therefore we need to handle ErrorCode
+	if result.ErrorCode != nil {
+		statusCode := http.StatusInternalServerError
+		code := bloberror.Code(*result.ErrorCode)
+		if code == bloberror.ConditionNotMet {
+			// We could check download options here, but just propagating ConditionNotMet should suffice
+			statusCode = http.StatusNotModified
+		}
+		w.WriteHeader(statusCode)
+		return nil
```
Acconut, line 280: For 304, `s3store` also forwards validators / cache headers (`ETag`, `Cache-Control`, `Last-Modified`, etc.) from the backend response. Would be good to do the same when those fields are present, so conditional GETs behave consistently across stores.

pkg/azurestore/azureservice.go
```diff
@@ -230,6 +234,103 @@ func (blockBlob *BlockBlob) Download(ctx context.Context) (io.ReadCloser, error)
...
+	}
+
+	// Add Accept-Ranges, Content-*, Cache-Control, ETag, Expires, Last-Modified headers if present in azure response
```
Acconut, line 283: Nit: please drop `Expires` from this comment. We don’t forward it, and Azure’s `ExpiresOn` (`x-ms-expiry-time`) is blob expiry — not the HTTP `Expires` cache header — so we shouldn’t map/forward it as such.

pkg/azurestore/azureservice.go
```diff
@@ -364,3 +470,49 @@ func checkForNotFoundError(err error) error {
...
+		bytesEnd := 0
+		if _, err := fmt.Sscanf(val, "bytes=%d-%d", &input.Range.Offset, &bytesEnd); err != nil {
+			if _, err := fmt.Sscanf(val, "bytes=%d-", &input.Range.Offset); err != nil {
+				return nil, err
+			}
+		}
+		if bytesEnd != 0 {
+			input.Range.Count = int64(bytesEnd) - input.Range.Offset + 1
+		}
+	}
```
Acconut, line 492: A few Range-parsing notes:

1. **Invalid / unusable `Range`**: RFC 9110 allows ignoring a bad `Range` (respond `200` with the full body) or rejecting with `416`. Please don’t return a parse error that becomes a 500 — ignore the header (like `s3store` effectively does by letting the backend decide) or map to 416.
2. Same for invalid `If-Modified-Since` / `If-Unmodified-Since`: RFC 9110 says recipients **MUST ignore** non-HTTP-dates; don’t error out.
3. Please add a short comment that **suffix ranges** (`bytes=-N`) and **multi-range** requests aren’t supported because Azure’s `HTTPRange` only expresses `offset`/`count` (no native suffix form). No need to implement suffix ranges in this PR.

pkg/azurestore/azureservice.go
```diff
@@ -364,3 +470,49 @@ func checkForNotFoundError(err error) error {
 	}
 	return err
 }
+
+// parse the Range, If-Match, If-None-Match, If-Unmodified-Since, If-Modified-Since headers if present
+func ParseDownloadOptions(r *http.Request) (*azblob.DownloadStreamOptions, error) {
```
Acconut, line 475: Please unexport this (`parseDownloadOptions`). It’s an internal helper; the tests can live in package `azurestore` (or otherwise call it without exporting).

pkg/azurestore/azurestore_test.go
```diff
@@ -435,6 +438,130 @@ func TestDeclareLength(t *testing.T) {
...
+			}
+			w.WriteHeader(http.StatusOK)
+
+			// Write response body
+			_, err := w.Write([]byte(expectedBody))
+			return err
+		},
+	).Times(1)
+
+	err := blockBlob.ServeContent(ctx, rec, req)
```
Acconut, line 493: This doesn’t actually test `AzUpload.ServeContent` / `AsServableUpload` — it mocks `AzBlob` and calls `blockBlob.ServeContent` directly. Could you wire it through the upload like `s3store`’s `TestS3ServableUploadServeContent`?

Also, `"CacheControl"` isn’t a real HTTP header name (`Cache-Control`); the assertion only passes because the mock sets the same key.

internal/e2e/e2e_test.go
```diff
@@ -120,6 +127,147 @@ func TestSuccessfulUpload(t *testing.T) {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// Stores behave differently here.
+	store := detectStoreBackend()
+	expectedStatusCode := http.StatusOK
+	switch store {
+	case STORE_BACKEND_S3:
+		// S3 will return 204 NoContent
+		expectedStatusCode = http.StatusNoContent
```
Acconut, line 238: Agree that Azure/file returning **200** for empty downloads is the right behavior (`http.ServeFile` does the same). S3’s 204 can be fixed in a follow-up so we can drop this special case later — not required in this PR.

Acconut (changes requested):
Thank you for picking this up and carrying it across the finish line — the Azure `ServeContent` path, conditional/range handling, and the new e2e download coverage are a solid step toward store parity.

I'm requesting a few changes for correctness and consistency with `s3store` / `http.ServeFile`. If you don't have time or interest to continue, I'm happy to implement this feedback myself on top of your branch.

### Must-fix
1. **`bytes=0-0` range parsing bug** — treating `bytesEnd == 0` as “no end” turns a single-byte request into an open-ended range.
2. **Incomplete uploads** — prefer something like S3’s `ERR_INCOMPLETE_UPLOAD` (400) over `ErrNotFound` (404) when the upload exists but content isn’t servable yet.
3. **Invalid `If-*-Since` / malformed `Range`** — per RFC 9110, **ignore** invalid dates and unusable ranges (full `200`, or `416` for range); don’t bubble parse errors into 500s.
4. **Don’t forward `X-Ms-*` on 412/416** — keep `Content-Range` (and `Date` if useful); avoid exposing the Azure backend to clients.
5. **Unexport `ParseDownloadOptions`** (and adjust tests accordingly).
6. **Drop “Expires” from the success-path header comment** — Azure’s `ExpiresOn` is `x-ms-expiry-time` (blob expiry), not HTTP `Expires`; we shouldn’t imply we forward it.
7. **Comment that suffix ranges (`bytes=-N`) and multi-range requests aren’t supported** because the Azure SDK `HTTPRange` has no suffix/multi-range form (same practical limitation as S3 for multi-range). No need to implement suffix ranges in this PR.

### Nice to have
- Forward typical 304 headers (`ETag`, `Cache-Control`, `Last-Modified`, …) like `s3store`.
- Fix `TestAZServableUploadServeContent` so it actually exercises `AzUpload.ServeContent` / `AsServableUpload` (today it calls the mock blob directly; also `"CacheControl"` isn’t a real header name).
- Prefer `t.Skip(...)` over `return` for GCS e2e skips.

### Agreed / out of scope
- **Empty blob → 200 OK** is correct (matches `http.ServeFile` / filestore / Azure). S3’s 204 can be aligned in a follow-up, not this PR.

PR #1394, opened 2026-09-13
e2e: Allow Go 1.27 shutdown exit delay in TestShutdown

Acconut:
## Summary
- `TestShutdown` started failing when CI's `setup-go: stable` moved to Go 1.27: the 503 still arrives at ~2s, but process exit takes an extra ~500ms because `net/http` sleeps `rstAvoidanceDelay` when closing a connection with unread request body data.
- Keep asserting the shutdown response is prompt (~2s ±10%), and allow process exit up to 3s so both Go 1.26 and 1.27 pass.

## Test plan
- [x] `go test -count=1 -run '^TestShutdown$' ./internal/e2e/` with a tusd binary built on Go 1.27
- [ ] Confirm End-to-end tests workflow is green on this PR


Acconut:
The S3 end-to-end tests are still failing because the Minio Docker image is no longer available. This will be fixed in another PR.

